PR #10, opened 2017-02-01
Client and Server expects account in ace_auth to be lowercase only.

LtRipley36706:
This is my first pull request, please let me know if I'm doing this incorrectly.

Simple change I made to the accountcreate command to correct and prevent the issue I experienced when setting up ACE.

The client always sends accounts as lowercase regardless of what the user types in the command line. The server itself resolves the difference and connects as well. The issue surfaces when attempting to create a character. That packet is sent, but no response is ever recv'd.


Rawaho:
Additional checks will be added on account login as well.

PR #11, opened 2017-02-02
Run faster, jump higher, teleport to coordinates (rather than XYZ position)

sbaum23:
These values are hardcoded and need to be replaced with actual values from the Player object.  This is mainly to let people run around faster in the emu until code it put in to send all player skills to client.

I'm Xen on Discord.


sbaum23:
Added /teleto debug command to teleport to coordinates.  Z value is currently hard coded to 200 because it needs to be read from a file (dat?).  

Also, the two methods to convert Coordinates to XYZ are in DebugCommands.cs because I'm not sure where you want to put them.

sbaum23:
I blew away my fork.

Issue #15, opened 2017-02-03
Password storage should use PBKDF2


zbuc:
Right now it's using salted SHA256, which works for now but isn't suitable for real password storage, taking into considerations things like password reuse across services and the fact that inexperienced server admins will likely be operating poorly configured ACEmulator servers.

To minimize the damage of leaking any password hashes from the DB, the supported standard for password storage on .NET is PBKDF2.

I am already working on converting password storage to use PBKDF2 on my branch.

https://dusted.codes/sha-256-is-not-a-secure-password-hashing-algorithm

PR #21, opened 2017-02-04
Change password hashing to use scrypt

zbuc:
I switched the password hashing over to use libsodium's scrypt
implementation. This involved adding a new migration.

I noticed that currently the password isn't checked. I started looking
into how to implement password checking and it looks like we will
probably need to have a launcher to do so.

The reason the commit touches so many files is because it also cleans up
the files with mixed line endings.


zbuc:
Fixes Issue #15

Mag-nus:
I like the idea, but, i don't feel it's necessary to visit until we know how account create will actually work.

Merging it into current though seems fine. Probably better sooner rather than later as I assume accounts will have to be recreated.

Thwargle:
Technically; we DO have a launcher. I will work with whoever necessary to implement this into the launcher when ready. It's currently stable and fully featured as far as i can tell; so it's at a stage where implementing true password passing between server and client launch is reasonable.

Mogwai-AC:
closing due to being outdated plus inability to merge.

PR #23, opened 2017-02-04
Add Config.json to .gitignore

zbuc:
This also forces first-time configuration.

Existing Config.json copied to Config.json.example

README.md has been updated with instructions.


LtRipley36706:
Makes a lot of sense to me. Resetting the current config.json for each branch/pull is a bit of a hassle currently, this way you only have to reset when the file NEEDS to change

PR #25, opened 2017-02-04
Option to Enable Automatic Account Creation

LtRipley36706:
Added a configurable variable to Config.json for allowing the server to create an account for the client when one does not exist on the server.


LtRipley36706:
See PR #34

PR #26, opened 2017-02-04
Fixing up Character Delete and Restore functions

LtRipley36706:
In looking at the character deletion and restore section it made sense to me that name in character database shouldn't be made a unique value
At present in master, even if that character was marked as deleted, the character name is held.

I've made changes to the code and database to better emulate how it worked before with respect to deletions and restores.

As is, no character is ever marked deleted so that is yet to be implemented, although I suspect that would be a server side SQL script or command within ACE running at an interval/manually via console as determined by a configuration variable(s) possibly in config.json for how often it checks, how long the restore window is and if to mark characters deleted or not which would remove them from restoration via client.

Going further than original design, perhaps it could be changed to hold the character name for that account only while the restoration window is active, as well as only allowing that account to create a brand new character with that deleted name.


LtRipley36706:
Closing this due to conflicts, see PR #35

PR #30, opened 2017-02-05
Fix for NAT issues on internal connections to WorldServer

LtRipley36706:
This should help alleviate the issue some people will see when configuring all IPs to external, but connecting to the server on internal IPs.


LtRipley36706:
Did the merge wrong again, apologies.. sending new PR shortly.

PR #32, opened 2017-02-05
Fix for NAT issues on internal connections to WorldServer

LtRipley36706:
This fix is meant help alleviate a NAT issue some people will see when configuring server IP to external address. 

Due to problematic NAT routing, connecting to external IP while on internal network leads to black screen on hand-off to world server. Others may not even be able to connect to server using external IP address.

This fix should allow for external IP connection for those outside of network, and internal IP connection for those inside the network.


LtRipley36706:
See PR #37 instead

Issue #20, opened 2017-02-04
Server should have a way to apply DB migrations


zbuc:
The server should have a way to track which DB migrations it has applied, and automatically apply missing ones in the correct order.

@Mag-nus suggested a naming convention of `YYYY_MM_DD_tablename.sql`. This would work, but would make dealing with pull requests and merges annoying, if multiple came in one day.

I would suggest a naming scheme more along the lines of `$patchNumber + '_' + $YYYY_MM_DD + '_' + $descriptiveFeatureName`, for example `01_2017_02_03_add_monster_information.sql` and not limiting them to any particular tables.

Picking the second naming system will make dealing with pull requests easier and help developers understand what updates are for. Additionally, the leading number provides a good way of tracking our schema revision.

zbuc:
I suggest a reasonable first version of this might look like:

1) an update to the core .sql file and `Config.json` to add a new `ace_global` database with a table `schema_revisions` tracking database schema versions

The columns might be something like:

`id int(auto increment), schema_name varchar, schema_version int`

also insert initial values into the `schema_revisions` table to set the current `schema_version` at 0 on all databases (`ace_auth`, `ace_character`, `ace_world`).

2) add a task that runs during server startup that uses the database names from `Config.json` and compares the .sql files found on the filesystem against the results from `ace_global.schema_revisions` and runs the new ones, and updates the database.

zbuc:
Almost done with this...

Mogwai-AC:
this was recently done with a powershell script that was merged.  if you're ok with using that, please close the issue.

zbuc:
The script can only recreate the initial database table state completely so it won't be useful for server operators.

If it were extended with a second script solely for running upgrades that could keep track of the currently run database upgrades somewhere, for example in a flat-file in a temporary directory, so they wouldn't be run again that would be useful.

Mogwai-AC:
The scripts would need to be altered to be re-run safe.  MySQL's support for this is spotty, but it could be done.

zbuc:
For a solution that doesn't require migration authors to worry about that, do something similar to the pull request I created where it tracks the database migrations that have run and only runs them once, in sequential order. 

That is relatively standard for database-backed applications.

You will also run into difficulty with migrations where they are interdependent, i.e. one migration needs to run before another one can be run, because it creates or alters tables used by the other migration. 

My pull request took all this into account. 

With your PowerShell solution you could probably track the migrations that have run in a text file or registry key.

[...]

PR #36, opened 2017-02-05
Add automatic database migrations to startup

zbuc:
Database migrations will automatically be run.
Name them according to the format (`#_YYYY-MM-DD-descriptive_name.sql`)
and put in the appropriate directory and it will run on the next server
startup.

Migrations are only run once, in ascending numeric order. The current
database state is tracked in a new database called `Global`.

This simplifies server setup: now the only step is to create the four
databases. The base SQL and migrations will be run automatically from
that point on.

This also simplifies migrations, which will no longer need to
incorporate complex conditional logic in order to be repeatable.


zbuc:
I'm not a C# guy and there are some stylistic things I don't like, mainly calling out to `DatabaseManager.Global` in the main `Database` class is very hacky, as is what I had to do with the `Global` prepared statements.

But, it all works and I think is good for a first pass.

zbuc:
Fixes #20

zbuc:
I cleaned up and simplified some of the code, and rebased against `master`.

All current migrations run and the database is in a working state after compiling and running off this branch.

Forbiddenz:
This is s great idea but I do wonder if this entire process should be separated into its own project in the solution.  I.e. Patcher because this is not really a server process but more so a patcher / installer issue

Mag-nus:
I'm not fond of the complexity for this. What i'd like to see instead is us merge our updates down into the base sql file at some point, and, in the future, when new updates are added, also update the base sql file as well, so a new install wouldn't need to apply the updates, as they'd be.. up to date. (i'm by no means a db person so i don't know if this is against the norm)

Forbiddenz:
Hi Mag-nus,
The problem with your idea is some asshat will figure out how to screw up their original db.

I worked on many production level database patches and they pretty much all start at step x and move forward where x can be the current software version.

Your right as a new installer needs to run from version 0 and then if version 50 removes a table in version 0 it will be a useless step and take more time.

But the benefits of the system far outweigh the useless garbage.

With a design that goes by version and up you can guarantee that the database is always standard and it eliminates the room for error with things getting out of sync or even missed.

When you have the step process it's a little more work and sometimes seems dumb but it keeps all the db updates on a forward track even if they seem to be doing things the slow way.

PR #37, opened 2017-02-05
Add option to send different host address than external host address during world handoff

LtRipley36706:
The purpose of this code is to allow a server operator to configure their external IP for clients connecting from the internet as well as allowing clients connecting on the intranet to use the local network IP address of the server instead.

The options are configurable via config.json and set to false by default.

### DIFF
--- a/Source/ACE/Config.json.example
+++ b/Source/ACE/Config.json.example
@@ -4,6 +4,8 @@
         "Welcome": "Welcome to this ACE server!\nFor more information visit http://www.acemulator.org.",
         "Network": {
             "Host": "127.0.0.1",
+            "InternalHost": "127.0.0.1",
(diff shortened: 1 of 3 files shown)


Mag-nus:
rawaho made a change that might have fixed the issue you were trying to fix.

I think IP binding should be explicit. We might get a case where a user has 2 ip addresses on the same box and choses to bind each ip to a unique world, using the same port.

LtRipley36706:
it makes sense to me add variables for each ip address we explicitly send. It wouldn't really fix what this PR intention was which is to basically let me set up an external ip for external clients, and an internal ip for internal clients. I understand if its not something we want to put in to the master code, I'll just end up keeping a branch off my fork to roll in each time i update my server.

PR #50, opened 2017-02-09
Spending XP in Abilities, Vitals, and Skills, Saving Property buckets…

Mogwai-AC:
….  /grantxp command, moved persistence out of Player object for later refactoring of entity and data access project.


Source/ACE/Network/GameEvent/GameEventOpcode.cs
```diff
@@ -3,10 +3,33 @@
     // order by opcode name
```
Rawaho: Does anyone read comments?

sbaum23: I hope this is not the only issue holding this awesome PR back.

PR #73, opened 2017-02-18
Network Refactor

Zegeger:
Created GameMessage class.  Currently GameEvent inherits from
GameMessage.
GameMessage just build MemoryStream for message.
Moved all opcodes that are outside of GameEvent in protocol over to be a
GameMessage.
Created some basic (test) functions in NetworkManager to send the
GameMessages.
Lots more to do, but this compiles and logs in successfully.
Need to move group to per message property.
Need to create per session packet queue.
Need to rewrite packet pipe to better process multifragment packets.
Need to finish testing multifragment packets and get working.


Mogwai-AC:
I like it, but you need to resolve the merge conflicts first.

Mag-nus:
Also, double check your namespaces. It looks like GameMessageOpcode isn't in the correct namespace. Namespaces should match the folder depth of the file.

PR #76, opened 2017-02-18
Reworked Appear Offline functionality

sbaum23:
Reworked appear offline functionality so that it doesn't change IsOnline, which other things might depend on.

Also fixed issue with sending friends list updates that was introduced with the network refactor merge.


Source/ACE/Entity/Player.cs
```diff
@@ -31,6 +31,8 @@ public class Player : MutableWorldObject
         
         private Character character;
 
+        private Dictionary<SingleCharacterOption, bool> characterOptions; // Might want to move this to Character class
```
Mogwai-AC: Yes, move it.  Easily stored/saved as a bool property.

Mogwai-AC:
Resolve the conflicts and we'll approve as is.  We'll deal with persistence later.

PR #58, opened 2017-02-11
Added support on ACE console for setting per character admin rights and character renaming

LtRipley36706:
Also included in this PR is a fix for the propertybools (IsAdmin, etc) not always being set from database when a character logs in.

Issue #44, opened 2017-02-06
Add column to authenticationDB.account for access rights


LtRipley36706:
Need to add a column for access rights so to allow accounts to be able to create characters with IsAdmin or IsEnvoy bools set.

Mogwai-AC:
Kaos: We just need a simple rank system and /setrank command or similar that flows through. 1=registered, 2=envoy, 3=sentinel, 4=admin, etc... Everyone goes under 1 automatically and an admin can use the /setrank command to give out any additional ranks. Value should be stored in DB.

LtRipley36706:
for reference, values assigned to characters:

SetPropertyBool(PropertyBool.IsPsr, false);
SetPropertyBool(PropertyBool.IsAdmin, true);
SetPropertyBool(PropertyBool.IsArch, false);
SetPropertyBool(PropertyBool.IsSentinel, false);
SetPropertyBool(PropertyBool.IsAdvocate, false);

would think it goes like this:

0=player
1=advocate (isAdvocate)
2=sentinel (isSentinel)
3=envoy (isPSR)
4=developer? (IsArch)
5=admin (isAdmin)

Mogwai-AC:
Had a discussion about this in Discord tonight.  The client sends 2 uints for IsEnvoy and IsAdmin.  Need to clarify if this is 1 field with a bitmask, 1 field with tiered elevation, or multiple fields.

PR #77, opened 2017-02-19
Add AccessLevel to Accounts

LtRipley36706:
This addresses issue #44 and also contains some of PR #58 rolled in. 

Of note, CommandManager will now respect AccessLevels specified where previously it didn't care.

Set-AccountAccess will set the account's accesslevel
Set-CharacterAccess will set the character's accesslevel

CommandManager only cares about character's accesslevel as of this PR. Per discussion with @Mogwai-AC  I think we should code up a SUDO like command which then references to Account's AccessLevel for privileges. This way we emulate how it seemed to work on AC live (per character) in addition to then allowing accounts to self-promote characters, as they did with the admin button on character creation, as well as other admin functions on non-admin characters. 

Probably something along the lines of "SUDO set-characteraccess [mynewcharacter] sentinel" which just does a check back to session.accesslevel.


LtRipley36706:
Added functionality for /sudo as described above.

LtRipley36706:
Adjusted the set commands to work on both the console and in-game. The first account has to be created via console and also set as an admin for these to work in game.

Source/ACE/Command/Handlers/AccountCommands.cs
```diff
@@ -11,19 +11,128 @@ namespace ACE.Command.Handlers
...
-            Account acc = new Account(accountId, account, salt, password);
+            uint accountId          = DatabaseManager.Authentication.GetMaxId() + 1;
+            string account          = parameters[0].ToLower();
+            string salt             = SHA2.Hash(SHA2Type.SHA256, Path.GetRandomFileName());
+            string password         = SHA2.Hash(SHA2Type.SHA256, parameters[1]);
+            AccessLevel accessLevel = AccessLevel.Player;
+
+            if (parameters.Length > 2)
+            {
+                switch (parameters[2].ToLower())
```
Mogwai-AC: You should be able to force an enum.Parse on these instead of essentially duplicating the list.  Would simplify the code a lot.

Source/ACE/Command/Handlers/AccountCommands.cs
```diff
@@ -11,19 +11,128 @@ namespace ACE.Command.Handlers
...
+                else
+                    ChatPacket.SendSystemMessage(session, "Account " + accountName + " does not exist.");
+                return;
+            }
+
+            AccessLevel accessLevel = AccessLevel.Player;
+
+            if (parameters.Length > 1)
+            {
+                switch (parameters[1].ToLower())
```
Mogwai-AC: again, Enum.Parse should do here.  No need to do a messy switch (and less maintenance).

Source/ACE/Command/Handlers/CharacterCommands.cs
```diff
@@ -0,0 +1,74 @@
...
+        [CommandHandler("set-characteraccess", AccessLevel.Admin, CommandHandlerFlag.None, 1)]
+        public static void HandleCharacterTokenization(Session session, params string[] parameters)
+        {
+            uint characterId = 0;
+            string characterName = parameters[0];
+
+            AccessLevel accessLevel;
+            if (parameters.Length > 1)
+            {
+                switch (parameters[1].ToLower())
```
Mogwai-AC: As before, Enum.Parse

LtRipley36706:
@fantoms rightly pointed out that the most recent master changes changed SendSystemMessage to SendServerMessage with an enum argument for message type. Corrected that issue as well.

PR #92, opened 2017-02-26
Add in support for the old global Admin channels

LtRipley36706:
Resurrected the various Admin global channels used by the Advocates, Sentinels, Envoys, and Admins, stubbed the @on and @off commands and added support for @index and @clist commands.

Some TODOs have been put into the code for when we get a plan in place for for db loading/saving of channel subscriptions, as well as properly set subscriptions for that matter.


LtRipley36706:
See #95

PR #93, opened 2017-02-26
Support easy permissions in ACEmulator

LtRipley36706:
Improved best of both worlds scenario.

Server operators have the choice to emulate access permissions via more complicated system used on official worlds or they can choose to use simplified account based permissions which would become the default install option for ACEmulator.


LtRipley36706:
see #94

PR #107, opened 2017-03-06
Add DAT loading into main process

kdmxp256:
Obtains the two DAT file locations from the config.json file.  Code also handles file not found exceptions and prints out an error.

### DIFF
--- a/Source/ACE.DatLoader/DatManager.cs
+++ b/Source/ACE.DatLoader/DatManager.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Linq;
+using System.IO;
+
+using ACE.Common;
+
+namespace ACE.DatLoader
+{
+    public class DatManager
+    {
+        public static DatDatabase celldat;
+        public static DatDatabase portaldat;
+
+        private static string datFile;
+        private static int count;
+
+        public static void Initialise()
+        {
+            try
+            {
+                datFile = ConfigManager.Config.Server.DatFilesDirectory + "\\client_cell_1.dat";
+                celldat = new DatDatabase(datFile);
+                count = celldat.AllFiles.Count();
+                Console.WriteLine($"Successfully opened {datFile} file, containing {count} records");
+            }
+            catch (FileNotFoundException ex)
+            {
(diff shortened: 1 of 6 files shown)


kdmxp256:
So the file names themselves would be hard-coded and just pull in the directory location from the config file, thereby allowing the two config directives to condense into one.

LtRipley36706:
Exactly. I can't imagine a reason why the file names couldn't or shouldn't be hardcoded, but I can think of a few reasons why one might want to change the directory location. One of which could be having multiple versions of DAT patches to switch between perhaps?

kdmxp256:
Stupid tab vs spaces kept tripping me up.  I also noticed that I missed a whole file that should have been included in the original commit.  'git add <filename>, ftw.  It has now been added into the PR with the changes to the config variable(s).

Source/ACE.DatLoader/DatManager.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Linq;
+using System.IO;
+
+using ACE.Common;
+
+namespace ACE.DatLoader
+{
+    public class DatManager
+    {
+        public static DatDatabase celldat;
```
Mogwai-AC: public properties should be Proper Case - stylecop should be complaining about this (ie, CellDat and PortalDat).  Also, this is a bad use of a static member because it can get into an invalid state (ie, before Initialize is called).  This shouldn't happen in practice, I understand, but it's bad class design.

If you want to keep the static properties, that's fine, but Initialize logic should be moved to a static constructor if so.

Source/ACE/Config.json.example
```diff
@@ -16,7 +16,8 @@
             "DefaultAccessLevel": 0
             // 0 = Player | 1 = Advocate | 2 = Sentinel | 3 = Envoy | 4 = Developer | 5 = Admin
             // New accounts are created with the above AccessLevel unless explicitly changed using accountcreate command
-        }
+        },
+        "DatFilesDirectory": "c:\\ACE"
```
Mogwai-AC: put a comment here about what file names are expected, please.

PR #113, opened 2017-03-07
Max Rank Fireworks and Correct Skill Rank Up text Color

fantoms:
- I could not find a way to check the skill rank after it was set, so I added a function to check max rank. If the function returns a Boolean true, then the server will play fireworks when reached max rank

- Changed the message type to Advancement when increasing skills too reflect the correct skill advancement chat color.

- Added a few summary comments and fixed remarks/known issues.

### DIFF
--- a/Source/ACE/Entity/Player.cs
+++ b/Source/ACE/Entity/Player.cs
@@ -333,6 +333,28 @@ namespace ACE.Entity
             return result;
         }
 
+        /// <summary>
+        /// Check a rank against the skill charts too determine if the skill is at max
+        /// </summary>
+        /// <returns>Returns true if skill is max rank; false if skill is below max rank</returns>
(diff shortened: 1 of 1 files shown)


Source/ACE/Entity/Player.cs
```diff
@@ -348,13 +370,24 @@ public void SpendXp(Skill skill, uint amount)
 
             if (result > 0u)
             {
-                messageText = $"Your base {skill} is now {newValue}!";
+                //if the skill ranks out at the top of our xp chart 
+                //then we will start fireworks effects and have special text!
+                if (IsSkillMaxRank(ranks, status))
+                {
+                    //fireworks on rank up is 0x8D
+                    PlayParticleEffect(0x8D);
```
LtRipley36706: We should get these into an Enum soon

http://pastebin.com/mYAQKMzf

PR #116, opened 2017-03-09
Character Rename command added

LtRipley36706:
Works via console and client.

### DIFF
--- a/Source/ACE/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE/Command/Handlers/AdminCommands.cs
@@ -6,6 +6,7 @@ using ACE.Managers;
 using ACE.Network;
 using ACE.Network.GameMessages.Messages;
 using ACE.Common;
+using ACE.Database;
 
 namespace ACE.Command
 {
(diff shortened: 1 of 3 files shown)


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -877,13 +878,40 @@ public static void HandleRaise(Session session, params string[] parameters)
...
-        [CommandHandler("rename", AccessLevel.Envoy, CommandHandlerFlag.RequiresWorld, 2)]
+        // rename <Current Name> <New Name>
+        [CommandHandler("rename", AccessLevel.Envoy, CommandHandlerFlag.None, 2)]
         public static void HandleRename(Session session, params string[] parameters)
         {
             // @rename <Current Name>, <New Name> - Rename a character. (Do NOT include +'s for admin names)
 
-            //TODO: output
+            // As currently implemented below, the command does not exactly mimic retail usage in that the command was @rename oldName, newName
+            // and for us is @rename oldName newName || is this a big deal?? -Ripley
```
LtRipley36706: If we want to be stringent about sticking to using the comma as a separator instead of the space, let me know and I'll revise this.

PR #121, opened 2017-03-13
Landblocks, logging, object factory prep work

lurker-mcdoogle:
there are a lot of // TODO comments all over this thing.

here's what I know works:
/telepoi town1
/createlifestone
/telepoi town2
/telepoi town1
... and the lifestone is still there.

Features:
* Landblock Management
* Landblock Handoff of teleporting objects
* Broadcasting of objects on a landblock based on existence
* Update of objects in a landblock based on Movement
* Landblock adjacency broadcasting
* basic object factories
* log4net - not having console output was killing me and we needed a logging framework anyway. nuget to the rescue.
* moved "in front of" calculations to hang off the Position object for better reuse
* PackedDWORD object
* Delayed update of position after entering portal space seems to have helped porting stability.

Known Issues
* Adjacency matrix has issues broadcasting to the correct adjacencies. Shouldn't be a big deal - I'll fix it later.
* local chat is untested and possibly/likely broken.

Mag is showing up in the commit log because I squashed a commit with his name on it when attempting a rebase.  Ended up just cherrypicking into a new branch.

### DIFF
--- a/Source/ACE.Common/ACE.Common.csproj
+++ b/Source/ACE.Common/ACE.Common.csproj
@@ -30,6 +30,10 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="log4net, Version=2.0.8.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a, processorArchitecture=MSIL">
+      <HintPath>..\packages\log4net.2.0.8\lib\net45-full\log4net.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
       <HintPath>..\packages\Newtonsoft.Json.9.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
       <Private>True</Private>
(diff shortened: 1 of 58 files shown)


lurker-mcdoogle:
Known Issue: Adjacency matrix has issues broadcasting to the correct adjacencies. Shouldn't be a big deal - I'll fix it later.

Mag-nus:
Landblock.cs, line 289. Needs to be changed from > to >=

movedObjects = movedObjects.Where(p => p.LastUpdatedTicks > p.LastMovementBroadcastTicks).ToList();

// flag them as updated now in order to reduce chance of missing an update
movedObjects.ForEach(m => m.LastMovementBroadcastTicks = WorldManager.PortalYearTicks);


What's happening is a player is moving on the same PortalYearTick, right before Landblock.UseTime is called. When this happens, the > fails, and the cobject isn't processed as a move. That move gets skipped over.

Even though you set the LastMove to PortalYearTicks to indicate the object as processed, the next time UseTime is called, it's impossible for PortalYearTicks to have the same value as the previous call, thus, the >= won't be called in duplicate/error.

Mag-nus:
Landblock.cs UseTime() needs a Thread.Sleep(1) as the last line of the While() loop.

We also have variables that are being referenced across multiple threads. WorldObject.LastUpdateTicks and MutableWorldObject.LastMovementBroadcastTicks... I'm gonna think about this some more today.

PR #125, opened 2017-03-13
Save Player Properties to db

fantoms:
-Saves every 60 "ticks". Please advise on the best stats saving interval.

-Implemented database update functions for character properties.

-Added a character save ticker. Please advise on the proper location for this, I did this quickly.


Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -49,7 +49,13 @@ private enum CharacterPreparedStatement
...
             CharacterPropertiesDoubleInsert,
-            CharacterPropertiesStringInsert
+            CharacterPropertiesStringInsert,
+
+            CharacterPropertiesBoolUpdate,
+            CharacterPropertiesIntUpdate,
+            CharacterPropertiesBigIntUpdate,
+            CharacterPropertiesDoubleUpdate,
+            CharacterPropertiesStringUpdate
         }
```
fantoms: Created some prepared statements that work with SQL UPDATE functions. I did not see any update functions listed, but I don't know if this is by design. This seems weird to me though, as the database guy inside of me is saying that we should use the correct functions to manipulate what's already available, instead of trying to over-write everything.

Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -151,12 +164,14 @@ public bool IsNameAvailable(string name)
             return (charsWithName == 0);
         }
 
-        public Task UpdateCharacter(Character character)
+        public async void UpdateCharacter(Character character)
```
fantoms: Changed this function to async, to allow the occurrence of events independent of the main program flow. The idea was mainly derived from the CreateCharacter function below.

Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -382,6 +397,24 @@ public void SaveCharacterProperties(DbObject dbObject, DatabaseTransaction trans
                 transaction.AddPreparedStatement(CharacterPreparedStatement.CharacterPropertiesStringInsert, dbObject.Id, (ushort)prop.Key, prop.Value);
         }
 
+        public void UpdateCharacterProperties(DbObject dbObject, DatabaseTransaction transaction)
```
fantoms: Sends ALL of the character properties to the database. This function utilizes the SQL "update" syntax.

Source/ACE.Database/Database.cs
```diff
@@ -66,8 +66,15 @@ public void AddPreparedStatement<T>(T id, params object[] parameters)
                             using (var command = new MySqlCommand(query.Item1.Query, connection, transaction))
                             {
                                 for (int i = 0; i < query.Item2.Length; i++)
+                                {
                                     command.Parameters.Add("", query.Item1.Types[i]).Value = query.Item2[i];
-
+#if NETWORKDEBUG
```
fantoms: This debug will output the values sent to the database in the console. I would like to change the flag from `NETWORKDEBUG` to `DBDEBUG`, or equivalent term.

Source/ACE.Database/ICharacterDatabase.cs
```diff
@@ -20,7 +20,7 @@ public interface ICharacterDatabase
         
         Task<bool> CreateCharacter(Character character);
 
-        Task UpdateCharacter(Character character);
+        void UpdateCharacter(Character character);
```
fantoms: This is required to be a void due to changes in lines 167 of the CharacterDatabase.cs file.

Source/ACE/Entity/Player.cs
```diff
@@ -583,6 +583,11 @@ public void SaveOptions()
 
             // TODO: Save other options as we implement them.
         }
+
+        public void SaveCharacter()
```
fantoms: Added a hook to allow characters to be saved.

Source/ACE/Network/NetworkSession.cs
```diff
@@ -309,6 +309,9 @@ private void SendBundle(NetworkBundle bundle)
                             currentMessageFragment = new MessageFragment(bundle.Messages.Dequeue());
                         }
 
+                        //possibly an issue with messages being null
+                        //uncomment to test: if (currentMessageFragment.message == null) break;
```
fantoms: I am noticing that this crashes my server when the message is null, because you can't have a length of a null member?

Source/ACE/Network/Session.cs
```diff
@@ -108,6 +110,27 @@ public void Update(double lastTick)
                 logOffRequestTime = DateTime.MinValue;
                 SendFinalLogOffMessages();
             }
+            if (LastSaveTick == 0)
```
fantoms: For sake of readability, I left this very open. The timing will need to be changed to a better interval on the lastSaveTick. The is currently no local veriable for the interval, so I will need add that. Also believe that to increase performance, the CharacterOptions may need their own delay or moved into the logoff portion of Update.

fantoms:
Network Conflicts resolved.

Source/ACE.Database/CharacterDatabase.cs
fantoms: Also, in future code too save skills, we can duplicate this work for the other databases, then `properties`.

fantoms:
Closing now, this was merged into PR #141

PR #126, opened 2017-03-13
improved dat file content exporting

lurker-mcdoogle:
* improved cell.dat and portal.dat exporting
* created admin server commands for exporting
* unit tests that can be uncommented and run for exporting


lurker-mcdoogle:
Reopening with squashed commits.

PR #131, opened 2017-03-16
Create Training Wand

ogmage78:
Added some test code to create a training wand.   As part of this work, I put in some fixes in sizes, added AnimationFrame.   While this is not all that spectacular - it perfectly matches live server pcaps in every field.   

I would like to work on persisting this out to the database.  I assume we will need to implement a next object_id sequence as part of this.   In the test code it just sets an arbitrary object_id for the wand since it is not yet persisted.   Any feedback on the code would be appreciated - I am way out of date, but studying hard to get caught up so I can help.   Pea gave me the final piece of the puzzle - without being able to set animation frame, setting the flag caused an alignment mismatch.   God this was fun.   I forgot how much I used to love coding.


ogmage78:
I think I did something wrong.   Going to close this for now.

PR #132, opened 2017-03-16
Fixed message fragmentation

Zegeger:
Fixed fragmentation logic; we should not fragment messages that do not fill a full packet. Removed NetworkSession.Flush as this change fixed all cases where that seemed needed. Fixed NETWORKDEBUG section which was missed by network refactor

### DIFF
--- a/Source/ACE/Network/NetworkBundle.cs
+++ b/Source/ACE/Network/NetworkBundle.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 
 using ACE.Network.GameMessages;
+using System.Threading;
 
 namespace ACE.Network
 {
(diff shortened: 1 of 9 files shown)


Source/ACE/Network/NetworkBundle.cs
```diff
@@ -41,6 +42,9 @@ public bool SendAck
 
         public bool encryptedChecksum { get; set; }
 
+        public long currentSize = 0;
```
LtRipley36706: private?

Source/ACE/Network/NetworkBundle.cs
```diff
@@ -41,6 +42,9 @@ public bool SendAck
 
         public bool encryptedChecksum { get; set; }
```
LtRipley36706: private?

PR #134, opened 2017-03-16
Added attribute upper bounds check and max attribute fireworks and special text

fantoms:
-Added upper bounds checks for the Ranks so that Attributes will not crash the server.

-Added fireworks and special text on max rank, too match skills.


fantoms:
Closing now, since this was merged into PR #141

PR #142, opened 2017-03-18
Character will now level up on xp grants. Skill credits are also being assigned but are still not spendable.

damnubber:


Source/ACE/ACE.csproj
```diff
@@ -235,6 +236,9 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
     <None Include="App.config" />
+    <None Include="Config.json">
```
LtRipley36706: Config.json is being intentionally being left out of the csproj file and should not be included here

PR #146, opened 2017-03-19
Create Training Wand Demo code and start inventory management Drop/Equip/pickup

ogmage78:
Added Test Code to create Training Wand
Updated for new wo and landblock changes
Started work to drop item.   I want to add this interum work which compiles and works so that I might could get some assistance with the correct way to implement the event animation.

Thanks

Charles / Og II


ogmage78:
Jesus - I tried to squash it down to 1 commit and it went from 7 to 9.   I will get git - but damn it is kicking my ass right now.

PR #148, opened 2017-03-19
Add a link to ace-db docker image in the README

maxc0c0s:


Mogwai-AC:
So, here's the thing... Devs that know what to do with Docker don't need this - we can run scripts just fine.  Players that want to host their own servers don't even have a clue what Docker is.  Plus, you're linking to a docker image offsite that we can't control.

We don't think this is necessary, but in the end we might host a db image on acemulator.org when we cut releases.  I just don't think we're at a point where this is needed.

Thanks!

maxc0c0s:
No problem,

Since i'm using it, i thought maybe you guys might find it useful.

I'll find other way to contribute :)

Gordian3:
If we had the actual source and wanted to move it away from SQL Server and Windows Server, then deploying each piece (Database, GLS, etc) in a docker would have provided a great way to link all of the components.  Then have a VM OVF/OVA with CentOS or similar OS and have it load the dockers would have been a decent deployment model.

PR #149, opened 2017-03-20
added code for palettes to work correctly.   PaletteGuid - corrected …

ogmage78:
Small PR to fix bug with Palette method.   Added PaletteGuid and write it out as required.   Fixed Spelling.   The fix for this was originally implemented by @LytelthorpeThistledown     Getting this in sooner rather than later will help a lot of people as they are starting to work on this area of the code.

### DIFF
--- a/Source/ACE/Entity/ModelData.cs
+++ b/Source/ACE/Entity/ModelData.cs
@@ -11,23 +11,24 @@ namespace ACE.Entity
     /// </summary>
     public class ModelData
     {
-        
-        private List<ModelPallete> modelPalletes = new List<ModelPallete>();
+        public uint PaletteGuid = 0;
+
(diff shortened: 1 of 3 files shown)


Source/ACE/Entity/ModelData.cs
```diff
@@ -11,23 +11,24 @@ namespace ACE.Entity
     /// </summary>
     public class ModelData
     {
-        
-        private List<ModelPallete> modelPalletes = new List<ModelPallete>();
+        public uint PaletteGuid = 0;
```
Mogwai-AC: never expose public fields.  always use properties.
public uint PaletteGuid { get; set; } = 0;

ogmage78: Thanks - I will make the change now and will know that going forward.   I appreciate the feedback.

PR #152, opened 2017-03-20
Very basic animation

Zegeger:
Added basic animation message.  
Send logout animation when logging out.
Created a sequence manager and related classes, trying to better structure the numerous sequences we must store/track.  (Only started with animation sequences, but if we like this, can convert other sequences over)
General cleanup


Source/ACE/Network/GameMessages/Messages/GameMessageAnimation.cs
```diff
@@ -0,0 +1,118 @@
...
+            Writer.Write((uint)generalFlags);
+            foreach (var animation in animations)
+            {
+                Writer.Write((ushort)animation.Animation);
+                Writer.Write(animationTarget.Sequences.GetNextSequence(Sequence.SequenceType.Animation));
+                Writer.Write(animation.Speed);
+            }
+        }
+    }
+    public struct AnimationItem
```
Mogwai-AC: 1 object per .cs file, please.

PR #153, opened 2017-03-20
Feature: Spend Skill Credits

fantoms:
Within this request I've sent code that will allow the client to attempt too spend skill credits. If enough skill credits are available, the server will grant the skill to the the character. If the skill credits are not available or the skill has already been trained, then the server makes sure to send the correct game messages with Skill.None, to keep the client from locking up.

On the bottom of the Skill.cs, I've added an extension that allows for skills with proper names, example:
 `Sneak Attack` instead of `SneakAttack`.

### DIFF
--- a/Source/ACE.Entity/Character.cs
+++ b/Source/ACE.Entity/Character.cs
@@ -183,6 +183,21 @@ namespace ACE.Entity
         }
 
         /// <summary>
+        /// Sets the skill to trained status for a character
+        /// </summary>
+        /// <param name="skill"></param>
+        public bool TrainSkill(Skill skill, uint creditsSpent)
+        {
+            if(Skills[skill].Status != SkillStatus.Trained && Skills[skill].Status != SkillStatus.Specialized)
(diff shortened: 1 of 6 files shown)


Source/ACE.Entity/Character.cs
```diff
@@ -183,6 +183,21 @@ public Character(uint id, uint accountId)
         }
 
         /// <summary>
+        /// Sets the skill to trained status for a character
+        /// </summary>
+        /// <param name="skill"></param>
+        public bool TrainSkill(Skill skill, uint creditsSpent)
+        {
+            if(Skills[skill].Status != SkillStatus.Trained && Skills[skill].Status != SkillStatus.Specialized)
+                if(PropertiesInt[PropertyInt.AvailableSkillCredits] >= creditsSpent) {
```
Mogwai-AC: again, not going to fail a PR over it, but we use C# curly brace formatting over Java/Javascript (curly braces always go on the next line).

PR #156, opened 2017-03-20
Heal: Update Player object, then send its data

LtRipley36706:
Original code only made client aware of change, didn't update the server's player object.


fantoms:
Tested this utilizing multiple characters and passes were successful. 

Aside from this code, your command really helps points out the bug/non-issue in the client, where your total health is greater than the actual health - 

Example - In-game action sequence after Spending one rank up, for endurance (Total 61): 

1. Health Total: `1/31`
2. type: `@heal`
3. Health Total: `30/31`

PR #163, opened 2017-03-22
Support for specific server host binding

damnubber:


Source/ACE/Config.json.example
```diff
@@ -2,9 +2,10 @@
     "Server": {
         "WorldName": "ACEmulator",
         "Welcome": "Welcome to this ACE server!\nFor more information visit http://www.acemulator.org.",
-        "Network": {
-            "Port": 9000
-        },
+      "Network": {
+        "Host":  "127.0.0.1",
```
LtRipley36706: Nitpickmode: kill the extra space here to make line `"Host": "127.0.0.1",`

Source/ACE/Config.json.example
```diff
@@ -3,7 +3,7 @@
         "WorldName": "ACEmulator",
         "Welcome": "Welcome to this ACE server!\nFor more information visit http://www.acemulator.org.",
       "Network": {
-        "Host":  "127.0.0.1",
+        "Host": "127.0.0.1",
```
LtRipley36706: upon further thought, change this to `"Host": "0.0.0.0",`

That way, we're basically set to IPAddress.Any by default, then the server op can change it to be more specific.

as is, the default of 127.0.0.1 means the server is only gonna listen for local loopback

LtRipley36706:
Now, really, all good here 👍

PR #167, opened 2017-03-23
weenies weenies everywhere.  object structure pass for review.

Mogwai-AC:
yeah, so this works....


LtRipley36706:
Client currently crashes when the db objects are sent to it...

LtRipley36706:
Lifestones.. Lifestones everywhere!
139vy

mcreedjr:
Works for me, too Mogwai. Exciting to see this materialize through the process. Generalized object factory is going to quickly open a lot of other doors as well. Thanks for the effort!

PR #169, opened 2017-03-25
Updated Marketplace Recall

LtRipley36706:
@mp command better reflects how it worked in AC1Live. 

The movement check/cancel is not yet implemented.


Source/ACE/Entity/Player.cs
```diff
@@ -953,6 +953,18 @@ public void Logout(bool clientSessionTerminatedAbruptly = false)
...
+        public void SendToMarketPlace()
+        {
+            // TODO: link to Town Network marketplace portal destination in db, when db for that is finalized and implemented.
+            // private static readonly Position marketplaceDrop = new Position(23855548, 49.206f, -31.935f, 0.005f, 0f, 0f, -0.7071068f, 0.7071068f); // PCAP verified drop
+            // Session.Player.Teleport(marketplaceDrop);
+
+            // TODO: check current position of player against recorded position at time of command issuance, if they don't match, send the you've moved too far error (and thunk sound?).
+            // Enum_0498 of StatusMessageType1 ??
+            // Not sure what kind of tolerance there was with this check in respect to how far from initial start point one could move..
+            Session.Player.Teleport(new Position(23855548, 49.206f, -31.935f, 0.005f, 0f, 0f, -0.7071068f, 0.7071068f));
```
Mogwai-AC: This should still be a static position _somewhere_.  For lack of a better place, static values inside the Position class itself is a good place.  Also, you're already in the Player object so no need to do Session.Player.  Ideally, this line would be:
Teleport(Position.Marketplace);

Source/ACE/Network/GameAction/Actions/GameActionTeleToMarketplace.cs
```diff
@@ -14,7 +14,7 @@ public class GameActionTeleToMarketPlace : GameActionPacket
         public GameActionTeleToMarketPlace(Session session, ClientPacketFragment fragment) : base(session, fragment) { }
 
         // TODO: link to Town Network marketplace portal destination in db, when db for that is finalized and implemented.
```
Mogwai-AC: clean up commented out code.  We have history in git.

Source/ACE/Network/Session.cs
```diff
@@ -110,6 +112,12 @@ public void Update(double lastTick)
                 SendFinalLogOffMessages();
             }
 
+            if (marketplaceRequestTime != DateTime.MinValue && marketplaceRequestTime.AddSeconds(14) <= DateTime.UtcNow)
```
Mogwai-AC: There are a number of things for which a player has to wait for a timer.  Allegiance Recall, House Recall, Lifestone Recall, Marketplace, whatever.  It feels like the best solution is to use a generic Timer that does the work instead of one that's ultra-specific to "marketplaceRequestTime".  Also, the "end" time should be set upon receipt of the command, not calculated every game loop.

For example:
in Handle, you'd call this method of the player
```
Session.Player.SetTimedTeleport(Timespan delay, Position destination)
{
    delayedTeleportDestination = destination;
    waitingForDelayedTeleport = true;
    timeToTeleport = DateTime.UtcNow + delay;
}
```

Also remember, you'd then need to have Autonomous updates to the player's position cancel the timed teleport.
```
    waitingForDelayedTeleport = false;
    delayedTeleportDestination = null;
```

The Session Update would then do something like the following:

```
if (WaitingForDelayedTeleport && DateTime.UtcNow > Player.TimeToTeleport)
{
    Player.WaitingForDelayedTeleport = false;
    Player.Teleport(Player.DelayedTeleportDestination);
}
```

LtRipley36706:
See #223

PR #172, opened 2017-03-26
Implements GameAction/Event for Update Health

Lidefeath:
This is limited for Players currently but should be adjustable once creatures are in.

### DIFF
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -186,6 +186,23 @@ namespace ACE.Entity
             }
         }
 
+        /// <summary>
+        /// get the WorldObject specified by objectID
+        /// </summary>
+        public WorldObject GetWorldObjectByGuid(ObjectGuid objectId)
+        {
+            WorldObject wo = null;
+
+            lock (objectCacheLocker)
+            {
+                if (this.worldObjects.ContainsKey(objectId))
+                {
+                    wo = this.worldObjects[objectId];
+                }
+            }
+            return wo;
+        }
+
         public void SendChatMessage(WorldObject sender, ChatMessageArgs chatMessage)
         {
             // only players receive this
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -9,6 +9,8 @@ using ACE.Managers;
 
 using log4net;
 using ACE.Database;
+using ACE.Network.GameEvent.Events;
+using ACE.Network;
 
 namespace ACE.Entity
 {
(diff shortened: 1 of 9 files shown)


Source/ACE/Network/GameAction/Actions/GameActionQueryHealth.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public override void Handle()
+        {
+            // TODO: Extend to creatures as well, as we momentarily only have players
+
+            if (objectId.IsPlayer())
+            {
+                var tmpLandblock = this.Session.Player.Position.LandblockId;
+                var pl = (Player)LandblockManager.GetObjectByGuid(tmpLandblock, objectId);
```
Mogwai-AC: you already have the player.... this.Session.Player

There should be next to zero reason to return objects from the Landblock / LandblockManager.  That was a red flag to me, this is incorrect usage.  It's also not safe to extract as the player may cross a landblock boundary while running across the open world.

Source/ACE/Managers/LandblockManager.cs
```diff
@@ -52,6 +52,17 @@ public static void RelocateObject(WorldObject worldObject)
         }
 
         /// <summary>
+        /// Find a worldobject by its guid in the specified landblock
+        /// </summary>
+        public static WorldObject GetObjectByGuid(LandblockId blockId, ObjectGuid objectId)
```
Mogwai-AC: Remove this method.  This might be needed eventually, but for now it's a red flag for "you're doing it wrong."

Source/ACE/Entity/Landblock.cs
```diff
@@ -186,6 +186,23 @@ public void RemoveWorldObject(ObjectGuid objectId, bool adjacencyMove)
             }
         }
 
+        /// <summary>
+        /// get the WorldObject specified by objectID
+        /// </summary>
+        public WorldObject GetWorldObjectByGuid(ObjectGuid objectId)
```
Mogwai-AC: Remove this method.  This might be needed eventually, but for now it's a red flag for "you're doing it wrong."

Lidefeath:
The last commit (b5480f2) has the requested changes

Source/ACE/Entity/Landblock.cs
```diff
@@ -186,6 +188,39 @@ public void RemoveWorldObject(ObjectGuid objectId, bool adjacencyMove)
...
+                {
+                    if (this.worldObjects.ContainsKey(targetId))
+                        pl = (Player)this.worldObjects[targetId];
+                }
+                if (pl == null)
+                {
+                    // check adjacent landblocks for the targetId
+                    foreach (var block in adjacencies)
+                    {
+                        lock (objectCacheLocker)
```
Mogwai-AC: Good thinking, but unfortunately this particular lock isn't going to help you.  You're locking THIS landblock's mutex and poking your head into other landblocks.  You need to lock block.objectCacheLocker.

Source/ACE/Entity/Landblock.cs
```diff
@@ -186,6 +188,39 @@ public void RemoveWorldObject(ObjectGuid objectId, bool adjacencyMove)
...
+                        {
+                            if (block.Value.worldObjects.ContainsKey(targetId))
+                                pl = (Player)this.worldObjects[targetId];
+                        }
+                    }
+                }
+                float healthPercentage = (float)pl.Health.Current / (float)pl.Health.MaxValue;
```
Mogwai-AC: pl can be null at this point, too (player portals off the block).  Just a quick null check will prevent exceptions.

PR #182, opened 2017-03-29
Drop Item - REVIEW ONLY Do not merge

ogmage78:
This runs and works.   If you want to test drop item do the following:

Log in 
@ctw me  (This will create a training wand in your backpack)
OR
@ctw ground (not very interesting - does the same thing but spawns it on the ground.

Once you have the wand in your backpack from calling the first command - drop that bad boy.   Down she goes with a thud and your radar lights up.   Up until about an hour ago, he would do the correct animation and put the damn wand on the ground.   I fell into git hell and between squash and revert - I am too tired to fix it.   You can see it on the ground.   I did not finish the complementary action to be able to pick it up.

The problem is - something I did is messing up the packet stream.   It looks like it all works, but I can't open the pcap file it creates.   That tells me something is amiss.   Hopefully someone can see it.

ALSO

Note - I started this a few weeks ago and I have learned a lot since then.   If any of this is in the wrong namespace or I have made some other noobie mistake - it is not by design it is just a lack of knowledge on my part.    Let me know the better way to do it and you won't have to tell me twice.   :)

### DIFF
--- a/Source/ACE.Entity/Position.cs
+++ b/Source/ACE.Entity/Position.cs
@@ -100,7 +100,7 @@
             {
                 payload.Write(Facing.Y);                
             }
-            
+
             if ((updatePositionFlags & UpdatePositionFlag.Placement) != 0)
             {
                 // TODO: this is current animationframe_id when we are animating (?) - when we are not, how are we setting on the ground Position_id.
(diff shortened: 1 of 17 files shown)


ogmage78:
In looking at this this morning - I think I see what the issue is.   I was working on this right before someone was doing some work on animation.   I had started down a path that I needed to create a properly formed F74C MovementEvent.   He and I both created objects and enums that were the same but named differently.   Both are in here and I am now sending a bogus F74 message.   That is why my neckid dude will not bend down to drop the wand.   I won't be able to untangle this until later today - but anyone that has time, other than that message being fubar - the implementation of the container inventory dictionary, the landblock logic for dropping an item - it would be helpful to know if I am on the right track with how it is implemented.   Thanks

ogmage78:
This is fubar.   Going to fix it and resubmit.

PR #183, opened 2017-03-29
all the database things!

Mogwai-AC:
From PCAPs..
* All weenie objects
* All static objects


LtRipley36706:
Getting several crashes with the current db load in, needs a bit more pruning of objects as there are lots of player created objects in this db in addition to static objects, which probably leads to crashing in some of housing settlements or towns

Mogwai-AC:
Ok, I'll do the rebase without all the changes and go from there.

PR #185, opened 2017-03-30
Network updates for packet/fragment reorder, fragment combining

Zegeger:
Added network session ID to WorldManager to lookup user's sessions and updated related components to handle this.
Refactored GameAction handling to no longer used Activator instantiated objects, they are now delegates like the game message handlers.
Added packet/fragment reorder queues and handling.
Added fragment recombining.
A complete fragment is sent upstream as a new ClientMessage object. Updated all locations where ClientPacketFragment was used.
TODO checksum verification, improve some logging/comments


Source/ACE/Network/NetworkSession.cs
```diff
@@ -164,59 +191,149 @@ private void ProcessPacket(ClientPacket packet)
...
+            ClientMessage message = null;
+            // Check if this fragment is split
             if (fragment.Header.Count != 1)
             {
-                lock (partialFragments)
+                // Packet is split
+                log.DebugFormat("[{0}] Fragment {1} is split, this index {2} of {3} fragments", session.Account, fragment.Header.Sequence, fragment.Header.Index, fragment.Header.Count);
+                MessageBuffer buffer = null;
+
+                // partialFragments is only ever used here in the main update loop, so no need to lock.
```
Zegeger: This is not true, need to lock this actually, as this is coming in on ProcessPacket

Source/ACE/Network/NetworkSession.cs
```diff
@@ -164,59 +191,149 @@ private void ProcessPacket(ClientPacket packet)
...
+                // First check if this message is the next sequence, if it is not, add it to our outOfOrderFragments
+                if (fragment.Header.Sequence == lastReceivedFragmentSequence + 1)
+                {
+                    log.DebugFormat("[{0}] Handling fragment {1}", session.Account, fragment.Header.Sequence);
+                    HandleFragment(message);
+                }
+                else
+                {
+                    log.DebugFormat("[{0}] Fragment {1} is early, lastReceivedFragmentSequence = {2}", session.Account, fragment.Header.Sequence, lastReceivedFragmentSequence);
+                    // outOfOrderFragments is only ever touched in the main Update loop, so no need to lock.
```
Zegeger: This is not true, need to lock this actually, as this can come in on ProcessPacket

Source/ACE/Managers/WorldManager.cs
```diff
@@ -10,14 +10,17 @@
...
 {
     public static class WorldManager
     {
-        // commented: unused.  uncomment if you'll use it.
-        // private static uint sessionTimeout = 150u; // max time between packets before the client disconnects
+        private static readonly ILog log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
-        private static readonly List<Session> sessionStore = new List<Session>();
+        public const ushort ServerId = 0xB;
+        private static Session[] sessionMap = new Session[128]; // TODO Placeholder, should be config MaxSessions
```
Mogwai-AC: can you comment on what the array does?  it's not obvious to me.

Source/ACE/Managers/WorldManager.cs
```diff
@@ -10,14 +10,17 @@
...
 namespace ACE.Managers
 {
     public static class WorldManager
     {
-        // commented: unused.  uncomment if you'll use it.
-        // private static uint sessionTimeout = 150u; // max time between packets before the client disconnects
+        private static readonly ILog log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
-        private static readonly List<Session> sessionStore = new List<Session>();
+        public const ushort ServerId = 0xB;
```
Mogwai-AC: Why did you chose 0xB?  Mostly curious, but this feels like a config thing to me for people that want to host multiple worlds.

Source/ACE/Network/NetworkSession.cs
Mogwai-AC: I'll take your word for it :)

Source/ACE/Managers/WorldManager.cs
Zegeger: It was already B, just hard coded buried somewhere.  It honestly doesn't matter what it is currently as we are single process.  If we ever move to a multi-process/multi-server, then this would need to be changed to allow for that.  I can add a comment to that effect.

Source/ACE/Managers/WorldManager.cs
Zegeger: During login, the client can be assigned a unique Id, which is then sent with every packet that client sends.  The array ensures we are allocating available ID's and reserving them for that specific session.  We in turn leverage that to find sessions by a single int array lookup for each packet processed, instead of iterating the full session list doing IpEndPoint comparison.

PR #186, opened 2017-03-30
Animation - fix and enhancement

ogmage78:
In this PR - I extended on the work Zegeger did.   His work was for animations only.   The system supports a 0 to N animation sequence.   If 0, you are doing a basic motion.   Moving forward, sidestep, turnto.   They use the same message F74C but are intelligently coded depending on what you want to do.    
TODO: I have not looked to see if you can have a motion followed by an animation(s)   I coded it so it supports it, but I did not have a ready pcap to confirm.   Easy enough to fix if they are mutually exclusive.

I cleaned up some emums.   Still to do is to refactor the autonomous position message - I think we have some overlap and duplication there.    

To test - use the debug command @amimation <animation> to see that Zegeger's work still functions as expected.   

Use debug command @movement to see the drop sequence.   You can look at the produced pcaps and see the delta and that they match live servers. 

As usual - any coding faxpas are not by design.   Feedback is wanted and welcome.   If there is a better way to do what I did - please let me know.


Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -177,6 +177,17 @@ public static void animation(Session session, params string[] parameters)
             session.Network.EnqueueSend(new GameMessageMotion(session.Player, session, (MotionCommand)animationId, 1.0f));
         }
 
+        [CommandHandler("movement", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0)]
+        public static void Movement(Session session, params string[] parameters)
+        {
+            var movement = new MovementData { ForwardCommand = 24, MovementStateFlag = MovementStateFlag.ForwardCommand};
+
+            session.Network.EnqueueSend(new GameMessageMotion(session.Player, session, MotionActivity.Idle, MovementTypes.Invalid, MotionFlags.None, MotionStance.Standing,movement));
```
Mogwai-AC: 1, you've got some inconsistent whitespace in this function.  2, can you document what this function does?  It's non-obvious to me.

Source/ACE/Network/GameMessages/Messages/GameMessageMotion.cs
```diff
@@ -13,40 +17,46 @@ public GameMessageMotion()
...
+            Writer.Write((ushort)stance); // called command in the client
+
         }
 
-        private void WriteAnimations(Entity.WorldObject animationTarget, List<MotionItem> items)
+        private void WriteAnimations(WorldObject animationTarget, List<MotionItem> items)
         {
-            uint generalFlags = (uint)items.Count << 7;
-            Writer.Write((uint)generalFlags);
+            if (animationTarget == null) throw new ArgumentNullException(nameof(animationTarget));
```
Mogwai-AC: throwing here is probably not the best idea.  should probably just make a log warning entry and return.

ogmage78:
I made the requested changes.   Let me know if you see anything else.

Source/ACE/Network/GameMessages/Messages/GameMessageMotion.cs
```diff
@@ -46,15 +53,18 @@ public GameMessageMotion(Entity.WorldObject animationTarget, Session session, Mo
...
             Writer.Write((ushort)stance); // called command in the client
 
         }
 
         private void WriteAnimations(WorldObject animationTarget, List<MotionItem> items)
         {
-            if (animationTarget == null) throw new ArgumentNullException(nameof(animationTarget));
+            if (animationTarget == null)
+            {
+                Log("We have a null for animationTarget - that is wrong, wrong wrong.");
```
Mogwai-AC: great, now short circuit the function before you throw a null ref exception!

ogmage78: Oops.   Can't I am rusty can you 😛   Will fix as soon as I get home.   Thanks for reviewing for me

ogmage78:
Done.

PR #187, opened 2017-03-30
World fix and fireball test

Forbiddenz:
World Object Fixes + My test Spell firewall - sorta.. so I could test it some more.
New AceVector Object.


Source/ACE.Entity/Vector3.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.IO;
+using System.Numerics;
+
+namespace ACE.Entity
+{
+    public class AceVector3
```
Mogwai-AC: 3 private floats, that you can't get out once set?  What's the point?  What does this object give us that the existing Vector3 object doesn't?

Source/ACE/Factories/SpellFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    /// <summary>
+    /// factory class for creating objects related to administration or content-recreation
+    /// </summary>
+    public class SpellFactory
+    {
+        /// <summary>
+        /// creates a spell with velocity directly in front of the player's position provided
+        /// </summary>
+        public static ImmutableWorldObject CreateCastedSpell(ushort weenieClassId, Position newPosition, ObjectGuid objectid, AceVector3 velocity)
```
Mogwai-AC: Spells are very much mutable.  They have motion and do things, so use MutableObject.

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -195,5 +196,23 @@ public static void CreatePortal(Session session, params string[] parameters)
...
+
+        [CommandHandler("testspell", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 3)]
+        public static void TestSpell(Session session, params string[] parameters)
+        {
+            float x = float.Parse(parameters[0], CultureInfo.InvariantCulture.NumberFormat);
+            float y = float.Parse(parameters[1], CultureInfo.InvariantCulture.NumberFormat);
+            float z = float.Parse(parameters[2], CultureInfo.InvariantCulture.NumberFormat);
+
+            // we got to keep track because we going to send a effect..
+            AceVector3 velocity = new AceVector3(36f, 9.8f, -1.3f);
```
Mogwai-AC: What do these float values represent?  Document them please.

Source/ACE/Entity/PhysicsData.cs
```diff
@@ -30,10 +30,10 @@ public class PhysicsData
         public float Friction;
         public float Elastcity;
         public uint AnimationFrame;
-        public Position Acceleration;
+        public AceVector3 Acceleration;
```
Mogwai-AC: Again, we already had System.Numerics.Vector3.  See previous comments.

Source/ACE/Factories/SpellFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+
+            // all we need right now ..  + required..
+            var weenie = WeenieHeaderFlag.Spell;
+
+
+            ImmutableWorldObject wo = new ImmutableWorldObject(ObjectType.None, objectid, "Ball of Death", weenieClassId, ObjectDescriptionFlag.None, weenie, newPosition);
+
+            // flame bolt
+            wo.GameData.Spell = (ushort)27;
```
Mogwai-AC: Make a spells enum instead of hard-coding 27.  This data existed in Decal before, so I'm sure we can get it now.

Source/ACE/Factories/SpellFactory.cs
```diff
@@ -0,0 +1,44 @@
...
+            wo.PhysicsData.PhysicsDescriptionFlag = PhysicsDescriptionFlag.CSetup | PhysicsDescriptionFlag.Stable | PhysicsDescriptionFlag.Petable | PhysicsDescriptionFlag.Position | PhysicsDescriptionFlag.Velocity | PhysicsDescriptionFlag.Friction | PhysicsDescriptionFlag.Elastcity | PhysicsDescriptionFlag.DefaultScript | PhysicsDescriptionFlag.DefaultScriptIntensity;
+
+            // model id 0x000026 is one of several lifestone IDs
+            wo.PhysicsData.CSetup = (uint)33555469;
+            wo.PhysicsData.Stable = (uint)536870967;
+            wo.PhysicsData.Petable = (uint)872415237;
+            wo.PhysicsData.Velocity = velocity;
+            wo.PhysicsData.Friction = (uint)1f;
+            wo.PhysicsData.Elastcity = (uint)0;
+            wo.PhysicsData.DefaultScript = (uint)90; // this needs a enum
```
Mogwai-AC: I agree, make an enum :p

Source/ACE/Entity/PhysicsData.cs
Forbiddenz: System.Numerics.vector3 does not write packet payloads in the correct order.
But I will add the ability to return a vector.

But AceVector follows the position class, although position is over loaded

Source/ACE.Entity/Vector3.cs
Forbiddenz: Will add ability to get a real vector 3 but it's not what this was intended for but it could be added as a way to retrieve it through a protected function

Forbiddenz: Will add ability to get a real vector 3 but it's not what this was intended for but it could be added as a way to retrieve it through a protected function

Source/ACE/Entity/PhysicsData.cs
Mogwai-AC: Ok, makes sense.  Just give us a way to get the values out after it's been created.  We can let these vectors live longer than 1 send and just update the values they have in them.  Also, physics engine will need to read them out at some point.

Forbiddenz:
Updated

Forbiddenz:
Removed some useless comments I noticed too.
- Spell is now a mutable object

Forbiddenz:
Spells ! We have some fun spells.. I will work to improve it but this is worth it to submit it now as it should works in multiplayer for giggles..

Forbiddenz:
Closing this, I found some issues with it.
I also am thinking we need some kind of naming convention between Entity and Enum..
Starting to Get confusing.. IE
PlayerEntity, SpellEntity, Etc as we build more entitys.. enums should just be Spell, Player, etc as needed.

Issue #144, opened 2017-03-18
Add the Ability to drop an item from your backpack.


ogmage78:
I am working on this at the moment.   Can it be assigned to me?

ogmage78:
If the first part gets checked in - I will implement pickup item - then add saving your inventory to the database for persistence.   I will add inscription next.

PR #189, opened 2017-03-31
Inventory Phase 1 and Drop Item

ogmage78:
This is at a good place for code review and possible merge.   Once approved, this will allow others to start on the numerous tasks around inventory management.  You can spawn a Training wand in your backpack or on the ground.   That is just throw away code to test with.   It is in the debug handler.
usage:
@ctw me OR @ctw ground

Once the item is spawned, you can then drop it on the ground.   

Included in this PR.  Starts to address Issue #144 

World object now has inventory dictionary.   Added this here as anything that can be a container needs to have this ability.   Not all world objects are containers - so also added property IsContainer

Added methods to:

- Add to inventory
- Remove from inventory
- Lookup item in inventory
- Handle Drop item: 
     1. Calls lookup
     2. Update Container burden
     3. Sets some properties in preparation for going 3D
     4. GameMessageUpdatePosition.
     5. Alerts Landblock to new arrival
     6. Removes the item the player's inventory 
Added GameActionDropItem
This handles the client's request to drop an item.
Added GameMessagePutObjectIn3d
 
As usual - any coding faux pas are not by design. Feedback is wanted and welcome. If there is a better way to do what I did - please let me know.


ogmage78:
I will fix this first thing in the morning.   I looked they are easy to resolve.   Thanks for checking it over.

ogmage78:
I fixed the merge issues.   When you get a chance to review this - if you look at: 
public void HandleDropItem(ObjectGuid objectGuid, Session session)   - I am not sure if it needs to be on world object or player.   I cannot remember if any NPC's would drop anything as part of a quest.   I put in in WO just in case.   That might be a mistake.   If it needs to go into player - I need to create a method on world object to update containerId.   Anyway - any feedback is appreciated.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -72,6 +85,10 @@ public ushort ForcePositionIndex
 
         public SequenceManager Sequences { get; }
 
+        public object InventoryMutex => this.inventoryMutex;
```
Mogwai-AC: I'm really hesitant to think a mutex needs to be public.  Red flag for other code being nosey into this class.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -72,6 +85,10 @@ public ushort ForcePositionIndex
...
+
+        public Dictionary<ObjectGuid, WorldObject> Inventory => this.inventory;
```
Mogwai-AC: this should be a ReadOnlyDictionary.   You'll have to new it up in the constructor.  The default for "inventory" may fire after this, rendering this default incorrect.  I'm not exactly sure when the "=>" operator binds on the property, so it may work after all.  If you feel this is correct as is, I'm open to discussion as always.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -87,6 +104,175 @@ protected WorldObject(ObjectType type, ObjectGuid guid)
...
+        /// <summary>
+        /// This is used to do the housekeeping on the server side to take an object from inventory into 3D world then tell the world
+        /// </summary>
+        public void HandlePutItemInContainer(ObjectGuid itemGuid, ObjectGuid containerGuid,  Session session)
+        {
+            lock (this.InventoryMutex)
+            {
+                // Find the item we want to pick up
+                var obj = LandblockManager.GetWorldObject(session, itemGuid);
+                if (Object.ReferenceEquals(null, obj))
```
Mogwai-AC: if (obj == null)

Source/ACE/Entity/WorldObject.cs
```diff
@@ -87,6 +104,175 @@ protected WorldObject(ObjectType type, ObjectGuid guid)
...
+
+                // let's move to pick up the item
+                this.PositionFlag = UpdatePositionFlag.Contact |
+                   UpdatePositionFlag.ZeroQy | UpdatePositionFlag.ZeroQx;
+                this.Position.positionX = obj.Position.positionX;
+                this.Position.positionY = obj.Position.positionY;
+                this.Position.positionZ = obj.Position.positionZ;
+                this.Position.rotationW = obj.Position.rotationW;
+                this.Position.rotationZ = obj.Position.rotationZ;
+
```
Mogwai-AC: whoa, does this actually work?

Source/ACE/Entity/WorldObject.cs
```diff
@@ -87,6 +104,175 @@ protected WorldObject(ObjectType type, ObjectGuid guid)
...
+                session.Network.EnqueueSend(new GameMessageUpdatePosition(this));
+
```
Mogwai-AC: remove large chunks of commented code.  If you need to keep them, feel free to make a blog page on the acemulator wiki.

Source/ACE/Managers/LandblockManager.cs
```diff
@@ -61,8 +61,17 @@ public static void HandleQueryHealth(Session source, ObjectGuid targetId)
         }
 
         /// <summary>
+        /// return wo in preparation to take it off the landblock and put it in a container.
+        /// </summary>
+        public static WorldObject GetWorldObject(Session source, ObjectGuid targetId)
```
Mogwai-AC: ugh... we really need that action queue...

Source/ACE/Entity/WorldObject.cs
ogmage78: Fixed.   I got confused there -

Source/ACE/Entity/WorldObject.cs
ogmage78: Fixed

Source/ACE/Entity/WorldObject.cs
ogmage78: Fixed

Source/ACE/Entity/WorldObject.cs
ogmage78: Fixed - I did not see I pushed that.   Comments gone.   Sorry about that.

PR #191, opened 2017-04-02
Fixed bug / incomplete implementation of instance_time stamp

ogmage78:
Small fix, but an important one.   A lot of messages take an instance_sequence or timestamp.   it is called timestamp in the client.   I fixed the places where this was still being set to 1.   If the instance_timestamp does not agree between calls it can cause you to hang in portal space or other bad things.   The PhysicsData.Position structure passes 9 (0-8 ) unnamed timestamps.   These were in place - two were set to 0 - however, they were out of order.   I am not sure of the other 8, but the last one is the instance_timestamp.   I verified this with live server pcaps.   I have made it so with what we have implemented so far, we are consistent.    It will be important that we get this information out to all contributors as we need to set it for all player messages to totalLogins.   I also make the character load and save properly update the totalLogins value.


ogmage78:
No conflicts now and this does fix a serious bug.   Let me know what you think.

ogmage78:
Fixed last conflict

PR #184, opened 2017-03-30
Feature: ORM Driven Character Positions for Issue #166

fantoms:
This huge PR is related to Issue #166. The code has been changed to allow more then one position type. A class was added for Character position types that match the requirements listed in the issue. A secondary Position Type class was added to match the client, provided by @LtRipley36706. 

To work with the ORM code introduced during the developer meeting, I've attempted to copy the pseudo 1:1 mapping into `CharacterPosition.cs`. In order to allow for more then one position, the `Character.Position` member was changed to dictionary containing a list of character positions by type. `Player.Position` was changed to `Player.PhysicalPosition` and when the server decides to save the positions to the database, the data is saved through the ORM object in `CharacterPosition.cs`. 

There were a few things left to do, that were not accomplished for Issue #166:
* Hook up /lifestone and /ls commands

ChangeLog:
* Changed the Player.Position to Player.PhysicalPosition to match the Developer meeting video and Issue #166.
* Created `CharacterPostion` and `CharacterPositionType` classes to store the Database types requested in Issue #166.
* Added `CharacterPostion` ORM logic.
* Built out helper extension functions to serve the `StartingLocation`, until we pull from the database. Also included an Invalid location as a default (0).
* Changed the `character_positions` table and Added `positionType` column.
* Set the primary key to include `positionType` column.
* Added `positionType` to required Critera for `character_positions`.
* Added `CharacterPositionInsert` and `CharacterPositionUpdate` to the Character Database.
* Created `CharacterPositionInsert` and `CharacterPositionUpdate` prepared SQL statements by utilizing the `ConstructStatement` function.
* Added the required `positionType` clause to the `CharacterPositionSelect` prepared statement.
* Added a position update statement for updating the player position.
* Reworked GetPosition to include the `CharacterPositionType`.
* Added function for getting a `CharacterPostion` from the database, requires a character Id and `CharacterPositionType`.
* Added logic to get location function to insert a location into the database, if not present.
* Reworked the UpdateCharacter and Update functions to save positions.
* Added functions for inserting new character positions on character creation.
* Added functions for saving/loading the character positions from the database on joining the World.
* Attempted to get log4net working properly and moved some code to logging.
* Added DBDEBUG project constant for debugging the database calls.
* Changed the `CharacterPositionSelect` prepared statement to a `ConstructStatement`.

GitLog:
- Changed position to phsysical/logical and added logical positions
- Updated comments on CharacterPositionType
- Added the sql for altering the positions table, to add the positionType
[...]

PR #194, opened 2017-04-02
Fix: Reverted SQL syntax to use AFTER, instead of BEFORE

fantoms:
- Changed `2017_03_27_character_positions_change.sql` and `2017_03_26_character_positions_positiontype.sql` to use AFTER instead of BEFORE, when aligning the column.  The `positionType` column has been moved after the `id` column.


fantoms:
In my defense, here is the previously requested change: https://github.com/ACEmulator/ACE/pull/184/files/0fda5e079b51c574b0565d7e01fa51e8548d5d0b

PR #195, opened 2017-04-02
World fix and fireball test

Forbiddenz:
Clean up - Synced with Latest Prs


Source/ACE/Entity/PhysicsData.cs
```diff
@@ -30,10 +30,10 @@ public class PhysicsData
         public float Friction;
         public float Elastcity;
         public uint AnimationFrame;
-        public Position Acceleration;
+        public AceVector3 Acceleration;
         public float Translucency;
-        public Position Velocity;
-        public Position Omega; // rotation
+        public AceVector3 Velocity;
+        public AceVector3 Omega; // rotation
```
fantoms: In `PhysicsData.cs` line 51, please consider if the constructor needs to have some base values:
```
      public PhysicsData()
        {
            Velocity = new AceVector3(0.0f, 0.0f, 0.0f);
            Omega = new AceVector3(0.0f, 0.0f, 0.0f);
        }
``` 

It is worth checking because I had been getting a crash where Velocity was null, when a character was logging into the server.

Source/ACE/Entity/SpellEntity.cs
```diff
@@ -0,0 +1,58 @@
...
+using ACE.Network.Enum;
+using ACE.Network.GameMessages.Messages;
+using log4net;
+
+namespace ACE.Entity
+{
+    /// <summary>
+    /// Spell Entity
+    /// </summary>
+    public class SpellEntity :  MutableWorldObject
```
Mogwai-AC: This needs a better name.  Is there a particular reason a spell gets an entire class definition, though?  The only difference I see here is that it has a PlayScript implementation that creates an Effect.  There's nothing about the class definition that tells me it's a spell.  I think this could have been done in MutableWorldObject as is, with all the Physics values being passed into a constructor, and then checking the PhysicsData.PhysicsState in the PlayScript method.

Also, I recall seeing a video of multiple effects.  Was that not yours?  Is that implemented here?

Source/ACE/Factories/SpellObjectFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+        public static WorldObject CreateSpell(uint templateId, Position position, AceVector3 velocity, float friction, float electicity)
+        {
+
+            Entity.SpellEntity wo = new Entity.SpellEntity(position, velocity, friction, electicity);
+
+            // todo: impletment more advanced templating.
+            switch (templateId)
+            {
+                case 0:
+                    wo.SpellID = Spell.FlameBolt;
```
Mogwai-AC: yeah, as I said above, these values should be pushed into the MutableWorldObject constructor instead of being replaced after instantiation.

Source/ACE/Entity/SpellEntity.cs
Forbiddenz: I wanted to keep it separate because I am not sure yet what all it will needed,
But PlayScript may have to be far more advanced as we learn more.

I don't want to put all that logic in all mutable objects, I think it would create a mess as things progress.  for now I thinks its best to keep them separated.

And if we find other objects with different code then PlayScript, we can then make it more generic if we keep them separated.  Code will be a little bigger but much easier to read in the long run as we have different kinds of Entities.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -19,7 +19,7 @@ public abstract class WorldObject
         /// <summary>
         /// wcid - stands for weenie class id
         /// </summary>
-        public ushort WeenieClassid { get; protected set; }
+        public ushort WeenieClassid { get; set; }
```
Mogwai-AC: I told you repeatedly to not do this.

Source/ACE/Factories/SpellObjectFactory.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <summary>
+    /// Spell Factory for creating objects related to Spells.
+    /// </summary>
+    public class SpellObjectFactory
+    {
+        public static WorldObject CreateSpell(uint templateId, Position position, AceVector3 velocity, float friction, float electicity)
+        {
+
+            ushort weenieClassId = 0;
+            MutableWorldObject mobj = new MutableWorldObject(ObjectType.MissileWeapon, new ObjectGuid(CommonObjectFactory.DynamicObjectId, GuidType.None), "Spell", weenieClassId, ObjectDescriptionFlag.None, WeenieHeaderFlag.Spell, position);
```
Mogwai-AC: pull your switch block above this to find out what weenieClassId you need and pass it in.  We discussed this several times.

PR #198, opened 2017-04-03
Add portals to ACE_World tables

kdmxp256:
* Add a fix to the current master ACE_World tables for weenie and ace objects to loosen the index restrictions
* Add Portal weenie table data
* Add Portal object table data

---  Duplicate portal objects removed; Floating City, Orphanage, Golem Sanctum, and Humming Crystal portals removed as they are not static object placements


Mogwai-AC:
Any chance you can concat the portal files down to a single file for purposes of the PR?  In the script directory, you should be able to do this (on Windows command line):
copy *.sql AllPortals.sql

We'd just need to know to run that via the SqlYog import tool rather than the query editor.

PR #211, opened 2017-04-06
WIP Inventory - Good place for review - Do not merge yet

ogmage78:
With help from Zegeger - found the sequence issue that was keeping item pick up from working.   Not for merge - just posting for review.   You can create, pickup and drop.   There is a bug in a pick up then drop again - the item takes off.   Either still have a sequence or a position flag issue on multiple drops and pickups.


ogmage78:
I regressed with my latest changes.   Going to close this for now.

PR #213, opened 2017-04-06
Added lifestone attunment, @ls recall, and position fixes.

fantoms:
Also worked out the log4net problems for debugging databases


Source/ACE.Database/Database.cs
```diff
@@ -78,8 +84,8 @@ public void AddPreparedStatement<T>(T id, params object[] parameters)
                 }
                 catch (MySqlException transactionException)
                 {
-                    Console.WriteLine($"An exception occured while commiting a transaction of {queries.Count} queries, a rollback will be performed!");
-                    Console.WriteLine($"Exception: {transactionException.Message}");
+                    log.Error($"An exception occured while commiting a transaction of {queries.Count} queries, a rollback will be performed!");
```
fantoms: Should this be set to Error channel? I believe it's more relevant to capture these exceptions as errors, if different logging levels are being utilized.

Source/ACE/Network/GameAction/Actions/GameActionUseItem.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (landBlockObject != null) {
+                foreach (uint lsType in System.Enum.GetValues(typeof(LifestoneWeenieClasses)))
+                {
+                    if (landBlockObject.WeenieClassId == lsType)
+                    {
+                        float dx = Math.Abs(landBlockObject.Position.PositionX - session.Player.Location.PositionX);
+                        float dy = Math.Abs(landBlockObject.Position.PositionY - session.Player.Location.PositionY);
+                        float dz = Math.Abs(landBlockObject.Position.PositionZ - session.Player.Location.PositionZ);
+
+                        // check if the distance is greater then
```
Mogwai-AC: This should use 3D pythagorean math, not checks on individual axis components.  It will work, but it's logically incorrect.

Source/ACE/Network/GameAction/Actions/GameActionUseItem.cs
```diff
@@ -0,0 +1,66 @@
...
+        [GameAction(GameActionType.EvtInventoryUseEvent)]
+        public static void Handle(ClientMessage message, Session session)
+        {
+            // this is specifically for lifestones:
+            // get the object id
+            uint fullId = message.Payload.ReadUInt32();
+            var objectId = new ObjectGuid(fullId);
+
+            // get the worldobject from the id, if it exists
+            AceObject landBlockObject = DatabaseManager.World.GetObjectsByLandblock(session.Player.Location.LandblockId.Landblock).Find(o => o.AceObjectId == objectId.Full);
```
Mogwai-AC: I applaud the effort here.  It's the best you could do with what you have.  We really just need that action queue - stuff like this isn't going to fly in the long run.

PR #217, opened 2017-04-07
Reveresed the logic for the distance check on lifestone usage

fantoms:
- After performing some empirical visual client testing with:
 `
float squaredDistance = player.Location.SquaredDistanceTo(obj.Location);
                                        player.Session.Network.EnqueueSend(new GameMessageSystemChat(squaredDistance.ToString(), ChatMessageType.Broadcast));
`, I found that a static use distance of at least `4.9f` worked _well_. I did not change this float, but I am noting it here.

- I also changed the `Lifestone Use` text to blue.

### DIFF
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -431,7 +431,7 @@ namespace ACE.Entity
                                         // validate within use range
                                         float radiusSquared = obj.GameData.UseRadius * obj.GameData.UseRadius;
 
-                                        if (player.Location.SquaredDistanceTo(obj.Location) <= radiusSquared)
+                                        if (player.Location.SquaredDistanceTo(obj.Location) >= radiusSquared)
                                         {
                                             serverMessage = "You wandered too far to attune with the lifestone!";
                                         }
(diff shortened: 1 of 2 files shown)


Source/ACE/Entity/Landblock.cs
```diff
@@ -431,7 +431,7 @@ private void HandleGameAction(QueuedGameAction action, Player player)
                                         // validate within use range
                                         float radiusSquared = obj.GameData.UseRadius * obj.GameData.UseRadius;
 
-                                        if (player.Location.SquaredDistanceTo(obj.Location) <= radiusSquared)
+                                        if (player.Location.SquaredDistanceTo(obj.Location) >= radiusSquared)
```
Mogwai-AC: Think about what this now means... If the player is FURTHER away than the object radius...

The "<=" was correct, it just wasn't taking the size of the lifestone into account.  If you add a fudge factor of 1.5 to the useRadius, this should correct the issue.  Here's an example:

float modelSize = 1.5f;
float radiusSquared = (modelSize + obj.GameData.UseRadius) * (modelSize + obj.GameData.UseRadius);

PR #224, opened 2017-04-10
Inventory Part 2 - Create, Drop and Pickup

ogmage78:
This is a minor extension of the work I was doing previously, but it did turn up a number of good to know items.   

1.  When you are dropping an item, you have to increment the portal sequence - the client treats items being added to 3D space like a teleport.   It is on overloaded flag.   God, do I hate that.

1. There was some optimization code in player that was checking "check for a short circuit.  if we don't need to update, don't!"   This has a bug in it that I could not quite fix.   I commented it out for now with a big TODO.   It never sends an update on object create or update after the initial object creation.   We are missing updating the LastUpdatedTicks somewhere.

1. In Landblock - I added a lambda expression to omit sending the stop tracking message if we are the one putting the item into our container.   Everyone else needs to forget about it, but it is still mine. 

Next steps:

- [ ] Put in code to move to the object being picked up.

- [ ] Check for success and respond to item no longer there

- [ ] Save off player inventory items to the database

- [ ] Load inventory items on player load.

### DIFF
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -157,8 +157,9 @@ namespace ACE.Entity
 
             lock (objectCacheLocker)
             {
-                allObjects = this.worldObjects.Values.ToList();
-                this.worldObjects[wo.Guid] = wo;
+                allObjects = worldObjects.Values.ToList();
+                if (!worldObjects.ContainsKey(wo.Guid))                                
+                    worldObjects[wo.Guid] = wo;                
             }
 
             var args = BroadcastEventArgs.CreateAction(BroadcastAction.AddOrUpdate, wo);
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -281,7 +281,7 @@ namespace ACE.Entity
             {
                 case BroadcastAction.Delete:
                     {
-                        Parallel.ForEach(players, p => p.StopTrackingObject(wo.Guid));
+                        Parallel.ForEach(players, p => p.StopTrackingObject(wo));
                         break;
                     }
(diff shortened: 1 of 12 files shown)


ogmage78:
Forgot to give a shoutout to Lidefeath for helping me with the lambda expression.  ;)

Source/ACE/Entity/Landblock.cs
```diff
@@ -157,8 +157,9 @@ public void AddWorldObject(WorldObject wo)
 
             lock (objectCacheLocker)
             {
-                allObjects = this.worldObjects.Values.ToList();
-                this.worldObjects[wo.Guid] = wo;
+                allObjects = worldObjects.Values.ToList();
+                if (!worldObjects.ContainsKey(wo.Guid))
```
Mogwai-AC: this wasn't at all necessary.  Dictionaries treat myDictionary[key] as an "add or set".

Source/ACE/Entity/WorldObject.cs
```diff
@@ -147,96 +147,121 @@ public void RemoveFromInventory(ObjectGuid objectGuid)
...
+                new GameMessagePrivateUpdatePropertyInt(session,
+                    PropertyInt.EncumbVal,
+                    (uint)session.Player.GameData.Burden));
+            GeneralMotion motion = new GeneralMotion(MotionStance.Standing, new MotionItem((MotionCommand)MovementTypes.General));
+            motion.MovementData.ForwardCommand = 24;
+            session.Network.EnqueueSend(new GameMessageUpdateMotion(session.Player, session, motion));
+
+            // Set Container id to 0 - you are free
+            session.Network.EnqueueSend(
+                new GameMessageUpdateInstanceId(objectGuid, targetContainer));
```
Mogwai-AC: odd place to split a line, yo.

PR #226, opened 2017-04-10
Fix for Animations

LtRipley36706:
Not sure why this sequence was changed, but this fixes the issue.


fantoms:
After pulling this patch, I was able to perform animations on different integers.

PR #227, opened 2017-04-10
Fix GameActionType.QueryHealth for targets in adjacent landblocks and…

Lidefeath:
… remove the old HandleUpdateHealth code which somehow was still in master.


Mogwai-AC:
changelog...

Lidefeath:
Fixed changelog.md

PR #229, opened 2017-04-11
Update to use Action Queue - Requested Changes Made 4/14/2017

ogmage78:
Can someone with more experience with Action Queue take a look at the file Landblock and the  case GameActionType.DropItem:       I don't want to refactor all of my stuff until someone can tell me if this looks correct.   It complies and works fine.   I would just like a sanity check.

### DIFF
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -444,13 +444,52 @@ namespace ACE.Entity
                         HandleSoundEvent(obj, soundEffect);
                         break;
                     }
-                case GameActionType.DropItem:                  
+                case GameActionType.PutItemInContainer:
+                    {
+                        var playerId = new ObjectGuid(action.ObjectId);
+                        var inventoryId = new ObjectGuid(action.SecondaryObjectId);
+                        if (playerId.IsPlayer())
+                        {
+                            Player aPlayer = null;
+                            WorldObject inventoryItem = null;
 
-                    {                        
-                        var g = new ObjectGuid(action.ObjectId);                        
+                            if (worldObjects.ContainsKey(playerId))
+                            {
+                                aPlayer = (Player)worldObjects[playerId];
+                                inventoryItem = worldObjects[inventoryId];                                
+                            }
+
+                            if ((aPlayer != null) && (inventoryItem != null))
+                            {                                
+                                var motion = new GeneralMotion(MotionStance.Standing);
+                                motion.MovementData.ForwardCommand = 24;                                
+                                aPlayer.Session.Network.EnqueueSend(new GameMessageUpdatePosition(aPlayer), 
+                                    new GameMessageUpdateMotion(aPlayer, aPlayer.Session, motion),
+                                    new GameMessageSound(aPlayer.Guid, Sound.PickUpItem, (float)1.0));
+                                
+                                // Add to the inventory list.
+                                aPlayer.AddToInventory(inventoryItem);
+
+                                motion = new GeneralMotion(MotionStance.Standing);
+                                aPlayer.Session.Network.EnqueueSend(new GameMessagePrivateUpdatePropertyInt(aPlayer.Session,
+                                       PropertyInt.EncumbVal,
+                                       aPlayer.GameData.Burden), 
+                                       new GameMessagePutObjectInContainer(aPlayer.Session, aPlayer, inventoryId),
+                                       new GameMessageUpdateMotion(aPlayer, aPlayer.Session, motion),
+                                       new GameMessageUpdateInstanceId(inventoryId, playerId),
+                                       new GameMessagePickupEvent(aPlayer.Session, inventoryItem));
+ 
+                                LandblockManager.AddObject(inventoryItem);
+                            }
+                        }
+                        break;
+                    }
+                case GameActionType.DropItem:
+                    {
+                        var g = new ObjectGuid(action.ObjectId);
                         // ReSharper disable once InconsistentlySynchronizedField
                         if (worldObjects.ContainsKey(g))
-                        {                         
+                        {
                             var playerId = new ObjectGuid(action.ObjectId);
                             var inventoryId = new ObjectGuid(action.SecondaryObjectId);
                             if (playerId.IsPlayer())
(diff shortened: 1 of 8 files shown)


Source/ACE/Entity/Landblock.cs
```diff
@@ -19,6 +19,11 @@
 
 namespace ACE.Entity
 {
+    using global::ACE.Entity.Enum.Properties;
+    using global::ACE.Network.Sequence;
+
+    using static LandblockManager;
```
Mogwai-AC: ????  This makes for non-obvious inline code.  Please no.

Source/ACE/Entity/Landblock.cs
```diff
@@ -19,6 +19,11 @@
 
 namespace ACE.Entity
 {
+    using global::ACE.Entity.Enum.Properties;
```
Mogwai-AC: move above namespace please.

Source/ACE/Entity/Landblock.cs
```diff
@@ -467,6 +472,90 @@ private void HandleGameAction(QueuedGameAction action, Player player)
...
+                        // ReSharper disable once InconsistentlySynchronizedField
+                        if (worldObjects.ContainsKey(g))
+                        {                         
+                            var playerId = new ObjectGuid(action.ObjectId);
+                            var inventoryId = new ObjectGuid(action.SecondaryObjectId);
+                            if (playerId.IsPlayer())
+                            {
+                                Player aPlayer = null;
+                                WorldObject inventoryItem = null;
+                                lock (objectCacheLocker)
```
Mogwai-AC: You know, I don't think you actually need to lock up the landblock's mutex here.  You're not modifying the object collection in this code block.

Source/ACE/Entity/Landblock.cs
```diff
@@ -467,6 +472,90 @@ private void HandleGameAction(QueuedGameAction action, Player player)
...
+                                    if (worldObjects.ContainsKey(playerId))
+                                    {
+                                        aPlayer = (Player)worldObjects[playerId];
+                                        inventoryItem = aPlayer.GetInventoryItem(inventoryId);
+                                        aPlayer.RemoveFromInventory(inventoryId);
+                                    }
+                                }
+                                
+                                // We are droping the item - let's keep track of change in burden
+
```
Mogwai-AC: Removing the item from the player's inventory should do all this, not the landblock.

Source/ACE/Entity/Landblock.cs
```diff
@@ -467,6 +472,90 @@ private void HandleGameAction(QueuedGameAction action, Player player)
...
+                                    // Send the container update again.   I have no idea why, but that is what they did in live.
+                                    aPlayer.Session.Network.EnqueueSend(
+                                        new GameMessagePutObjectIn3d(aPlayer.Session, aPlayer, inventoryId),
+                                        new GameMessageSound(aPlayer.Guid, Sound.DropItem, (float)1.0),
+                                        new GameMessageUpdateInstanceId(inventoryId, targetContainer));
+                                    // Set the flags and determine a position.
+
+                                    // TODO: I need to look at the PCAPS to see the delta in position from the dropper and the item dropped.   Temp position.
+                                    
+                                    inventoryItem.PositionFlag = UpdatePositionFlag.Contact
```
Mogwai-AC: modification of the inventoryItem's position/whatever data should stay here in the landblock, yes.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -51,7 +51,7 @@ public virtual Position Location
 
         public WeenieHeaderFlag2 WeenieFlags2 { get; protected set; }
 
-        public UpdatePositionFlag PositionFlag { get; protected set; } = UpdatePositionFlag.Contact;
+        public UpdatePositionFlag PositionFlag { get; set; } = UpdatePositionFlag.Contact;
```
Mogwai-AC: Um....  Not a huge fan of this.  This isn't the way to solve the always-falling-other-players bug.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -122,89 +122,30 @@ public void SetMotionState(MotionState motionState)
             p.Session.Network.EnqueueSend(updateMotion);
         }
 
-        public void AddToInventory(WorldObject worldObject)
+        // Inventory Management Functions
+        public virtual void AddToInventory(WorldObject inventoryItem)
         {
-            lock (this.inventoryMutex)
+            lock (inventoryMutex)
```
Mogwai-AC: also modify burden here

Source/ACE/Entity/WorldObject.cs
```diff
@@ -122,89 +122,30 @@ public void SetMotionState(MotionState motionState)
...
-                    return;
-
-                inventory.Add(worldObject.Guid, worldObject);
+                if (!this.inventory.ContainsKey(inventoryItem.Guid)) this.inventory.Add(inventoryItem.Guid, inventoryItem);
             }
         }
 
-        public void RemoveFromInventory(ObjectGuid objectGuid)
+        public virtual void RemoveFromInventory(ObjectGuid inventoryItemGuid)
         {
```
Mogwai-AC: also modify burden here

Source/ACE/Entity/Landblock.cs
ogmage78: Just making sure I understand the issue.   It looks like I must have added something in landblock that pulled in a reference to landblockmanager.   That is the issue right?

Mogwai-AC: You used some automatic tool that did the "using static" stuff which is new in .Net 6 and that quite frankly, I absolutely think is a bad thing they did.  It's akin to VB6's "with" statement, only scoped to the entire file.

ogmage78: I did not even notice it.   Thanks and I will clean that up - along with the other items.

ogmage78:
Updated and ready for merge upon approval.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -122,156 +122,45 @@ public void SetMotionState(MotionState motionState)
...
-
-        public void RemoveFromInventory(ObjectGuid objectGuid)
+        // Inventory Management Functions
+        public virtual void AddToInventory(WorldObject inventoryItem)
         {
             lock (inventoryMutex)
             {
-                if (inventory.ContainsKey(objectGuid))
-                    inventory.Remove(objectGuid);
+                if (!inventory.ContainsKey(inventoryItem.Guid)) inventory.Add(inventoryItem.Guid, inventoryItem);
```
Mogwai-AC: whitespace - logic on different line than condition.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -122,156 +122,45 @@ public void SetMotionState(MotionState motionState)
...
-            session.Network.EnqueueSend(
-                new GameMessageSound(session.Player.Guid, Sound.PickUpItem, (float)1.0));
-
-            // Add to the inventory list.
-            lock (this.inventoryMutex)
+            lock (inventoryMutex)
             {
-                if (!this.inventory.ContainsKey(obj.Guid)) 
-                    this.inventory.Add(obj.Guid, obj);
+                if (!this.inventory.ContainsKey(inventoryItemGuid)) return;
```
Mogwai-AC: as before.  no inline ifs.

Source/ACE/Entity/WorldObject.cs
ogmage78: Ok got rid of inline if's, think I fixed the whitespace - I may need remedial training on that to make sure I understand our style.   I fixed my last bug and think it is all ready to go.

Source/ACE/Entity/Landblock.cs
```diff
@@ -442,6 +442,99 @@ private void HandleGameAction(QueuedGameAction action, Player player)
...
+                        var inventoryId = new ObjectGuid(action.SecondaryObjectId);
+                        if (playerId.IsPlayer())
+                        {
+                            Player aPlayer = null;
+                            WorldObject inventoryItem = null;
+
+                            if (worldObjects.ContainsKey(playerId))
+                            {
+                                aPlayer = (Player)worldObjects[playerId];
+                                inventoryItem = worldObjects[inventoryId];
```
Zegeger: We can't assume inventoryId is in this landblock.  We need something like FindWorldObject which walks this and adjacent landblocks to find it.

ogmage78:
Stackoverflow pointed out a part of code that was not checking for existence prior to usage - I made that fix.

Source/ACE/Entity/Landblock.cs
ogmage78: Thanks for the review.   We were chatting about it in Discord -  was going to address that later as it is such an edge case.   I would have to try to pickup something someone had dropped across a landblock boundary.   At the moment all it should do is fail to pick it up.   I can refactor to use the FindWorldObject later if that is ok?   Thanks

PR #236, opened 2017-04-12
Concept review: Broadcast Generic Messages for Effects and Sounds

fantoms:
Please comment and discuss the code below -

The code length was getting quite large for landblock action queue broadcasting, so I came up with a generic solution that packs any `GameMessage` inside of an `OutboundMessageArgs` object.

This allows for any `GameMessage` to potentially be sent over the `Landblock` `Broadcast` function.

I also implemented changes in landblock to broadcast the effects and sounds, with generic functions.

PLEASE NOTE: One obstacle that this method does not address, is per session sequences. If a Game Message has been extended into another message type (like `GameEventPlayerKilled`), then the initial sequence for the source session will be used. This does not matter for game messages that do not use sequences, like effects or sounds, but a solution will need to created to address this issue. This also means that the PlayerKilled message is currently bugged in this PR.

The functionality that was worked on to produce the proof of concept, are effects and sounds, so when reviewing you may find it useful to start your breakpoints at `ActionApplyVisualEffect`.

All discussion is welcome for this PR - Thanks!


Zegeger:
I have a thought on this, but will require some refactor work.
I think we need to move all GameMessage/GameEvents to store inputted data, and then we add a 
byte[] GetPayload(Session session) 
function which actually writes the content for sending.  Then a single event can be used repeatedly to broadcast to many sessions.

fantoms:
Thanks for taking a look!

PR #239, opened 2017-04-15
Creature kill corpse v2

Lidefeath:
This is the new version of the creature kill and corpse spawning with the smite command. Now the animations work without a sleep command but by introducing delays into the GameActionQueue. More details in the changelog.


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -397,7 +397,44 @@ public static void HandleSmite(Session session, params string[] parameters)
         {
             // @smite [all] - Kills the selected target or all monsters in radar range if "all" is specified.
 
-            // TODO: output
+            if (parameters?.Length > 0)
+            {
+                if (parameters[0] == "all")
```
Mogwai-AC: The more I think about it, the more I don't like having "/smite all" as a thing.  Not needed by players, not needed by admins (really), and creates some unnecessary design complications.

Lidefeath:
I removed the smite command as requested - although I don't see why implementing it was so bad if it's in the client.

PR #240, opened 2017-04-15
Feature update: Death - do not merge

fantoms:
This is an update to the player death mechanics.

- Adding player death broadcasts initial code
- Adding the game message for player killed
- Added player messages and events
- Death messages are now on the action queue
- Added you have/name has to message
- Added queued teleport too landblocks
- Added properties to character and added to player kill
- Changed vitae till I find a the formula
- Update health to resurect
- Changed order of broadcasts
- Added game event for purgeing character enchantments on death
- Implemented a more Proper death, ie deathV2
- Limit death rebroadcast by landblock
- Used correct health update

Please take a good look at the action queue and tell me if you know of a better way to add the events to the queue, specifically line 439 of the landblocks.cs file.

- Also note that there were some CR:LF problems on a few files, from previous commits.

- Another known issue to note: if you use `/die` in a landblock adjacent to a lifestone, an adjacent player may see the death message multiple times.


Source/ACE/Network/GameAction/GameActionType.cs
```diff
@@ -166,6 +166,9 @@ public enum GameActionType
 
         MovementEvent = 0xF74C,
 
+        // ACE ONLY GameActionEvent to send Events over the "Action" queue:
+        GameActionEvent = 0xF819, // Only used when checking the action queue for an event.
```
Mogwai-AC: Does this action code exist in the NetworkProtocol?  I don't see it in the (outdated) Skunkworks docs.  This is a pretty dangerous precedent.

Source/ACE/Network/GameAction/Actions/GameActionDie.cs
```diff
@@ -6,7 +6,7 @@ public static class GameActionDie
         [GameAction(GameActionType.EvtCharacterSuicide)]
         public static void Handle(ClientMessage message, Session session)
         {
-            session.Player.Kill();
+            session.Player.Kill(session.Player.Guid);
```
Mogwai-AC: Should enqueue a GameAction here so the landblock can do the corpse-dropping logic.

Source/ACE/Entity/Landblock.cs
```diff
@@ -320,6 +328,13 @@ private void Broadcast(BroadcastEventArgs args, bool propogate, Quadrant quadran
                         Parallel.ForEach(players, p => p.SendMovementEvent(args.Motion, args.Sender));
                         break;
                     }
+                case BroadcastAction.BroadcastDeath:
+                    {
+                        // players never need an update of themselves
+                        players = players.Where(p => p.Guid != args.Sender.Guid).ToList();
```
Mogwai-AC: // TODO: Filter to players in range and include adjacencies

Source/ACE/Entity/Player.cs
```diff
@@ -240,83 +223,91 @@ public void Kill()
             Location = character.Location;
 
             // TODO: Move this all into Character Creation and store directly in the database.
-            if (DatManager.PortalDat.AllFiles.ContainsKey(0x0E000002))
+            try
```
Mogwai-AC: I don't think you meant to change all this stuff...

Source/ACE/Entity/Player.cs
```diff
@@ -710,6 +701,102 @@ public void SpendXp(Skill skill, uint amount)
             Session.Network.EnqueueSend(xpUpdate, skillUpdate, soundEvent, message);
         }
 
+        public void Kill(ObjectGuid killerId)
```
Mogwai-AC: I think I'm going to be a really big fan of overriding the existing Kill function (renaming it to OnKill).  Then, here in Player.cs, you'd just override, call base.OnKill(...), then do player respawning stuff.  You could also update the CorpseFactory to pull items off the character, etc, that would drop on death.

Source/ACE/Entity/Landblock.cs
```diff
@@ -466,6 +481,28 @@ private void HandleGameAction(QueuedGameAction action, Player player)
         {
             switch (action.ActionType)
             {
+                case GameActionType.TalkDirect:
```
Mogwai-AC: So, about TalkDirect....  as we discussed, this is a hack.  I don't see where you're enqueueing this ActionType at all.  Is it even possible to hit this code block?

Issue #242, opened 2017-04-16
Missing character position crashes server


LtRipley36706:
Deleting a character's position of PositionType=1 causes server to crash when character is loaded.

Background for this:
I used @spacejump command which put me in an endless portal (command should probably be removed or fixed). I figured I'd just delete that character's current position and it would default back to fallback position but it did not work. Probably needs some logic built in the load-in if no current position is in the db to reset to a safe location, either lifestone if set and failing that probably holtburg drop (or race specific starter town drop).

cc: @fantoms

PR #243, opened 2017-04-16
Fix: Position missing update and LandblockId db object fix

fantoms:
This code was created to address issue #242.

-Fixed bugs in `LandblockId` in position and server now resets the  position if missing.
-Added checks to prevent a missing location from crashing the server.
-Removed unnecessary `lastportal` recall position save, this is set on demand, this also prevents invalid data from entering the database.

Please give this a test by deleting a previously created character's position location (position type 1) and then try logging in, you should be at the current starting dungeon.

### DIFF
--- a/Source/ACE.Database/CharacterDatabase.cs
+++ b/Source/ACE.Database/CharacterDatabase.cs
@@ -277,8 +277,28 @@ namespace ACE.Database
 
                 // Loads the positions into the player object and resets the landlock id
                 LoadCharacterPositions(c);
-                c.Location = c.Positions[PositionType.Location];
-                c.Location.LandblockId = new LandblockId(c.Location.Cell);
+
+                // Load the best location from the database
+                // This may be needed if a position is removed from the database, after a character has been created.
+                // But should now be totally un-neccssary.
+                var invalidPosition = CharacterPositionExtensions.InvalidPosition(c.Id, PositionType.Undef);
+
+                if (c.Positions.ContainsKey(PositionType.Location))
+                    c.Location = c.Positions[PositionType.Location];
+                else if (c.Positions.ContainsKey(PositionType.LastPortal))
+                    c.Location = c.Positions[PositionType.LastPortal];
+                else if (c.Positions.ContainsKey(PositionType.Sanctuary))
+                    c.Location = c.Positions[PositionType.Sanctuary];
+
+                // Check to see if the database position is valid
+                if (c.Location.PositionX == invalidPosition.PositionX && 
+                    c.Location.PositionY == invalidPosition.PositionY && 
+                    c.Location.PositionZ == invalidPosition.PositionZ)
+                {
+                    // If the cell is also invalid, we will set to initial character starting location.
+                    if (c.Location.LandblockId.Raw == 0x0 || c.Location.LandblockId.Raw == 0x255)
+                        c.Location = CharacterPositionExtensions.StartingPosition(c.Id);
+                }
 
                 uint characterOptions1Flag = result.Read<uint>(0, "characterOptions1");
                 uint characterOptions2Flag = result.Read<uint>(0, "characterOptions2");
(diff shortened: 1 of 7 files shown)


Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -452,19 +472,24 @@ public void LoadCharacterPositions(Character character)
...
+                }
+            }
+
             // Check for positions and insert defaults if missing:
-            if (dbPositionList.Count == 0)
+            if (dbPositionList.Count == 0 || !containsLocation)
             {
                 // load the default position
                 Position newCharacterPosition = CharacterPositionExtensions.StartingPosition(character.Id);
-                Position newRecallPosition = CharacterPositionExtensions.InvalidPosition(character.Id, PositionType.LastPortal);
```
fantoms: This will be set on demand in the future, when the character uses a portal.

Source/ACE.Entity/Position.cs
```diff
@@ -97,8 +107,7 @@ public Position InFrontOf(double distanceInFront = 3.0f)
 
         public Position(uint characterId, PositionType type, uint newCell, float newPositionX, float newPositionY, float newPositionZ, float newRotationX, float newRotationY, float newRotationZ, float newRotationW)
         {
-            LandblockId = new LandblockId(Cell);
-            
+            LandblockId = new LandblockId(newCell);
```
fantoms: Bug fix: this was saving the previous value as the `LandblockId`.

Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -277,8 +277,28 @@ public bool IsNameAvailable(string name)
 
                 // Loads the positions into the player object and resets the landlock id
                 LoadCharacterPositions(c);
-                c.Location = c.Positions[PositionType.Location];
-                c.Location.LandblockId = new LandblockId(c.Location.Cell);
```
fantoms: This line is no longer necessary to set `LandblockId`, as the position's `LandblockId` should now be set when changing the Cell or `LandblockId` of the Position.

Source/ACE.Database/CharacterDatabase.cs
```diff
@@ -277,8 +277,28 @@ public bool IsNameAvailable(string name)
...
+                var invalidPosition = CharacterPositionExtensions.InvalidPosition(c.Id, PositionType.Undef);
+
+                if (c.Positions.ContainsKey(PositionType.Location))
+                    c.Location = c.Positions[PositionType.Location];
+                else if (c.Positions.ContainsKey(PositionType.LastPortal))
+                    c.Location = c.Positions[PositionType.LastPortal];
+                else if (c.Positions.ContainsKey(PositionType.Sanctuary))
+                    c.Location = c.Positions[PositionType.Sanctuary];
+
+                // Check to see if the database position is valid
```
fantoms: I fully see that this is unnecessary, but prevents an imaginary scenario where the position has become 0 for xyz coordinates and the landblock is invalid. This may happen if someone uses the reset-pos command to set location as an invalid position.

Source/ACE.Entity/Position.cs
```diff
@@ -10,7 +10,17 @@
     [DbGetList("vw_character_positions", 23, "character_id")]
     public class Position
     {
-        public LandblockId LandblockId { get; set; }
+        public LandblockId LandblockId
+        {
+            get
+            {
+                return new LandblockId(Cell);
```
Mogwai-AC: This doesn't feel helpful.  Essentially re-allocating a new LandblockId every time it is accessed.  I agree the section below in the constructor was a bug, but I don't see the bug in this part.

Source/ACE.Database/CharacterDatabase.cs
Mogwai-AC: Do X/Y/Z really matter?  Just seems like unnecessary overhead.  Will also likely cause problems in dungeons that go outside the bounds of "normal" X/Y/Z values of the open world.

fantoms: The only real case this check prevents is in where someone has used `@reset-pos 1` and their location has become invalid. 

I will remove this check and the database debugging command as well, too try and prevent attempt at preventing invalid locations.

Source/ACE.Entity/Position.cs
fantoms: I do see how creating a new object every time will be inefficient, so I will work to create private accessed variable, instead.

fantoms:
I've made changes to each of the commented sections - the code is now ready for another review.

PR #248, opened 2017-04-18
Creature respawn, corpse despawn, random spawns

Lidefeath:
Corpses will despawn after a set amount of time for testing. The retail calculation for the despawn time is currently commented out. Creatures will respawn after a set amount of time too.
Introduced loading randomly generated spawns from the database. See changelog on which table to edit to alter the location.

Please don't mind the Remove Whitespaces commit that was an attempt to get my git to do what I wanted :-)

### DIFF
--- a/Source/ACE/Entity/Container.cs
+++ b/Source/ACE/Entity/Container.cs
@@ -14,6 +14,11 @@ namespace ACE.Entity
 
         private readonly object inventoryMutex = new object();
 
+        // <summary>
+        /// Time when this object will despawn, -1 is never.
+        /// </summary>
+        public double DespawnTime { get; set; } = -1;
+
         public Container(ObjectType type, ObjectGuid guid, string name, ushort weenieClassId, ObjectDescriptionFlag descriptionFlag, WeenieHeaderFlag weenieFlag, Position position)
             : base(type, guid)
         {
--- a/Source/ACE/Entity/Container.cs
+++ b/Source/ACE/Entity/Container.cs
@@ -14,11 +14,6 @@ namespace ACE.Entity
 
         private readonly object inventoryMutex = new object();
 
-        // <summary>
-        /// Time when this object will despawn, -1 is never.
-        /// </summary>
-        public double DespawnTime { get; set; } = -1;
-
         public Container(ObjectType type, ObjectGuid guid, string name, ushort weenieClassId, ObjectDescriptionFlag descriptionFlag, WeenieHeaderFlag weenieFlag, Position position)
             : base(type, guid)
         {
--- a/Source/ACE/Entity/Creature.cs
+++ b/Source/ACE/Entity/Creature.cs
@@ -46,6 +46,11 @@ namespace ACE.Entity
         /// </summary>
         public bool IsAlive { get; set; }
 
+        /// <summary>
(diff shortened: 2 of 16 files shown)


Source/ACE/Entity/Creature.cs
```diff
@@ -164,6 +171,10 @@ public void Kill(Session session)
             corpse.Location.PositionY -= corpse.PhysicsData.ObjScale;
             corpse.Location.PositionZ -= corpse.PhysicsData.ObjScale / 2;
 
+            // Corpses stay on the ground for 5 * player level but minimum 1 hour
+            // corpse.DespawnTime = Math.Max((int)session.Player.PropertiesInt[Enum.Properties.PropertyInt.Level] * 5, 360) + WorldManager.PortalYearTicks; // as in live
+            corpse.DespawnTime = 20 + WorldManager.PortalYearTicks; // only for testing
+
             // Remove Creature from Landblock and add Corpse in that location via the ActionQueue to honor the motion delays
             QueuedGameAction removeCreature = new QueuedGameAction(this.Guid.Full, this, true, true, GameActionType.ObjectDelete);
             QueuedGameAction addCorpse = new QueuedGameAction(this.Guid.Full, corpse, true, GameActionType.ObjectCreate);
```
Mogwai-AC: Doesn't the corpse creation need to get broadcast to all players?  And added to the Landblock?

Source/ACE/Entity/Landblock.cs
```diff
@@ -359,14 +372,20 @@ public void UseTime()
...
                 List<WorldObject> movedObjects = null;
                 List<Player> players = null;
+                List<Container> despawnObjects = null;
+                List<Creature> deadCreatures = null;
 
                 lock (objectCacheLocker)
                 {
                     movedObjects = this.worldObjects.Values.OfType<WorldObject>().ToList();
                     players = this.worldObjects.Values.OfType<Player>().ToList();
+                    despawnObjects = this.worldObjects.Values.OfType<Container>().ToList();
```
Mogwai-AC: I would think that we would want to be able to despawn anything.  Case in point, junk loot you drop on the ground needs to go away eventually.  Perhaps use WorldObject instead of Container?

Source/ACE/Entity/Container.cs
```diff
@@ -14,6 +14,11 @@ public class Container : WorldObject
 
         private readonly object inventoryMutex = new object();
 
+        // <summary>
+        /// Time when this object will despawn, -1 is never.
+        /// </summary>
+        public double DespawnTime { get; set; } = -1;
```
Mogwai-AC: I'm inclined to think this should be moved to WorldObject.  This certainly applies to corpses, but it also applies to random items on the ground.

Source/ACE/Entity/Landblock.cs
```diff
@@ -417,6 +436,26 @@ public void UseTime()
                         HandleGameAction(action, p);
                 }
 
+                // despawn objects
+                Parallel.ForEach(despawnObjects, deo =>
```
Mogwai-AC: I don't think Parellel is worth the the thread overhead here.  You're not doing anything substantial in the loop.  Same with respawning creatures below.

PR #250, opened 2017-04-18
Add Misc Object

LtRipley36706:


Mogwai-AC:
What's the use case for this?

LtRipley36706:
Use case is for house objects but per our discussion I should look further at this to refine to a more specific object

LtRipley36706:
reworking at another time

PR #251, opened 2017-04-18
Added Folders to Project Structure to make maintaining and developing easier

Forbiddenz:


Lidefeath:
I like this structure. 👍

Mogwai-AC:
Folders should match namespaces, it's even a StyleCop rule. This type of object organization is not normal practice in industry and I'm not a fan.

Forbiddenz:
Yea, I was wondering about that, I do like the folders though as they make things more clear.
What I we added the folders and added the name modified the namespaces accordingly?

Mogwai-AC:
If we had a hundred or more objects in our hierarchy, I could get on board with something, but what you're asking is actually technically not possible.  You can't have "ACE.Entity.WorldObject" and "ACE.Entity.WorldObject.Container" - it won't even compile because it you can't have a class share a namespace segment.

PR #255, opened 2017-04-19
Phase 1 Refactor of Session, NetworkSession, Action handlers

Zegeger:
Merged SocketManager into WorldManager
Made Session inherit from NetworkSession, flattening the model
Moved GameMessage handlers into Session class, partial classes
Moved GameAction handlers into Player class, partial classes
This move makes functionality encapsulated into these classes, allowing a move to fewer public members.
Phase 2 will be to build on WorldObject model to improve queued game actions, broadcast, and action ordering/timing.


LtRipley36706:
So far the only issue i've encountered that's new is I can sometimes get to the character select screen and get no characters in the list upon server start up. waiting a bit tends to mitigate that out. Perhaps connections should be blocked until everything is ready to go?

Mag-nus:
It looks like the namespace for the player handler classes in ACE.Network.Handlers is not correct. Many are ACE.Entity when they should be ACE.Network.Handlers

The session handlers appear to have the wrong namespaces as well.

PR #258, opened 2017-04-19
Streaming objects

Forbiddenz:
Streaming Objects now seem to be fully working.
Needs testers!  could cause some code elsewhere to have some issues..

Maybe teleport needs to clear this, not to sure on that yet.
But working pretty good.


changelog.md
```diff
@@ -1,5 +1,21 @@
...
+* Add an ActionQueue switch case to implement Portal travel
+
+### 2017-04-18
+[Jyrus]
+* Create a new table adding Portal destinations that reference the WeenieClassID of the portal object
+* Add an ActionQueue switch case to implement Portal travel
+
+# ACEmulator Change Log
+
 ### 2017-04-19
```
Lidefeath: The changelog is a bit out of order.

Source/ACE/Entity/Landblock.cs
```diff
@@ -369,22 +408,27 @@ public void UseTime()
...
                 List<Creature> deadCreatures = null;
 
                 lock (objectCacheLocker)
                 {
-                    movedObjects = this.worldObjects.Values.OfType<WorldObject>().ToList();
-                    players = this.worldObjects.Values.OfType<Player>().ToList();
-                    despawnObjects = this.worldObjects.Values.ToList();
-                    deadCreatures = this.worldObjects.Values.OfType<Creature>().ToList();
+                    allworldobj = this.worldObjects.Values.OfType<WorldObject>().ToList();
                 }
```
Lidefeath: Isn't it faster to just use allworldobj = this.worldObjects.Values.ToList(); ?

Forbiddenz: Yea, I missed that one.
I use to need to cast them to their types but then I changed it to lock in all world objects instead of each one, since I now need to know all of them and not just players, etc

Will fix tomorrow

changelog.md
Forbiddenz: yep, did not realize that happened when I merged it back on to master.
I will update that too

changelog.md
```diff
@@ -1,4 +1,8 @@
 # ACEmulator Change Log
+### 2017-04-20
```
LtRipley36706: Don't duplicate dates here, move your section to above the most recent

Forbiddenz: Ok, Fixed

Source/ACE/Entity/Player.cs
```diff
@@ -1,1226 +1,1231 @@
...
+                Teleport(Positions[PositionType.Sanctuary]);
+            else
+                Teleport(Positions[PositionType.Location]);
+
+            // create and send the death event
+            var yourDeathEvent = new GameEventYourDeath(Session);
+            Session.Network.EnqueueSend(yourDeathEvent);
+        }
+
+        public async Task Load(Character preloadedCharacter
```
LtRipley36706: I killed this line in PR #268. it's doing more harm than good. A great example of that is to `@telepoi holtburg` and run to the portal down the hill, walk backwards from it and watch objects blink as they get repopulated for no good reason. it didn't work like that retail, it shouldn't here either.

I know you didn't add this.. just noting it for you

Source/ACE/Entity/Landblock.cs
```diff
@@ -36,6 +36,7 @@ internal class Landblock
         private const float outDoorChatRange = 75f;
         private const float indoorChatRange = 25f;
         private const float maxXY = 192f;
+        private const float maxobjectRange = 15625;
```
LtRipley36706: Change name to maxObjectRange and value to the 20k per discussion

LtRipley36706:
More successful testing tonight.. I dumped in a significant number of objects for this to stream it it mostly held up.. One time it failed to stream in objects when I was on an autorun but it also could have been due to the sheer number of objects as well.. Something to try to reproduce if possible.

Source/ACE/Entity/Landblock.cs
```diff
@@ -165,6 +167,37 @@ public void StartUseTime()
             new Thread(UseTime).Start();
         }
 
+        private void AddPlayerTracking(List<WorldObject> wolist, Player player)
+        {
+            Parallel.ForEach(wolist, (o) =>
+            {
+                if (o.Guid.IsCreature())
```
Mogwai-AC: could be simplified:
if (!o.Guid.IsCreature() || (o.Guid.IsCreature && (o as Creature).IsAlive)
    player.TrackObject(o);

Source/ACE/Entity/Landblock.cs
```diff
@@ -290,41 +365,49 @@ private void Broadcast(BroadcastEventArgs args, bool propogate, Quadrant quadran
...
                     {
-                        Parallel.ForEach(players, p => p.StopTrackingObject(wo));
+                        players = players.Where(p => p.Location?.IsInQuadrant(quadrant) ?? false).ToList();
+                        Parallel.ForEach(players, p => p.StopTrackingObject(wo, true));
                         break;
                     }
                 case BroadcastAction.AddOrUpdate:
                     {
+                        // supresss updating if player is out of range of world object.= being updated or created
+                        players = GetWorldObjectsInRange(wo, maxobjectRange, true).OfType<Player>().ToList();
```
Mogwai-AC: you could significantly improve the performance of this by only looking for players instead of looking for everything then filtering down to players, especially since this is being done in several places.

Forbiddenz: I am not so sure because keep in mid objects need to be marked as they come into range.   Player is just for players detected in range but objects also need to be detected

Forbiddenz: Oh maybe, I just realized where that line is.  That may be a good idea the more I think about it

PR #267, opened 2017-04-21
Phase 1 Refactor of Session, NetworkSession, Action handlers

Zegeger:
Merged SocketManager into WorldManager
Made Session inherit from NetworkSession, flattening the model
Moved GameMessage handlers into Session class, partial classes
Moved GameAction handlers into Player class, partial classes
This move makes functionality encapsulated into these classes, allowing a move to fewer public members.
Phase 2 will be to build on WorldObject model to improve queued game actions, broadcast, and action ordering/timing.


Source/ACE/Entity/Landblock.cs
```diff
@@ -87,8 +85,8 @@ public Landblock(LandblockId id)
             var creatures = DatabaseManager.World.GetCreaturesByLandblock(this.id.Landblock);
             foreach (var c in creatures)
             {
-                Creature cwo = new Creature(c);
-                worldObjects.Add(cwo.Guid, cwo);
+                // Creature cwo = new Creature(c);
+                // worldObjects.Add(cwo.Guid, cwo);
```
Lidefeath: You disabled creature spawning here. Is there another place where this is happening now, as I couldn't really find it anywhere else.

Zegeger: Reverted it.  For some reason that section causes an Exception for me, so was easier to comment it out.

Source/ACE/Entity/Objects/PlayerAllegiance.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Threading.Tasks;
+
+namespace ACE.Entity.Objects
+{
+    public partial class Player
+    {
+        [GameAction(GameActionType.AllegianceUpdateRequest)]
+        private void AllegianceUpdateRequestAction(ClientMessage message)
+        {
+            var unknown1 = message.Payload.ReadUInt32();
```
Mogwai-AC: This is bad for the Player class.  The Player class, which represents a person at the keyboard playing the game, is now responsible for parsing network Game Actions?  Please no.

Source/ACE/Entity/Objects/PlayerCombat.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ACE.Entity.Objects
+{
+    public partial class Player
+    {
+        [GameAction(GameActionType.Suicide)]
+        private void CharacterSuicide(ClientMessage message)
```
Mogwai-AC: Well this used to work...

PR #268, opened 2017-04-21
Kill UpdateObject sending until we use it right.

LtRipley36706:
…y other things which leads to issues.


LtRipley36706:
This will make multiplayer look rather static but honestly the twitching isn't better. the UpdateObject could be replaced with CreateObject messages and it actually looks smoother, albeit still wrong. Clients would appear to skate/slide across the screen. Spams Pcaps however, which is still wrong.

PR #270, opened 2017-04-21
Fix IsDerivedPlayer logic to allow smite to work

fantoms:
This was always true, due to a changes I made last night to the Creature Kill command. The changes made the Creatures also work like a player, but after I made this change, the types can now differentiate; allowing `@smite` and `@smite all` began to work correctly again.

### DIFF
--- a/Source/ACE/Entity/Creature.cs
+++ b/Source/ACE/Entity/Creature.cs
@@ -155,10 +155,7 @@ namespace ACE.Entity
         {
             IsAlive = false;
             // This will determine if the derived type is a player
-            var isDerivedPlayer = typeof(Player).GetMember("OnKill",
-                   BindingFlags.NonPublic
-                 | BindingFlags.Instance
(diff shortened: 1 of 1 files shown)


Source/ACE/Entity/Creature.cs
```diff
@@ -155,10 +155,7 @@ public virtual void OnKill(Session session)
         {
             IsAlive = false;
             // This will determine if the derived type is a player
-            var isDerivedPlayer = typeof(Player).GetMember("OnKill",
-                   BindingFlags.NonPublic
-                 | BindingFlags.Instance
-                 | BindingFlags.DeclaredOnly).Length == 0;
+            var isDerivedPlayer = GetType().Equals(session.Player.GetType());
```
Mogwai-AC: Reflection is really slow. Why not just use ObjectGuid.IsPlayer?

fantoms: Yes, that seems to work well; thank you for this!

PR #271, opened 2017-04-21
State machines

Forbiddenz:
State Machines - I built this off steaming, so it should go on top of that.
-- I got no way to tested it good yet, but it looks good from code point of view.

We need some things that use state machines, like doors!


Forbiddenz:
Yep, it conflicts with my other or for sure plus more now.

I will pull it and redo it in a few days.
I may have some better ideas to for this state machine.  Right now is kind of a rules engine of states.  We may want forward states too so we can move to next state in a ordered list for toggle switches.

Forbiddenz:
Closing it, it's in another pr with one code

PR #273, opened 2017-04-22
Read the Generators from portal.dat file 0x0E00000D and enhance DatRe…

Lidefeath:
…ader to handle obfuscated strings.


Source/ACE.DatLoader/DatReader.cs
```diff
@@ -93,16 +93,37 @@ public float ReadSingle()
...
             byte[] thestring = new byte[stringlength];
             Array.Copy(Buffer, Offset, thestring, 0, stringlength);
             Offset += stringlength;
             return System.Text.Encoding.ASCII.GetString(thestring);
         }
 
+        /// <summary>
+        /// Returns a string as defined by the first byte's length and removes the obfuscation
+        /// </summary>
+        public string ReadOString()
```
Mogwai-AC: What's an "OString"?  heck, what's a "PString" for that matter.  I realize you're deobfuscating strings, but let's deobfuscate method names while we're at it.

Lidefeath:
Ok, changes are in.

Source/ACE.DatLoader/DatReader.cs
```diff
@@ -93,16 +93,37 @@ public float ReadSingle()
...
             byte[] thestring = new byte[stringlength];
             Array.Copy(Buffer, Offset, thestring, 0, stringlength);
             Offset += stringlength;
             return System.Text.Encoding.ASCII.GetString(thestring);
         }
 
+        /// <summary>
+        /// Returns a string as defined by the first byte's length and removes the obfuscation
+        /// </summary>
+        public string ReadObfuscatedString()
```
Zegeger: I think a better name would be ReadBigEndianString

Source/ACE.DatLoader/DatReader.cs
```diff
@@ -93,16 +93,37 @@ public float ReadSingle()
             return data;
         }
 
+        /// <summary>
+        /// Returns a string as defined by the first byte's length
+        /// </summary>
         public string ReadPString()
```
Zegeger: PString just seems to be their version of a basic string.  This seems to be identical to ReadString except it reads the length from the first byte.  I don't see any references to ReadString being used, so should this just replace ReadString?

PR #275, opened 2017-04-23
Added 65 more portals

MiachofTD:
Added portals for areas -
Yaraq - Starting Town
Zaikhal - Capital
Al-Arqas
Al-Jalima
Ayan Baqur
Khayyaban


MiachofTD:
Changing how we are adding portals.

PR #279, opened 2017-04-23
More portal work

kdmxp256:
Hope it isn't too much if-then'ing.  I did use some switch statements, where it looked appropriate.  Much of this UseItem for the portals, if not all, should eventually be moved into Portal.cs, within the OnCollide method.  However, it may take passing more than just the player object into the OnCollide method to make it work, as I have constructed the statement block for making portals work.


Database/Updates/World/007_2017_04_21_portal_destination.sql
```diff
@@ -91,3 +91,12 @@ INSERT INTO `portal_destination` (`weenieClassId`, `cell`, `x`, `y`, `z`, `qx`,
...
 (43067,459094,80,-70,0.005,0,0,0.70710677,0.70710677,0,0);
+
+INSERT INTO `portal_destination` (`weenieClassId`, `cell`, `x`, `y`, `z`, `qx`, `qy`, `qz`, `qw`, `min_lvl`, `max_lvl`) VALUES
+(29337,3663003677,84.8,99,20,0,0,0,1,0,0),
+(29338,2847146009,84,7.1,94.01,0,0,-0.08,1,0,0),
+(29339,869859336,7.5,184.35,52.01,0,0,0,-1,0,0),
+(29340,2103705613,31.9,104.6,11.95,0,0,-0.82,0.58,0,0),
+(43002,4135714867,145.7,49.85,58.01,0,0,-0.88,-0.47,0,0),
+(31061,2248344094,50,-54,0.0065,0,0,-1,0,0,0),
+(29334,2248344259,119,-141,0.0065,0,0,1,0,0,0);
```
LtRipley36706: I suspect it would be very bad practice to retroactively update previous scripts. Whatever changes need to be made would be in new scripts and the assumption made they are executed in order.

PR #280, opened 2017-04-24
Update: Vitals

fantoms:
* Working on Vitals, Changed `Current` to `UnbuffedValue` for the `ICreatureStats` interface within `Character.cs`

* Added a health update message for abilities, so endurance can properly change the health/steam, when increased.

* `@heal` should now work again.


Zegeger:
Looks good to me.

PR #284, opened 2017-04-24
Add portals, portal routing, and updates to 2 previous portals

kdmxp256:
As noted in the changelog, PHPMYADMIN does not like the ON DUPLICATE KEY UPDATE statements in the sql file; however, I personnally had success using SQLyog.  From my reading MySQL ( MariaDB ) should be able to accept the commands, as they are valid MySQL statements, from its command line.  I would guess that HeidiSQL supports them, also.

### DIFF
--- a/Source/ACE/Entity/Landblock.cs
+++ b/Source/ACE/Entity/Landblock.cs
@@ -790,10 +790,119 @@ namespace ACE.Entity
 
                                             if (portalDestination != null)
                                             {
-                                                player.Session.Player.Teleport(portalDestination.Position);
-                                                // always send useDone event
-                                                var sendUseDoneEvent = new GameEventUseDone(player.Session);
-                                                player.Session.Network.EnqueueSend(sendUseDoneEvent);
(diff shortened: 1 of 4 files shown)


Source/ACE/Entity/Landblock.cs
```diff
@@ -790,10 +790,119 @@ private void HandleGameAction(QueuedGameAction action, Player player)
...
-                                                player.Session.Network.EnqueueSend(sendUseDoneEvent);
+                                                if ((player.Level >= portalDestination.MinLvl) && ((player.Level <= portalDestination.MaxLvl) || (portalDestination.MaxLvl == 0)))
+                                                {
+                                                    Position portalDest = portalDestination.Position;
+                                                    switch (obj.WeenieClassid)
+                                                    {
+                                                        /// <summary>
+                                                        /// Setup correct racial portal destination for the Central Courtyard in the Training Academy
+                                                        /// </summary>
+                                                        case 31061:
```
Mogwai-AC: Create constants for hard-coded values here.

PR #285, opened 2017-04-24
Add 16 more portals from around Shoushi

Jyard1:


Database/Updates/World/010_2017_04_24_destination_portal.sql
```diff
@@ -0,0 +1,15 @@
+INSERT INTO `portal_destination` (`weenieClassId`, `cell`, `x`, `y`, `z`, `qx`, `qy`, `qz`, `qw`, `min_lvl`, `max_lvl`) VALUES
+(1124, 32965104, 40, -140, 0.005, 1, 0, 0, 0, 0, 0),
+(1096, 3662938154, 135.1000061, 42.70000076, 34, -0.96592581, 0, 0, -0.25881913, 0, 0),
```
Mogwai-AC: I think you have your qx and qz values backwards, here.  QX rotation is like doing barrel-roll rotations.  No bueno.

Mogwai-AC:
Also, you need to put a comment in the changelog to reflect what you did.

Database/Updates/World/010_2017_04_24_destination_portal.sql
Jyard1: Ah didn't see that /teleloc used a different order, will fix tanks.

Jyard1:
Not sure what AppVeyor's problem is with the update to the changelog.

PR #291, opened 2017-04-25
Added XpTable class to ACE.DatLoader and removed the JSON Xp files. 

OptimShi:
Some additional cleanup in the other DatLoader.FileTypes classes to make them easier to use.


Source/ACE/Entity/Player.cs
```diff
@@ -536,7 +536,8 @@ public QueuedGameAction ExaminationQueuePop()
         public void GrantXp(ulong amount)
         {
             // until we are max level we must make sure that we send
-            var chart = DatabaseManager.Charts.GetLevelingXpChart();
+            XpTable xpTable = XpTable.ReadFromDat();
```
Mogwai-AC: These should be read once and cached in a static variable.  I'd rather have higher memory requirements than create performance issues of reading from dat files every time somebody clicks the up arrow.

OptimShi: The XpTable class is already a "singleton" and caches the dat loading. (I learned that from you what I did the CharGen stuff) You had me questioning if this was working, but I threw a breakpoint in there to confirm and it is working as intended.

PR #294, opened 2017-04-25
Renamed the last 3 SQL files to be in correct order

MiachofTD:
Renamed the last 3 SQL files to be in correct order and made changes to changelog and MySqlInstall.bat file to reflect those changes.


LtRipley36706:
See #295, this PR will end up being undone because of it.

PR #297, opened 2017-04-26
Feature: Admin `@boot` command

fantoms:
This request should allow a player or account holder, too be kicked from the server. There is an assumption that when the player is sent boot packet, the server will then receive a disconnect message from the client, where finally the server can process the session's finally "log off".

### DIFF
--- a/Source/ACE/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE/Command/Handlers/AdminCommands.cs
@@ -8,6 +8,7 @@ using ACE.Network.GameMessages.Messages;
 using ACE.Common;
 using ACE.Database;
 using ACE.Network.Enum;
+using System.Collections.Generic;
 
 namespace ACE.Command.Handlers
 {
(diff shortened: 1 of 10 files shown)


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -87,15 +89,117 @@ public static void HandleBanlist(Session session, params string[] parameters)
...
+                    // Boot the player
+                    playerSession.BootPlayer();
+
+                    // Send an update to the admin, but prevent sending too if the admin was the player being booted
+                    if (session != null)
+                    {
+                        // Log the player who initiated the boot
+                        bootText = "Player: " + session.Player.Name + " has booted " + bootText;
+                        if (session != playerSession)
+                            session.Network.EnqueueSend(new GameMessageSystemChat(bootText, ChatMessageType.Broadcast));
```
LtRipley36706: there's one more message to send here to the audit channel, give me a bit to look up the phrasing..

basically, any admin in-game is auto subscribed to it and it sent out copies of all important commands

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -87,15 +89,117 @@ public static void HandleBanlist(Session session, params string[] parameters)
...
+            // Did not find a player
+            string errorText = $"Error locating the player or account to boot.";
+            // Send the error to a player or the console
+            if (session != null)
+            {
+                session.Network.EnqueueSend(new GameMessageSystemChat(errorText, ChatMessageType.Broadcast));
+            } else
+            {
+                Console.WriteLine(errorText);
+            }
```
LtRipley36706: single line if-else need no brackets, that will also fix spacing issue with else

Source/ACE/Command/Handlers/AdminCommands.cs
fantoms: I agree and I would like to move change to a "//TODO:" for a later time. We should wait until after the server group chat foundation has been built out, too support this type of messaging.

PR #300, opened 2017-04-27
Movement 2 - Ready for review and Merge

ogmage78:
* Added Turn to Object and Turn To Heading messages.
* Cleaned up earlier work
* You can now test move to object and put item in container.:
@ctw ground. 
Select wand
Move away from the area with the item still selected. Over 15f to see run under to see walk.
Press the use key - you will walk or run to the wand depending on your distance, then stop and pick it up.
Also implements the first use of StackOverflows state machine
Contains view for character - to start implementing orm for character
NOTE: There is a known bug if you create the training wand in your backpack and drop it, the move to does not work correctly yet. I think it is related to the landblock not doing updates atm.


ogmage78:
Found the bug - it is working now.

PR #303, opened 2017-04-29
chore: update various git settings

delasteve:


.gitignore
```diff
@@ -251,5 +268,22 @@ paket-files/
 .idea/
 *.sln.iml
 
-# Ignore customized Config.json
-Source/ACE/Config.json
```
LtRipley36706: we lost this in the in the change, we would still want to ignore this file, yeah?

delasteve: Yes. From the review comment, it sounded like it was a base `.gitignore` from GitHub, and I didn't bother to check. Sorry.

LtRipley36706: no problem, all fixed

PR #304, opened 2017-04-29
chore: add editorconfig

delasteve:


Mag-nus:
I'm not sure if this is even possible, or relevant to this config. What I see is missing is the line break type. We use Windows CR LF.

Can you add this?

delasteve:
Fixed

PR #305, opened 2017-04-29
Hooked up the IdentifyObject event and returned a debug string.

OptimShi:
Also added the flags for when we get to the point we are passing actual data.


changelog.md
```diff
@@ -2,6 +2,12 @@
 
 ### 2017-04-26
 [OptimShi]
+* Hooked up the "ExaminationQueuePop" in Player.cs to fire.
+* Added IdentifyResponseFlags, for when we start to send actual data.
+* Added a basic debug string to any ID. It simply returns the objects GUID.
+
+### 2017-04-26
```
LtRipley36706: combine

LtRipley36706: Wait, sorry, the date needs to be changed not combined. My bad.

Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
```diff
@@ -1,15 +1,24 @@
...
+            Writer.Write((uint)flags); // Flags
+            Writer.Write(1u); // Success bool
+
+            // Write a simple debug strings. Thanks to Pea for this idea.
+            Writer.Write(1u + (1 << 16)); // One String, + table size
+            Writer.Write(16u); // Long Description String
+            Writer.WriteString16L("This item has a guid of " + objectID.ToString() + " (0x" + objectID.ToString("X") + ")");
         }
     }
 }
```
LtRipley36706: end on new line please

Source/ACE/Network/Enum/IdentifyResponseFlags.cs
```diff
@@ -0,0 +1,25 @@
...
+        ArmorProfile = 0x80,
+        CreatureProfile = 0x100,
+        WeaponProfile = 0x20,
+        HookProfile = 0x40,
+        ArmorEnchantmentBitfield = 0x200,
+        WeaponEnchantmentBitfield = 0x800,
+        ResistEnchantmentBitfield = 0x400,
+        ArmorLevels = 0x4000,
+    }
+}
```
LtRipley36706: new line here as well

Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
OptimShi: Is there a way to add this to the StyleCop settings? It didn't throw up any alert on this. I have made a mental note, however, for the future.

OptimShi:
Per discussion, I modified this to only spit out debug info on "DebugObjects" (as opposed to proper WorldObjects, which as of now are players, life stones and portals.) Also added the WeenieClassId for further debugging fun (which required modifying how this works a little bit, but all for the best)

Source/ACE/Entity/Landblock.cs
```diff
@@ -477,6 +477,13 @@ public void UseTime()
                     QueuedGameAction action = player.ActionQueuePop();
                     if (action != null)
                         HandleGameAction(action, player);
+                    else
```
Mogwai-AC: Is there a reason we'd only process the Examination Queue when there wasn't a pending action?  I don't get the point of putting this in an else block.

Source/ACE/Entity/Landblock.cs
```diff
@@ -596,6 +603,18 @@ private void HandleGameAction(QueuedGameAction action, Player player)
                         HandleSoundEvent(obj, soundEffect);
                         break;
                     }
+                case GameActionType.IdentifyObject:
+                    {
+                        var g = new ObjectGuid(action.ObjectId);
```
Mogwai-AC: we know the retail servers throttled this, we might want to put a "TODO" comment here to that effect.

Source/ACE/Entity/Landblock.cs
OptimShi: It was a way of giving lower priority to the examine request and also ensure only one action is performed per "tick" so as not to try and do too much at once. Can just as easily modify it to try and perform both an GameAction and an ExamineAction, if any, though.

PR #306, opened 2017-04-29
Portal class changes

kdmxp256:
* I did not combine base_ace_object, ace_object, ace_portal_object into one view at this juncture, as doing so would remove all portals that didn't have destinations set from the world, as the view would only show entries for portal objects that have data from all three tables present.

* In the portal object table restructuring, I added a foreign key constraint to the weenieClassId field that points to that field in the ace_object table.  This addition has the effect of not allowing the ace_portal_object table to be populated until the rest of the ace_world tables are filled.  I also included a portal object that was missing from the ACE-World DB release that was preventing the ace_portal_object table data from being imported.

* An apparent bug with the login position was revealed, when I added saving the PositionType.LastPortal position, when a portal was used.  When a player logs in, the PositionType.LastPortal position gets used, rather than the one saved during the character save action.


kdmxp256:
The societyId field in the new table structure is currently a placeholder that can be changed, if needed, for when Societies and their bound portals are implemented.

Source/ACE.Entity/AcePortalObject.cs
```diff
@@ -0,0 +1,63 @@
...
+        public uint MinLvl { get; set; }
+
+        [DbField("max_lvl", (int)MySqlDbType.UInt32)]
+        public uint MaxLvl { get; set; }
+
+        [DbField("societyId", (int)MySqlDbType.UByte)]
+        public byte SocietyId { get; set; }
+
+        [DbField("isTieable", (int)MySqlDbType.Bit)]
+        public bool IsTieable { get; set; }
```
fantoms: Initially, I am getting a server crash for 'System.ArgumentException': Object of type 'System.Byte' cannot be converted to type 'System.Boolean', when I use a portal. This happens in Database.cs line the portal out of the database.

changelog.md
```diff
@@ -1,4 +1,12 @@
 # ACEmulator Change Log
+### 2017-04-29
+[Jyrus]
+* Changed the structure of the table that contains the Portal Destinations to include other portal properties
+* Renamed table from portal_destination to ace_portal_object and added a foreign key constraint to the weenieClassId field
+* !!! Please note that the ACE-World tables must be filled before importing the ace_portal_object's table / the Desert Mine sql
+	file contains an entry that is currently missing from the ACE-World DB release that also needs to be imported before the
+	ace_portal_object table is loaded.
+* Implemented a portal recall debug command - An apparent bug was revealed that causes players on login to start at the last saved portal recall position
```
fantoms: Bugs should also be reported in the Github issues section or noted in the bugged code section.

Mogwai-AC:
Couple comments:
* If you make the join a "LEFT OUTER JOIN", you won't break existing portals.  In order to support this, the destination data properties will need to be nullable (to make an int nullable, change it to "int?").
* World data needs to go to the ACE-World repository.  Table/View structures here, data there.

PR #309, opened 2017-04-30
Landblocks

OptimShi:
Added client_cell.dat reading for Landblocks and proper z value for @tele commands.

I attempted to add the client_cell.dat reading into the Position class itself, but this caused a cyclical reference. Not sure what the best way to address that is (or if there is even a need?)

Next step is to either complete the other sections of the client_cell.dat reading or to cache these. I don't like that it doesn't right now, since monsters & players & projectiles are going to generate an awful lot of file reads!


Forbiddenz:
I would keep it separated but you may need to see why the the ref issue is happening and if it all that ref is still needed.

Logically I say cell data would not have any ref at all and would be ref by ace, and other classes.

The circular ref occurres when the ace ref points to cell and the cell ref points to ace.  That is poor ok code design so you got to make sure through proper project layout and code design that you only use the ref project as a ref project and not try to interconnect the too.

Hope that helps a little

Source/ACE.DatLoader/FileTypes/CellLandblock.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+/// <summary>
+/// Very special thanks to David Simpson for his early work on reading the cell.dat. Even bigger thanks for his documentation of it!
+/// </summary>
+namespace ACE.DatLoader.FileTypes
+{
+    public class CellLandblock
```
Mogwai-AC: "Landcell" would have been much preferred.

Source/ACE.DatLoader/FileTypes/CellLandblock.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// A landblock is divided into 8 x 8 tiles, which means 9 x 9 vertices reporesenting those tiles. 
+        /// (Draw a grid of 9x9 dots; connect those dots to form squares; you'll have 8x8 squares)
+        /// It is also divided in 192x192 units (this is the x and the y)
+        /// 
+        /// 0,0 is the bottom left corner of the landblock. 
+        /// 
+        /// Height 0-9 is Western most edge. 10-18 is S-to-N strip just to the East. And so on.
+        /// </summary>
+
+        // Places in the inland sea, for example, are false. Should denote presence of xxxxFFFE (where xxxx is the cell).
```
Mogwai-AC: "xxxx" is the Landblock

PR #318, opened 2017-05-02
Land block diag phase1

Forbiddenz:
Phase 1 of Landblock Dia Diag gui.
I need the tool to help improve the landblock loading and unloading code.

This is initial draft of a gui tool built right into ace that can be used to track the landblock activity in real time.  Future use will be to export diags out.

Usage type diag in the console window


Source/ACE.sln
```diff
@@ -45,44 +45,86 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ACE.Tests", "ACE.Tests\ACE.Tests.csproj", "{D8EE6B89-EA54-4F84-988E-BF323C3662C5}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ACE.Diagnostics", "ACE.Diagnostics\ACE.Diagnostics.csproj", "{A4625C11-22BE-4D82-9A6C-63DCB5CAFFB6}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
Mogwai-AC: Please remove the "Any CPU" configurations.

Source/ACE/Entity/Landblock.cs
```diff
@@ -483,7 +493,9 @@ public void UseTime()
                     QueuedGameAction examination = player.ExaminationQueuePop();
                     if (examination != null)
                         HandleGameAction(examination, player);
+                    curplayer++;
                 });
+                UpdateStatus(curplayer);
```
Mogwai-AC: This is a really crude way to collect a player count.  You could have done "UpdateStatus(allplayers.Count);" instead.

Source/ACE/Managers/LandblockManager.cs
```diff
@@ -87,6 +86,8 @@ private static Landblock GetLandblock(LandblockId landblockId, bool propogate)
                     {
                         // load up this landblock
                         Landblock block = new Landblock(landblockId);
+                        block.Load();
```
Mogwai-AC: By having a mandatory function call after the constructor, you leave open the possibility for the Landblock to be in an invalid state: instantiated, but not loaded.  The design of having the constructor do the Loading work was very intentional and should remain that way.  As best I can tell, you broke this design paradigm for the simple purpose of artificially creating more states.  Please revert the constructor changes.  If that means you lose a state, fine, it's a state that shouldn't exist.

Forbiddenz: this was just the start of loading and unloading the landblock.. It can be removed.. it has no bearing on this diag tool

Forbiddenz:
Removed load code, placed that back into its init.
Fixed crude player count.
Let me know.

Source/ACE/Managers/LandblockManager.cs
Forbiddenz: has been removed. =)

Issue #296, opened 2017-04-26
Mana does not update when xp is spent on abilities that should increase it


LtRipley36706:

jensbodal:
Can this issue be closed then?

PR #319, opened 2017-05-03
Fix: Mana on XP Spend

fantoms:
* Fixed character's total `Mana` points after spending experience points on `Self`.
* Noticed that the `@heal` command may prevent vitals increasing.


fantoms:
This pr was created to address issue #296. To address the mana updates, a mana update message is now sent when self is updated.

After changing the `@heal` command to use a different message, the health and mana rankUp work correctly.

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1025,9 +1025,9 @@ public static void HandleHeal(Session session, params string[] parameters)
             session.Player.Stamina.Current = session.Player.Stamina.UnbuffedValue;
             session.Player.Mana.Current = session.Player.Mana.UnbuffedValue;
 
-            var updatePlayersHealth = new GameMessagePrivateUpdateAttribute2ndLevel(session, Vital.Health, session.Player.Health.Current);
-            var updatePlayersStamina = new GameMessagePrivateUpdateAttribute2ndLevel(session, Vital.Stamina, session.Player.Stamina.Current);
-            var updatePlayersMana = new GameMessagePrivateUpdateAttribute2ndLevel(session, Vital.Mana, session.Player.Mana.Current);
```
LtRipley36706: GameMessagePrivateUpdateAttribute2ndLevel is the right message for what this command is doing, I think you've found an error with how we're handling the sequence for this message however. 
vitalupdates

3 different numbers in a row based on this retail pcap...

PR #320, opened 2017-05-03
Portal class changes - Ready for review and possible merge

kdmxp256:
I need someone else(s) to look this code/SQL view over.  It results in a crash in Database.cs because of NULL fields, as a result of the left outer join.  System.DBNull cannot be converted to System.UInt32


Source/ACE.Database/Database.cs
```diff
@@ -350,7 +350,10 @@ public bool ExecuteConstructedGetStatement<T1>(T1 id, Type type, Dictionary<stri
                             {
                                 foreach (var p in properties)
                                 {
-                                    p.Item1.SetValue(instance, commandReader[p.Item2.DbFieldName]);
+                                    if (commandReader[p.Item2.DbFieldName] == DBNull.Value)
+                                        p.Item1.SetValue(instance, (byte)0);
```
Mogwai-AC: p.Item1.SetValue(instance, null);

Source/ACE.Entity/AcePortalObject.cs
```diff
@@ -0,0 +1,68 @@
...
+        [DbField("weenieClassId", (int)MySqlDbType.UInt16)]
+        public override ushort WeenieClassId { get; set; }
+
+        public Position DestPosition
+        {
+            get {
+                if (DestLandblockId == null)
+                    return null;
+                else
+                    return new Position(DestLandblockId.Value, DestPosX, DestPosY, DestPosZ, DestQX, DestQY, DestQZ, DestQW);
```
Mogwai-AC: could cache this instead of creating a new object every time.  this is just inefficient.

Source/ACE.Entity/AcePortalObject.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+
+        [DbField("destLandblockId", (int)MySqlDbType.UInt32)]
+        private uint? DestLandblockId { get; set; }
+
+        [DbField("destX", (int)MySqlDbType.Float)]
+        private float DestPosX { get; set; }
```
Mogwai-AC: given my previous suggestions in Database.cs, all these Dest properties will need to be nullable.

Source/ACE/Factories/GenericObjectFactory.cs
```diff
@@ -26,7 +27,9 @@ public static List<WorldObject> CreateWorldObjects(List<AceObject> sourceObjects
                 }
                 else if ((oDescFlag & ObjectDescriptionFlag.Portal) != 0)
                 {
-                    results.Add(new Portal(aceO));
+                    AcePortalObject acePO = DatabaseManager.World.GetPortalObjectsByAceObjectId(aceO.AceObjectId);
```
Mogwai-AC: This brings up a design flaw I have that I need to address.  Ugh.  This will certainly work, though.

Mogwai-AC:
I tried and failed to fix the merge conflicts :(

PR #325, opened 2017-05-04
database object factory

Mogwai-AC:


ogmage78:
This looks really good.   I had totally missed the significance of how that bit field was used.

PR #327, opened 2017-05-05
Full reset of repo

LtRipley36706:
This is an attempt at using the .gitattributes and .editorconfig settings to correct existing linefeed and whitespace issues.


Mogwai-AC:
Accepting without changelog, but you should have had a changelog!

PR #329, opened 2017-05-05
Dat Updates

OptimShi:
Added a caching system to .dat file reads. 
Added EnvCell FileType to ACE.DatLoader. This is dungeon blocks and indoor spaces (0x0100+ cell locations)
Updated the cell-export function to use DatReader class (if you previously exported these, delete them and start over--they are wrong!)


Mogwai-AC:
fubar'ed by Whitespace PR.  please resolve.

OptimShi:
A couple of pushes later, and the whitespace issues seems to be resolved in this PR. (Note that the bulk of the differences in the existing files are now indentation related, which is by design)

PR #330, opened 2017-05-06
Fix for ci & create complete and tested.   Ready for merge if approved.  

ogmage78:
Fixed
I fixed a lot of the issue with these two create object calls.    I modify some of the flags because the base ace object was made from a pcap - I think the logic is sound and it is how I had it working before.   I also put in some code to exclude the extra information from baseAceObject for portals, creatures and life stones.   We can expand that exclusion list if and when it is found necessary.

I added a new admin function cirand <typeId> that will allow you to easily test and exercise the ci function to insure it works for various itemTypes


Mogwai-AC:
Changelog entry?

ogmage78:
lol that gets me every time.   I am adding now.

Issue #254, opened 2017-04-19
Add support for vitals regeneration


Gordian3:
There is currently no support for a character to regenerate vitals to the current maximum values of the character.   This issue is asking for a mechanism to be added for the server to periodically increment these values and pass them along to the client session owning the character.

Mostly this would be visible after @grantxp and adding to any vital.

TBD:  Do we need to broadcast the vitals increases to other accounts on the landblock (or adjacent landblock)?

This regen mechanism can have the actual formula for the amount of regen or that can be left as a TBD and just use a default value.  If not done for this issue, then please create a new issue.

On logout or other client disconnect, the periodic vital increases should stop.

Do not increase beyond the current maximums (would account for debuffs or vitae if implemented)

Gordian3:
Some info on the Natural Regen rates can be seen at: http://acpedia.org/wiki/Natural_Resistances#Natural_Resistance_Bonus

Mostly this references the modifiers to the natural rates due to attributes (and spells).
I'm not sure where to find the actual natural rate.   Perhaps some math can be done on some pcaps showing vital restoration and we can work back from the characters stats and spells?

Gordian3:
I can't find the official natural regen formula documented.  Following is from a Letter to the Players, Nov 2001.

> b. New regeneration rates--As the game of Asheron's Call grows and progresses, certain design dynamics have become outdated. The hit point regeneration rate has become one of those antiquated systems. We wanted to decrease downtime, especially the downtime for those without life magic who also don't have healing. We think we've struck the right compromise with the current hit point and Stamina regen rates.

ddevec:
Update:

We now have the framework support for regenerating creature vitals (vitals will go up over time), but with completely bogus numbers.  The test hook for this is /setvital.

PR #338, opened 2017-05-13
Time-based updates to creature vitals

ddevec:
Trying to take a small piece for my first commit -- hopefully this is reasonable.

Goal is to add basic functionality required for time based updates of creature's vitals (health, mana, stamina).  This is a subset of the challenge for issue #254 .  Current update rates do not reflect original game's update algorithm.

Summary of changes:
- Added setvital debug function -- sets a vital (hp mp stam) to a specified or relative value
- Added per-creature update function to landblock main loop
- Added vital update code per creature
- Added UI notification messages to player, so UI updates when vitals are changed.

I'm new to working on distributed teams via github, please let me know if there is anything I need to correct.

### DIFF
--- a/Source/ACE/Entity/Player.cs
+++ b/Source/ACE/Entity/Player.cs
@@ -780,7 +780,7 @@ namespace ACE.Entity
                 }
                 else
                 {
-                    abilityUpdate = new GameMessagePrivateUpdateVital(Session, ability, ranks, baseValue, result, character.Abilities[ability].Current);
+                    abilityUpdate = new GameMessagePrivateUpdateVital(Session, ability, ranks, baseValue, result, character.Abilities[ability].MaxValue);
                 }
 
                 // checks if max rank is achieved and plays fireworks w/ special text
@@ -880,7 +880,12 @@ namespace ACE.Entity
 
             if (rankUps > 0)
             {
-                ability.Current += addToCurrentValue ? rankUps : 0u;
+                // FIXME(ddevec): This needs to be done for vitals only? Someone verify -- 
+                //      Really AddRank() should probably be a method of CreatureAbility/CreatureVital
+                CreatureVital vital = ability as CreatureVital;
+                if (vital != null) {
+                    vital.Current += addToCurrentValue ? rankUps : 0u;
+                }
                 ability.Ranks += rankUps;
                 ability.ExperienceSpent += amount;
                 character.SpendXp(amount);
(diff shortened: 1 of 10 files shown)


Source/ACE/Entity/Player.cs
```diff
@@ -1605,5 +1605,30 @@ public void ClearDelayedTeleport()
             DelayedTeleportTime = DateTime.MinValue;
             WaitingForDelayedTeleport = false;
         }
+
+        override public void GameLoopUpdate()
+        {
+            uint oldHealth = Health.Current;
```
Mogwai-TheFurry: I'm ok with this in general, but I had envisioned making an "IsDirty" flag with the CreatureAbility object, then sending an update when it's dirty.

ddevec: I could do that.  I didn't want to add unneeded data to a class which was used for more things than live updates (perhaps there should be a CreatureVital which extends CreatureAbility?).

Forbiddenz: Maybe we should start to add a interface to set the guidelines of things that tick or just add a base overidable tick to wo and call it as needed along with a isdirty flag.

I am not a fan of using tons of separate dictionary locks for each type such as creature, player, moved world objects.
 
I know this is a by much for this pr, but we should think about how this is going to work as the code base gets bigger..  Right now we have player loops, moved item loops and not creature vital loops..  seems to me like it could all be one lock of the list and each item can have the own method for what to do when it's ticked.

ddevec: I was looking through the code base, and I agree with this overall comment of managing how things are ticked/updated.

Right now there is a large amount of concurrency in the code (parallel landblock based/movement based object updates, and network based events).  This code-base is almost certain to develop considerable data-races and other concurrency errors without a good, standardized, encapsulated object update mechanism, such as "Updatable" objects inserted into "Update" queues (with most synchronization either managed by single threaded queues, or implicit locking via the queue handler).

I didn't want to take off too large a bite for my first commit, but its something I would be happy to look into in the future (although I'd imagine this would want further discussion via an issue request?).

Mogwai-TheFurry: Most discussion happens in discord - you will certainly get the most feedback and active discussion there.

To the point of the PR: I had envisioned the "IsDirty" thing because you can just set-and-forget.  It doesn't require locking.  If it sets erroneously, no big deal, and extra value to the client.

Source/ACE/Entity/Player.cs
```diff
@@ -1605,5 +1605,30 @@ public void ClearDelayedTeleport()
...
+            {
+                Session.Network.EnqueueSend(new GameMessagePrivateUpdateVital(Session, Enum.Ability.Stamina, Stamina));
+            }
+
+            if (Mana.Current != oldMana)
+            {
+                Session.Network.EnqueueSend(new GameMessagePrivateUpdateVital(Session, Enum.Ability.Mana, Mana));
+            }
+        }
     }
```
LtRipley36706: Keep in mind that for vitals changing positive or negative in value (IE heal over time) the proper message is GameMessagePrivateUpdateAttribute2ndLevel, not GameMessagePrivateUpdateVital. I think we've recently discovered there are unique sequences associated each vital.

ddevec:
This has some of the changes we discussed.  I didn't make changes sent via a "dirty" flag.  I think there are more important design decisions which inhibit my ability to do so.

For instance, what if multiple messages must be sent based on the stat update (e.g. your health, and the health bar for the character IDing you).  In this instance, when can I clear the dirty flag?  Hopefully once our method of pushing updates is more fleshed out this code can be reworked.

ddevec:
Also, it appears the build failed because the system couldn't pull MySql.data -- I don't think this is related to my code.  Am I reading this incorrectly?

Mogwai-TheFurry:
You forgot to put an entry in the changelog.  Accepting, but please submit another PR with it.

PR #341, opened 2017-05-16
Object Overhaul - ready for review and merge.    Still Work in Progress but a good spot.

ogmage78:
* Continued work on Overhaul 
* Created mapping cross reference file to facilitate documentation and cleaning up names - work in progress https://goo.gl/eaaNQb
* Created new database schema for world_object - modified mysqlinstall.bat to reflect new baseline
* created script to ETL old schema to new schema.   I have not included it as we will be refactoring the initial data load to use the new
* schema according to Ripley.   I have the script tested and posted on my dropbox if anyone wants it.   I also have a complete data dump that
* zips down to a manageable size.
* BaseAceObject is now using the new schema
* I re-hooked up telepoi to use the new schema.   Still to do, convert to constructed statement
* @moveto works which shows we can spawn world objects out of new data schema.
* Next steps - refactor AceObject and make sure world objects load again.
* re-enable and expand constructed statement test.   We have a constraint that the old test would fail.
* Once that is complete - start same process using clone of this schema for character as discussed.


Source/ACE.Entity/BaseAceObject.cs
```diff
@@ -5,12 +5,60 @@
...
+        public uint PhysicsTableId { get; set; }
+
+        [DbField("soundTableId", (int)MySqlDbType.UInt32)]
+        public uint SoundTableId { get; set; }
+
+        [DbField("header", (int)MySqlDbType.UInt32)]
+        public uint WeenieHeaderFlags { get; set; }
+
+        [DbField("spellId", (int)MySqlDbType.UInt16)]
+        public ushort SpellId { get; set; }
```
LtRipley36706: wonder if this should be a offshoot table, seen any examples of items with spellid being multiples?

Source/ACE/Entity/WorldObject.cs
```diff
@@ -160,101 +160,101 @@ public virtual void SerializeCreateObject(BinaryWriter writer)
             if ((WeenieFlags & WeenieHeaderFlag.PuralName) != 0)
                 writer.WriteString16L(GameData.NamePlural);
 
-            if ((WeenieFlags & WeenieHeaderFlag.ItemCapacity) != 0)
-                writer.Write(GameData.ItemCapacity);
+            if (((WeenieFlags & WeenieHeaderFlag.ItemCapacity) != 0) && GameData.ItemCapacity != null)
```
LtRipley36706: the way I read this (and the others belows) is if ItemCapacity is one of the flags in the WeenieFlag AND GameData.ItemCapacity is not null then write the data, is that right? If so, I believe this will lead the client to choke on the message because the WeenieFlag is sent telling it to expect ItemCapacity but then we provide it no data (or wrong data from another later write.

Source/ACE.Entity/BaseAceObject.cs
LtRipley36706: if not needed to make a separate table PropertyInt.ItemSpellcraft _might_ be the corresponding property

Source/ACE.Entity/BaseAceObject.cs
```diff
@@ -5,12 +5,60 @@
...
-    [DbGetList("base_ace_object", 15, "typeId")]
+    [DbTable("ace_object")]
+    [DbGetList("ace_object", 15, "typeId")]
     public class BaseAceObject
     {
-        [DbField("baseAceObjectId", (int)MySqlDbType.UInt32, Update = false, IsCriteria = true)]
+        [DbField("AceObjectId", (int)MySqlDbType.UInt32, Update = false, IsCriteria = true)]
         public virtual uint AceObjectId { get; set; }
+
+        [DbField("bitField", (int)MySqlDbType.UInt32)]
```
ogmage78: Thanks for taking the time to review and good catch.   I will fix it.   As you can tell - I am not yet actually pulling the data or that would have blown up.   :)

Source/ACE.Entity/BaseAceObject.cs
```diff
@@ -5,12 +5,60 @@
...
+        /// <summary>
+        /// PhysicsData.CSetup
+        /// </summary>
+        [DbField("modelTableId", (int)MySqlDbType.UInt32)]
+        public uint ModelTableId { get; set; }
+
+        [DbField("motionTableId", (int)MySqlDbType.UInt32)]
+        public uint MotionTableId { get; set; }
+
+        [DbField("physicsBitField", (int)MySqlDbType.UInt32)]
```
ogmage78: Will fix.

Source/ACE.Entity/BaseAceObject.cs
```diff
@@ -5,12 +5,60 @@
...
+        public uint PhysicsDescriptionFlag { get; set; }
+
+        [DbField("physicsScript", (int)MySqlDbType.UInt16)]
+        public ushort PhysicsScript { get; set; }
```
ogmage78: I misread that - I will rename the database field. and refactor in code to PlayScript.

Source/ACE.Entity/BaseAceObject.cs
ogmage78: We need to run this one down for sure.   I bet you are right - I bet it is spellcraft.   Can you see if you can tell how it is used in the client?

Source/ACE/Entity/WorldObject.cs
ogmage78: Well I might be wrong - but first - the binary writer will not write a null.   The first part of that were we are checking the weenieHeaderFlag against the specific mask != 0 means the flag is set and we want to write out the value.   I just had to add the check to make sure we are in fact not trying to write a null.   I don't think that is wrong.   Can someone else verify how that should be handled?   I either need to add that check or assert that it is not null.   Please advise.

fantoms: `(WeenieFlags & WeenieHeaderFlag.ItemCapacity) != 0) && GameData.ItemCapacity != null` requires that Weenieflag must be set, and the business object must be a non-null value for the writer to send data.

Ripley is stating that the client may be expecting this, if the flag is sent in a different writer.write packet in the inherited object:

line 149.

ogmage78: I must just be missing what you guys are driving at - can either of you give a concrete example?   If the flag for Item Capacity is set - you have to send something non null.   If you unset the flag - it does not matter what is in gamedata.ItemCapacity - it is not going to get sent to the client.   That is how it was working before I did anything to it.   The flag controls the write on our end and tells the client what to expect to read on the client side.   Sorry for being obtuse - but I don't understand what you guys are driving at.

Source/ACE.Entity/BaseAceObject.cs
```diff
@@ -107,66 +155,181 @@ public class BaseAceObject
...
+            {
+                return (byte?)AceObjectPropertiesInt.Find(x => x.IntPropertyId == (uint)PropertyInt.MaterialType)?.PropertyValue;
+            }
+            set
+            {
+                var listItem = AceObjectPropertiesInt.Find(x => x.IntPropertyId == (uint)PropertyInt.MaterialType);
+                if (value != null)
+                    AceObjectPropertiesInt.Find(x => x.IntPropertyId == (uint)PropertyInt.MaterialType).PropertyValue = (byte)value;
+                else
+                    if (listItem != null)
```
Mogwai-TheFurry: when the "else" is more than 1 line, as in this case, you need to use "{" "}" brackets.  that, and the "if" path needs indentation.  Pro tip: Ctrl+K, Ctrl+D to autoformat an entire file (don't do this in aligned enums tho)

Source/ACE/Entity/Door.cs
```diff
@@ -60,7 +60,7 @@ public Door(AceObject aceO)
             aceO.AnimationOverrides.ForEach(ao => this.ModelData.AddModel(ao.Index, ao.AnimationId));
             aceO.TextureOverrides.ForEach(to => this.ModelData.AddTexture(to.Index, to.OldId, to.NewId));
             aceO.PaletteOverrides.ForEach(po => this.ModelData.AddPalette(po.SubPaletteId, po.Offset, po.Length));
-            this.ModelData.PaletteGuid = aceO.PaletteId;
+            this.ModelData.PaletteGuid = (uint)aceO.PaletteId;
```
Mogwai-TheFurry: red flag for data type problems.  smells like packed Dword territory.

Source/ACE/Entity/GameData.cs
```diff
@@ -1,63 +1,57 @@
...
-using ACE.Network.GameMessages.Messages;
-using ACE.Network.Managers;
-using System.IO;
+﻿using ACE.Network.Enum;
 
 namespace ACE.Entity
 {
+    /// <summary>
+    /// GameData - which of these items get written out is controled by DescriptionFlags
+    /// </summary>
```
Mogwai-TheFurry: // TODO: replace with exiting data in AceObject rather than copying it to this object

Source/ACE/Entity/ModelData.cs
```diff
@@ -11,13 +11,13 @@ namespace ACE.Entity
     /// </summary>
```
Mogwai-TheFurry: /// TODO: convert to using AceObject backed data and get rid of the redundant copies of all the datum.

Source/ACE/Entity/ModelData.cs
```diff
@@ -60,8 +60,8 @@ public void Serialize(BinaryWriter writer)
             writer.Write((byte)modelTextures.Count);
             writer.Write((byte)models.Count);
 
-            if (modelPalettes.Count > 0)
-                writer.WritePackedDwordOfKnownType(PaletteGuid, 0x4000000);
+            if ((modelPalettes.Count > 0) && (PaletteGuid != null))
+                writer.WritePackedDwordOfKnownType((uint)PaletteGuid, 0x4000000);
```
Mogwai-TheFurry: // TODO: create PackedDwordType enum with constant values for offsets.

Source/ACE/Entity/WorldObject.cs
Mogwai-TheFurry: You'll have to apply this to all of these.  The basic issue is that we have flags that are set in bulk.  **Flags should be calculated based on the presence and absence of data, not arbitrarily set and then validated after being written**.

```cs
if ((WeenieFlags & WeenieHeaderFlag.ItemCapacity) != 0)
{
    if (GameData.ItemCapacity != null)
    {
        // write out items
    }
    else
    {
        // we wrote the flag in the header, failing to write at least SOMETHIGN will cause the client to barf on the packet.
        // let's try writing a 0 to test it
       writer.Write((uint)0);
    }
}

File: Source/ACE/Network/Enum/WeenieHeaderFlags.cs [hunk 11]
```diff
@@ -29,7 +29,7 @@ public enum WeenieHeaderFlag : uint
         BlipColour        = 0x00100000,
         Burden            = 0x00200000,
         Spell             = 0x00400000,
-        Radar             = 0x00800000,
+        RadarBehavior             = 0x00800000,
```
Mogwai-TheFurry: whitespace

Source/ACE/Entity/WorldObject.cs
LtRipley36706: the problem the calc'ing flags (at this point) is we inserted 0s for anything that didn't have data, i could see in at least some instances where you both needed the flag set and wanted to send the 0.

Database/Base/00_ace_world2.sql
```diff
@@ -0,0 +1,231 @@
...
+  `aceObjectDescriptionFlags` int(10) unsigned NOT NULL,
+  `animationFrameId` int(10) unsigned NOT NULL,
+  `currentMotionState` text NOT NULL,
+  `iconId` int(10) unsigned NOT NULL,
+  `iconOverlayId` int(10) unsigned NOT NULL,
+  `iconUnderlayId` int(10) unsigned NOT NULL,
+  `modelTableId` int(10) unsigned NOT NULL,
+  `motionTableId` int(10) unsigned NOT NULL,
+  `physicsDescriptionFlag` int(10) unsigned NOT NULL,
+  `physicsScript` smallint(5) unsigned NOT NULL,
```
LtRipley36706: change to playScript

LtRipley36706:
perhaps its now safe and at the point to move everything over to normalize ace_world?

PR #344, opened 2017-05-18
Event Handling Restructure

ddevec:
Hello.  My first commit was a small vital update change.  I've now broken out of my shell, and am looking at the bigger picture!  I hope this is useful, or inspires brainstorming future directions for this codebase!

This is an experimental and pretty major restructuring of our current event handling system.

This restructuring accomplishes several **goals**:

- A consistent and easily followed synchronization mechanism for cross-landblock operations
- A convenient mechanism to broadcast position based updates, without requiring objects maintain broadcast lists
- A simplification of the rapidly growing and complicated state machines used to manage character actions.

To achieve these goals I have restructured the **logical flow** of the code in the following ways:

- I have removed the prior player event queues.
- There is an ITickable interface, every landblock update loop all ITickables on the landblock are "Ticked"
- It is the responsibility of the ITickable to manage any updates and broadcasts it needs within its tick.
- To aid in the management of these updates, I introduce "ObjectActon" co-routines which allow a convenient way to structure the traditionally event-based code in an Object-Oriented (as opposed to event-based) manner.
- I created a (hopefully) simple-to-follow synchronization pattern, which can allow complex behaviors within 

actions between objects.

**Smaller Bits**:
Flow of information:
Network packets -> session.Player.RequestAction(action) - (on landblock tick) -> action()

- On Tick() ITickables are given (lazy populating) broadcast lists, letting them know nearby players that need to be notified of global updates.  
- All ITickables are locked by the landblock before ticking.  If they need to interact with another object, they must lock that object before using it (but the common-case self locking is handled by the landblock).

Please take a look, try it out, tell me if you have any questions!

Important Parts to look at:
Actions are now handled much more simply.  For example: before delayed teleport required a messy state machine with tendrils all over the code.  Now the delayed teleport are implemented in ActDelayedTeleport, and encapsulated cleanly.

Feel free to try checking the branch out and building an event for yourself (e.g. proper animation delays when dropping/picking up items).  Hopefully you'll find it is significantly easier to use.

Finally, as far as I know, all functionality on the mainline ACE emulator (as well as some extra, particularly multi-client operations) functions.


Source/ACE.Entity/Position.cs
```diff
@@ -132,6 +132,21 @@ public Position(uint characterId, PositionType type, uint newCell, float newPosi
             RotationW = newRotationW;
         }
 
+        public Position(Position pos)
+        {
+            this.LandblockId = pos.LandblockId;
+            this.CharacterId = pos.CharacterId;
+            this.PositionType = pos.PositionType;
+            this.Cell = pos.Cell;
```
Mogwai-TheFurry: this ought to be redundant with LandblockId

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -418,7 +418,7 @@ public static void HandleHome(Session session, params string[] parameters)
                 }
                 // If we have the position, teleport the player
                 if (session.Player.Positions.ContainsKey(positionType)) {
-                    session.Player.Teleport(session.Player.Positions[positionType]);
+                    session.Player.RequestAction(() => session.Player.ActTeleport(session.Player.Positions[positionType]));
```
fantoms: Small harmless note - I wanted to let you know that others have stated in the past that: long-hand names are preferred over shorthand names, for readability.

Source/ACE/Entity/Landblock.cs
```diff
@@ -543,19 +455,36 @@ public void UseTime()
...
+                    // We lock here -- it IS required becuase this is run in a Parallel.ForEach loop
+                    //   AND because of cross-landblock dependencies
+                    //   Any use of OTHER objects within the tick function (gotten through landblock by guid) requires locking
+                    //   updates of THIS object within Tick are safe
+                    var tickable = wo as ITickable;
+                    if (tickable != null)
+                    {
+                        lock (tickable)
+                        {
+                            // FIXME(ddevec): LBBroadcaster class was a fast hacky fix to the problem -- there is probably a more elegant solution..
```
Mogwai-TheFurry: given that the locked object is _inside_ the ForEach, this lock is useless.  you need to declare it outside the Parallel.ForEach.  The other issue that I don't see is what objects in this loop are mutable.  I tried to track down what part of this loop isn't thread-safe, but I couldn't find it.  Locking inside a parallel.ForEach just feels really dangerous.

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -140,10 +142,11 @@ public static void HandlePlaySound(Session session, params string[] parameters)
...
                     {
                         message = $"Playing sound {Enum.GetName(typeof(Network.Enum.Sound), sound)}";
-                        // add the sound to the player queue for everyone to hear
-                        // player action queue items will execute on the landblock
-                        // player.playsound will play a sound on only the client session that called the function
-                        session.Player.ActionApplySoundEffect(sound, session.Player.Guid);
+                        // Lock the player and play the sound...
+                        lock (session.Player)
+                        {
+                            session.Player.PlaySound(sound, session.Player.Guid);
```
fantoms: Does this limit the sound to the player only?

Source/ACE/Entity/Landblock.cs
ddevec: I'm not sure if I understand how locking inside of a Parallel.ForEach statement is useless.  The objects within the loop may be interacting with each other, e.g. a monster may attack a player and adjust her health, and the player may be regenerating that health.   Unless Parallel.ForEach does some black magic, that would result in a data-race, requiring locking.

I didn't think of the danger of deadlock when I wrote this code.  That is a very important point and I don't have a method of mitigating it.  Mitigating lock-based deadlocks is traditionally done by enforcing lock ordering, which would be difficult to do with this design...

Source/ACE/Command/Handlers/DebugCommands.cs
Forbiddenz: I don't think so, but keep in mind sound had a object guid attached to it but player is the one that need to hear it, it plays on another object on their client

Forbiddenz:
The tickable is a second lock, Its a overall change to how things are managed.  Now that each object is in charge of itself, you neeed to lock the wo itself instead of the list.  it does logical sounds a little tricky but I think I know why hes locking each one.  Because the wo now can be que functions and it may be Queing a function when it's accessed inside the dictionary.

Forbiddenz:
Tech I think you need two locks, one of the list and one per wo being processed in the game loop..  hence the foreach lock

Source/ACE/Entity/Landblock.cs
ddevec: I think I can switch from my current lock-based synchronization to a message based one and keep the co-routine interface, it'll be a little more tricky to implement than this (flawed) locking system.  I'll see what happens if I play around with the idea.

Mogwai-TheFurry: I had the impression that the locker object inside the loop would be declared/instantiated per thread.  Therefore, locking it would be like locking any other non-shared resource: a no-op.  I could very well be wrong, but that's always been my impression of how .Net executes lambdas.

Source/ACE/Entity/Player.cs
```diff
@@ -1419,54 +1351,327 @@ private void SendSelf()
...
+            yield return ObjectAction.CallCoroutine(ActMoveTo(item));
         }
 
-        private void DelayedUpdatePosition(Position newPosition)
+
+        public IEnumerator ActMoveTo(WorldObject wo)
         {
-            var t = new Thread(() =>
+            Position desiredLocation = null;
+            lock (wo)
```
Mogwai-TheFurry: this is almost certainly a deadlock, and has serious risks even if not.

Source/ACE/Entity/Player.cs
```diff
@@ -1419,54 +1351,327 @@ private void SendSelf()
...
+            {
+                NotifyAndDropItem(wo.Guid);
+            }
+            yield break;
+        }
+
+        public IEnumerator ActMoveTo(ObjectGuid itemGuid)
+        {
+            var item = LandblockManager.GetWorldObject(Session, itemGuid);
```
Mogwai-TheFurry: LandblockManager.GetWorldObject is either going to be deprecated or modified to return a new readonly object.  locking the result of this is not viable.

ddevec:
I"m closing this.  I misunderstood the underlying interface, and consequently the mechanism I implemented was flawed.

PR #347, opened 2017-05-20
Spell Tables

OptimShi:
Added SpellTable/SpellComponent parsing from portal.dat. Also added "@learnspell" debug command and corresponding UpdateSpell Event.


OptimShi:
I fixed the merge errors to crept into this PR... Should be good-to-go now.

PR #349, opened 2017-05-21
Object overhaul - part two.   Review and Comment

ogmage78:
I am having to refactor and touch a LOT of stuff.   It is looking good to me - but I am concerned about how much I am modifying without review.   Below is where I am at.   
### 2017-05-21
* Continued work on Overhaul 
* re-enabled more constructed statements
* Most of the work to use existence of data to set flags for both physics and weenieHeaderFlags - some cleanup remains - either logic issues or bad data
* modified cirand to take an optional second parameter to spawn X items at a time.   This speeds up testing.   I set a default to 10 if you omit the second parameter.
* base_ace_object and ace_object both seem to be loading.
* updated or created several views.
* have not tested with Ripley's new data export - I did not want to add a second variable to testing.   
* Next steps - use Ripley's data, create needed views, re-enable door, portal, NPC --- all the world things.


Database/Updates/World/01_05_21_2017_viewUpdate.sql
```diff
@@ -0,0 +1,227 @@
...
+/*!40111 SET @OLD_SQL_NOTES=@@SQL_NOTES, SQL_NOTES=0 */;
+CREATE DATABASE /*!32312 IF NOT EXISTS*/`ace_world` /*!40100 DEFAULT CHARACTER SET utf8 */;
+
+USE `ace_world`;
+
+/*Table structure for table `vw_ace_object` */
+
+DROP TABLE IF EXISTS `vw_ace_object`;
+
+/*!50001 DROP VIEW IF EXISTS `vw_ace_object` */;
```
Mogwai-TheFurry: Please get rid of the /*50001 garbage syntax.  If you run this script on an earlier version of MySql, it **SHOULD** fail.

Source/ACE.Entity/AcePortalObject.cs
```diff
@@ -4,22 +4,19 @@
 namespace ACE.Entity
 {
     [DbTable("vw_ace_portal_object")]
-    [DbGetList("vw_ace_portal_object", 14, "baseAceObjectId")]
+    [DbGetList("vw_ace_portal_object", 14, "AceObjectId")]
     public class AcePortalObject : AceObject
     {
```
Mogwai-TheFurry: it's entirely possible this object just goes away if we expose positions in AceObject itself.  We just have the portal code look for the proper position type.

Database/Updates/World/01_05_21_2017_viewUpdate.sql
ogmage78: Google and the mysql help file have failed me here.   No one seems to like them but no one has a way to turn them off.   I can try and find a parser to strip them - but it does not look like from my reading that there is any setting I can turn them off.   Do you have any idea how to kill them or do I need to hand edit any scripts?

Source/ACE.Entity/AcePortalObject.cs
ogmage78: Yea - I was just trying to get what we had working again - the only thing special about portals aside from other objects that can have position is they have both a position and a destination.   That can be done in generic handling code for sure.

Database/Updates/World/01_05_21_2017_viewUpdate.sql
kdmxp256: I think you may have to hand edit the scripts, as these pseudo-comments are acceptable syntax, as far as MySQL and MariaDB are concerned.

Mogwai-TheFurry:
You guys are in this branch more than I am - I'm happy to merge if everybody else likes it.

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1023,16 +1025,16 @@ public static void HandleCI(Session session, params string[] parameters)
                 return;
             }
             var loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
```
fantoms: The loot generation needs to be move up into the Try, to prevent invalid weenies from crashing the server.

fantoms: The loot generation needs to be move up into the Try, to prevent invalid weenies from crashing the server.

ogmage78:
Got to make a change to how I am looking at landblock.   Going to close this and open a new one.

PR #354, opened 2017-05-25
World Object Refactor & Code Cleanup.

ogmage78:
* Refactored World Object to have all of the properties and methods of Game Data, Physics Data and Model Data.

* Since I had to touch a ton of files, I cleaned up as much as I could.   Removed unneeded using statements,
* killed whitespace where I found it, removed this. where not needed and where the type was obvious switched to a more
* consistent use of var.
* Re-enabled the unit tests for database.   We don't have a delete = if you insert into the table, you have to manually kill
* the test data to avoid a duplicate key violation on a second running of the tests.    If those run in AppVeyor - will not be an issue as
* an fresh db is created each time.


ogmage78:
Opps - going to try again.

PR #359, opened 2017-06-03
First corrections and commenting out so the start to command line works

Lidefeath:


LtRipley36706:
I think this was also in #360 so closing for now, if not let me know.

PR #366, opened 2017-06-05
Extend ORM to allow aggregate functions i.e. MAX(column) and

Lidefeath:
implemented GetNextCharacterId using the MAX aggregate


ogmage78:
This looks good!

PR #369, opened 2017-06-06
Read character creation values from client_portal.dat 

OptimShi:
and assign to appropriate Character Properties


Source/ACE/Network/Handlers/CharacterHandler.cs
```diff
@@ -124,33 +124,107 @@ public static void CharacterRestore(ClientMessage message, Session session)
...
             character.Gender = reader.ReadUInt32();
-            Appearance appearance = Appearance.FromNetowrk(reader);
+            Appearance appearance = Appearance.FromNetwork(reader);
 
             // pull character data from the dat file
             SexCG sex = cg.HeritageGroups[(int)character.Heritage].SexList[(int)character.Gender];
-            
+
             character.MotionTableId = sex.MotionTable;
+            character.SetDataIdProperty(PropertyDataId.MotionTable, character.MotionTableId);
```
Mogwai-TheFurry: This should be redundant.  The MotionTableId ought to (under the hood) call SetDataIdProperty (and as far as I can tell, it does:
https://github.com/ACEmulator/ACE/blob/object-overhaul/Source/ACE.Entity/AceObject.cs, line 108

Same applies to all the properties below.

Source/ACE/Network/Handlers/CharacterHandler.cs
```diff
@@ -124,33 +124,107 @@ public static void CharacterRestore(ClientMessage message, Session session)
...
+
+            // Certain races (Undead, Tumeroks, Others?) have multiple body styles available. This is controlled via the "hair style".
+            if (hairstyle.AlternateSetup > 0)
+            {
+                character.ModelTableId = hairstyle.AlternateSetup;
+                character.SetDataIdProperty(PropertyDataId.Setup, character.ModelTableId);
+            }
 
-            // character.SetDataIdProperty(PropertyDataId.HeadObject, appearance.HairStyle);
+            character.SetDataIdProperty(PropertyDataId.EyesTexture, sex.GetEyeTexture(appearance.Eyes, isBald));
```
Mogwai-TheFurry: I'd REALLY like to see a new "EyesTexture" property on AceCharacter that does this on set, as the properties above do.  Same is true for all of these.

Source/ACE/Network/Handlers/CharacterHandler.cs
```diff
@@ -124,33 +124,107 @@ public static void CharacterRestore(ClientMessage message, Session session)
...
-            // character.SetDataIdProperty(PropertyDataId.MouthTexture, appearance.Mouth);
+            character.SetDataIdProperty(PropertyDataId.CombatTable, character.CombatTableId);
 
-            // character.SetDataIdProperty(PropertyDataId.HairPalette, appearance.HairColor);
-            // character.SetDataIdProperty(PropertyDataId.EyesPalette, appearance.EyeColor);
-            // character.SetDataIdProperty(PropertyDataId.SkinPalette, appearance.EyeColor);
+            // Check the character scale
+            if (sex.Scale != 100u)
+            {
+                character.SetDoubleProperty(PropertyDouble.DefaultScale, (sex.Scale / 100f)); // Scale is stored as a percentage
```
Mogwai-TheFurry: I'd REALLY like to see a new "DefaultScale" property on AceCharacter that does this on set.

OptimShi: I think I can do that... Looks simple enough, really.

OptimShi:
The requested changes have been made. Thanks for the feedback, I didn't dive that deep into how the AceObject/AceCharacter functioned, so your requests made me do that and learn a little more about how it works (which will be useful for when we need to take these visual properties and add them to the overrides on load)

PR #370, opened 2017-06-08
More fun in Zombieland - character loading.   

ogmage78:
I am hoping you will give me a pass on what looks like massive changes to ShardBase.sql - I think Ripley is using mysqlworkbench for his export and I am using Sqlyog.   The changes are minor and have been tested to work.

* Continued work loading character from ace_shard.   Rebased
* Cleaned up whitespace issues where I found them.
* Added cascade delete to all child tables in ace_shard.
* Completed and tested up to skillz loading.   Made a few changes on the db side for key consistancy.
* TODO items:
* Finish up loading - spells, friends, allegiance info, spell comps, spell bars etc. etc....


fantoms:
Note for MariaDB Users: I had to upgrade past 10.2.0, to make the SQL work. As @ogmage78 pointed out, version 10.2.1 is now needed.

Until 10.2.1:
`The STORED keyword is not supported.`

https://mariadb.com/kb/en/mariadb/virtual-computed-columns/

ogmage78:
Just a note - we can removed stored if the upgrade requirement is too heavy a lift.

Database/Base/ShardBase.sql
```diff
@@ -313,251 +245,168 @@ CREATE TABLE `ace_position` (
...
-  CONSTRAINT `FK_character_friends_friendId_character_guid` FOREIGN KEY (`friendId`) REFERENCES `character` (`guid`),
-  CONSTRAINT `FK_character_friends_id_character_guid` FOREIGN KEY (`id`) REFERENCES `character` (`guid`)
-  */
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
-/*!40101 SET character_set_client = @saved_cs_client */;
+  KEY `idx_landblock_raw` (`landblockRaw`),
+  KEY `idx_landblock` (`landblock`),
+  KEY `idx_cell` (`cell`),
+  CONSTRAINT `fk_position_ao` FOREIGN KEY (`aceObjectId`) REFERENCES `ace_object` (`aceObjectId`) ON DELETE CASCADE
+) ENGINE=InnoDB AUTO_INCREMENT=43262 DEFAULT CHARSET=utf8;
```
Mogwai-TheFurry: AUTO_INCREMENT=43262 ???

Database/Base/ShardBase.sql
```diff
@@ -313,251 +245,168 @@ CREATE TABLE `ace_position` (
...
--- Temporary view structure for view `vw_ace_object`
---
+/*Table structure for table `vw_ace_character` */
+
+DROP TABLE IF EXISTS `vw_ace_character`;
+
+/*!50001 DROP VIEW IF EXISTS `vw_ace_character` */;
+/*!50001 DROP TABLE IF EXISTS `vw_ace_character` */;
+
+/*!50001 CREATE TABLE  `vw_ace_character`(
```
Mogwai-TheFurry: comment shenanigans

Database/Base/ShardBase.sql
```diff
@@ -313,251 +245,168 @@ CREATE TABLE `ace_position` (
...
+ `LandblockRaw` int(10) unsigned ,
+ `posX` float ,
+ `posY` float ,
+ `posZ` float ,
+ `qW` float ,
+ `qX` float ,
+ `qY` float ,
+ `qZ` float 
+)*/;
+
```
Mogwai-TheFurry: is vw_ace_object still a thing?  can it be removed?  Same for many of the tables below.

Database/Base/ShardBase.sql
```diff
@@ -313,251 +245,168 @@ CREATE TABLE `ace_position` (
...
... `accountId`,`aops`.`propertyValue` AS `NAME`,`aopb`.`propertyValue` AS `deleted`,`aopbi`.`propertyValue` AS `deleteTime`,`ao`.`weenieClassId` AS `weenieClassId`,`awc`.`weenieClassDescription` AS `weenieClassDescription`,`ao`.`aceObjectDescriptionFlags` AS `aceObjectDescriptionFlags`,`ao`.`physicsDescriptionFlag` AS `physicsDescriptionFlag`,`ao`.`weenieHeaderFlags` AS `weenieHeaderFlags`,`aopi`.`propertyValue` AS `itemType` from ((((((`ace_object` `ao` join `ace_weenie_class` `awc` on((`ao`.`weenieClassId` = `awc`.`weenieClassId`))) join `ace_object_properties_string` `aops` on(((`ao`.`aceObjectId` = `aops`.`aceObjectId`) and (`aops`.`strPropertyId` = 1)))) join `ace_object_properties_bool` `aopb` on(((`ao`.`aceObjectId` = `aopb`.`aceObjectId`) and (`aopb`.`boolPropertyId` = 9001)))) join `ace_object_properties_int` `aopi` on(((`ao`.`aceObjectId` = `aopi`.`aceObjectId`) and (`aopi`.`intPropertyId` = 1)))) join `ace_object_properties_bigint` `aopbi` on(((`ao`.`aceObjectId` = `aopbi`.`aceObjectId`) and (`aopbi`.`bigIntPropertyId` = 9001)))) join `ace_object_properties_iid` `aopiidacc` on(((`ao`.`aceObjectId` = `aopiidacc`.`aceObjectId`) and (`aopiidacc`.`iidPropertyId` = 9001)))) */;
```
Mogwai-TheFurry: it was so pretty before!

Source/ACE.Entity/AceObject.cs
```diff
@@ -770,6 +773,13 @@ public void SetStringProperty(PropertyString propertyId, string value)
 
         public List<AceObjectPropertiesString> StringProperties { get; set; } = new List<AceObjectPropertiesString>();
 
+        public List<AceObjectPropertiesAttribute> AceObjectPropertiesAttributes { get; set; } = new List<AceObjectPropertiesAttribute>();
```
Mogwai-TheFurry: If we're going to add these, we need the reconcile them with the CreatureAbility stuff.  Please add a TODO comment for it.

Source/ACE.Entity/Position.cs
```diff
@@ -1,38 +1,38 @@
...
+using ACE.Entity.Enum;
+using ACE.Common;
+using MySql.Data.MySqlClient;
 
+namespace ACE.Entity
+{
     [DbTable("ace_position")]
     [DbGetList("ace_position", 23, "aceObjectId")]
-    public class Position
+    public sealed class Position
```
Mogwai-TheFurry: i don't disagree, but why?

Source/ACE.Entity/Position.cs
```diff
@@ -1,38 +1,38 @@
...
+            get { return landblockId.Raw != 0 ? landblockId : new LandblockId(Cell); }
             set
             {
                 landblockId = value;
             }
         }
 
         [DbField("aceObjectid", (int)MySqlDbType.UInt32, Update = false, IsCriteria = true)]
-        public virtual uint AceObjectId { get; set; }
+        public uint AceObjectId { get; set; }
```
Mogwai-TheFurry: I'd prefer this to be nullable, because (logically speaking) a position can exist without a corresponding AceObject.  A position is just a set of coordinates - this property is something we have to represent an optional relationship.

Source/ACE.Entity/Position.cs
```diff
@@ -172,13 +172,13 @@ public Position(float northSouth, float eastWest)
             uint baseX = (uint)(eastWest + 0x400);
             uint baseY = (uint)(northSouth + 0x400);
 
-            if (baseX < 0 || baseX >= 0x7F8 || baseY < 0 || baseY >= 0x7F8)
+            if (baseX >= 0x7F8 || baseY >= 0x7F8)
                 throw new Exception("Bad coordinates");  // TODO: Instead of throwing exception should we set to a default location?
```
Mogwai-TheFurry: I'm ok with the throw.  It's a legit bug.  Throw early and ugly.  Quietly swallowing errors allows bugs to become sneaky.

Database/Base/ShardBase.sql
ogmage78: not sure how that got in there.   I will fix

Database/Base/ShardBase.sql
ogmage78: I contacted sqlyog support this is their response:

No, there is not.

"/*!40101 .. */;" will be treated like a comment by the server if server version is <4.0.1. If server version is > 4.0.1 the server ill treat it like an executable statement. It is simply necessary to use such 'version-dependent comments' unless the client is designed for a specific server version only (what SQLyog is not) and script generated is knon to be run only against a specific server version. You can check with 'mysqldump' or any official client by MySQL/Oracle/MariaDB and they do exactly the same.

You may refer MySQL documentation at https://dev.mysql.com/doc/refman/5.7/en/comments.html "If you add a version number after the ! character, the syntax within the comment is executed only if the MySQL version is greater than or equal to the specified version number"

Database/Base/ShardBase.sql
ogmage78: yea - I had hand edited it and it was blowing up everyone for some reason.   I can try again - let me know.   I hate the way this generates - but the sqlyog guys say that is part of mysql.   I think we are swimming upstream on this one....

Source/ACE.Entity/AceObject.cs
ogmage78: will do

Source/ACE.Entity/Position.cs
ogmage78: mainly because refactor was barking at me as a suggestion.   I forget the exact wording but something like virtual method called in constructor?   I can put it back - it seems like a reasonable suggestion to me so I did it.

Source/ACE.Entity/Position.cs
ogmage78: yea - that is an oversight on my part.   I know that needs to be nullable.   I will fix.

Source/ACE.Entity/Position.cs
ogmage78: someone else's comment I think - I removed the checks for less than 0 as it always evaluated to false

Database/Base/ShardBase.sql
ogmage78: I found out how that happened and I fixed it but it is a FYI note to anyone doing a schema export - it you have data - and any of your tables use auto_increment - then it takes the highest current pk value in your db as the seed.   You either need to export schema from a empty db or remember to fix that.  just FYI

Source/ACE.Entity/Position.cs
ogmage78: Fixed

Database/Base/ShardBase.sql
Mogwai-TheFurry: I understand it is syntactically correct.  But it's functionally devastating to have a partial database.  IT NEEDS TO ERROR.

Source/ACE.Entity/Position.cs
```diff
@@ -22,7 +22,7 @@ public LandblockId LandblockId
         }
 
         [DbField("aceObjectid", (int)MySqlDbType.UInt32, Update = false, IsCriteria = true)]
-        public uint AceObjectId { get; set; }
+        public virtual uint AceObjectId { get; set; }
```
Mogwai-TheFurry: virtual doesn't make it nullable.  "int?" does.

ogmage78: I will add that next PR.   Sorry missed it.

PR #372, opened 2017-06-09
Cleanup of DbGetListAttribute

ddevec:
Adjusted DbGetList Attribute to be DbList.  Removed statementId from the attribute, it is now taken from the Id passed on ConstructStatement in ShardDatabase/Database.cs.  Cleaned up hacky Ids I committed earlier.

Also made minor changes to remove the tons of formatting warnings we were getting (one still remains on unimplemented async function).


ddevec:
Forgot to merge master back in -- will re-submit in a minute

PR #373, opened 2017-06-09
Fix the "You do not own this character" error.

Lidefeath:
With this fixes the view vw_ace_character still works with inner joins.


Source/ACE/Network/Handlers/CharacterHandler.cs
```diff
@@ -267,6 +267,10 @@ public static void CharacterRestore(ClientMessage message, Session session)
             }
 
             character.AccountId = session.Id;
+            character.Deleted = false;
+            character.DeleteTime = 0;
+            character.WeenieClassId = 1;
+            character.ItemType = 1;
```
Mogwai-TheFurry: We should consider putting these in the constructor.

PR #375, opened 2017-06-11
Refactoring, save/restore progress on object overhaul

ddevec:
This is an extension of Og II's current pull request.

Things I've added:
Separation of AceObjectProperties{Attribute, Attribute2nd, Skill} from CreatureAbility and CreatureSkill.
ORM Fixes.
Minor database fixes.
Fixes to character saving/updating.
Work getting landblock objects loading properly.
Character can now log-in and walk around, in-game objects don't appear (Although they do appear to pull properly from the DB now).

NOTE:  I've been using a separate ace_shard sql file Og sent me.  I didn't commit it, because a simple diff says its significantly different than the one in the current repo.  I also had to tweak it to get login working.  I'll submit it separately so we can figure out what to do with it.


Source/ACE.Database/ShardDatabase.cs
```diff
@@ -359,6 +365,9 @@ public void DeleteOrRestore(ulong unixTime, uint id)
             return objects;
         }
 
+        // FIXME(ddevec): This is racy.  If we have 2 sessions create a character at the same time this may return the same key 2x, crashing the server
+        // One solution would be to read at init into a local variable, then protect accesses to that variable with a lock
+        // I'm not that familiar with DB's as a whole, so I'll leave this for a DB person to fix.
```
Mogwai-TheFurry: Known issue - we're aware.

Source/ACE.Entity/AceObject.cs
```diff
@@ -296,7 +283,7 @@ public string Name
         /// </summary>
         public uint ItemType
         {
-            get { return GetIntProperty(PropertyInt.ItemType).Value; }
+            get { return IntProperties.Find(x => x.PropertyId == (uint)PropertyInt.ItemType).PropertyValue; }
```
Mogwai-TheFurry: What's the difference, here?

Source/ACE/Entity/Player.cs
```diff
@@ -456,6 +478,42 @@ public void Load(AceCharacter character)
             Session.Network.EnqueueSend(setTurbineChatChannels, general, trade, lfg, roleplay);
         }
 
+        public AceObject GetSavableCharacter()
+        {
```
Mogwai-TheFurry: So this is a code smell, too.  AceObject can never behave as an AceCharacter.  What's the use case here?

Source/ACE/Network/Handlers/CharacterHandler.cs
```diff
@@ -258,6 +258,15 @@ public static void CharacterRestore(ClientMessage message, Session session)
 
             character.IsAdmin = Convert.ToBoolean(reader.ReadUInt32());
             character.IsEnvoy = Convert.ToBoolean(reader.ReadUInt32());
+
+            character.WeenieClassId = 1;
+
+            // Required default properties for character login
+            // FIXME(ddevec): Should we have constants for (some of) these things?
+            character.ItemType = (uint)ObjectType.Creature;
```
Mogwai-TheFurry: These should go in the constructor.

Source/ACE.Entity/AceObject.cs
ddevec: That was an unintentional change.  It should use the GetIntProperty interface, will change.

Source/ACE/Entity/Player.cs
ddevec: This creates a logical copy of the player object as an AceObject for saving.  Since saving will be done asynchronously we must create a logical copy of the player when we save, otherwise we're subject to race conditions on the player's updates (unless we do extensive locking, but that would be a poor choice).

Source/ACE.Entity/AceObject.cs
Mogwai-TheFurry: There was a handful of them, I only commented on one.  We'll get the others later.

Source/ACE/Entity/Player.cs
Mogwai-TheFurry: Ok, but can of worms.  How do we save character-only data?  Are those all atomically separate?

PR #376, opened 2017-06-12
Object overhaul

ddevec:
Now successfully enters world, objects appear, and if you have the correct world data, you can even interact with the objects.

Getting close to prior functionality.

Refactored Position and CreatureAbility to front-end and back-end entities.


LtRipley36706:
Haven't had much chance to play with this but in my attempt at importing data exported from PCAPs, the WorldBase SQL file is missing some tables.

PR #377, opened 2017-06-12
Fear the Walking Dead - More work on our object overhaul - bug fixes and continued work

ogmage78:
* fixed issue with null exception
* Added OptimShi code to load the correct player appearance 
* Added sending the DataID and InstanceID fields on the player description event.
* Fixed assorted bugs.
* we can now go into and out of combat mode. 
* ci tested and working for all item types.
* telepoi tested and working.
* Trello board updated and cards moved into review - Pending Merge queue.
https://trello.com/b/DXQUfVHj/acemulator-server-project

### DIFF
--- a/Database/Base/WorldBase.sql
+++ b/Database/Base/WorldBase.sql
@@ -258,7 +258,7 @@ CREATE TABLE `ace_position` (
   KEY `idxPostionType` (`positionType`),
   CONSTRAINT `fk_ap_ao` FOREIGN KEY (`aceObjectId`) REFERENCES `ace_object` (`aceObjectId`),
   CONSTRAINT `fk_position_ao` FOREIGN KEY (`aceObjectId`) REFERENCES `ace_object` (`aceObjectId`) ON DELETE CASCADE
-) ENGINE=InnoDB AUTO_INCREMENT=178436 DEFAULT CHARSET=utf8;
+) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
 
 /*Table structure for table `ace_weenie_class` */
 
--- a/Database/Base/WorldBase.sql
+++ b/Database/Base/WorldBase.sql
@@ -1,5 +1,5 @@
 /*
-SQLyog Ultimate
+SQLyog Ultimate v12.4.2 (64 bit)
 MySQL - 10.2.6-MariaDB : Database - ace_world
 *********************************************************************
 */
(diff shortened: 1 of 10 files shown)


Database/Base/WorldBase.sql
```diff
@@ -21,201 +21,244 @@ USE `ace_world`;
...
+  `aceObjectDescriptionFlags` int(10) unsigned NOT NULL,
+  `weenieClassId` int(10) unsigned NOT NULL,
+  `weenieHeaderFlags` int(10) unsigned NOT NULL,
+  `physicsDescriptionFlag` int(10) unsigned NOT NULL,
+  `currentMotionState` text DEFAULT NULL,
   PRIMARY KEY (`aceObjectId`),
-  KEY `idx_weenie` (`weenieClassId`)
-) ENGINE=INNODB DEFAULT CHARSET=utf8;
+  KEY `idx_weenie` (`weenieClassId`),
+  CONSTRAINT `fk_weenie_ao` FOREIGN KEY (`weenieClassId`) REFERENCES `ace_weenie_class` (`weenieClassId`)
```
Mogwai-TheFurry: can't have a foreign key before you have the table.  script is not safe to run once and be done.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -282,51 +236,18 @@ public AceObject GetBaseAceObjectDataByWeenie(uint weenieClassId)
             // We can do this because aceObjectId = WeenieClassId for all baseAceObjects.
             // TODO: Ask Mogwai how would you query on a secondary key?
             var criteria = new Dictionary<string, object> { { "aceObjectId", weenieClassId } };
-            if (
-                !ExecuteConstructedGetStatement(
-                    WorldPreparedStatement.GetWeenieClass,
-                    typeof(AceObject),
-                    criteria,
-                    bao)) return null;
+            if (!ExecuteConstructedGetStatement(WorldPreparedStatement.GetWeenieClass, typeof(AceObject), criteria, bao)) return null;
```
Mogwai-TheFurry: return always on a new line

Source/ACE/Entity/Player.cs
```diff
@@ -473,10 +259,32 @@ public void Load(AceCharacter character)
...
+            PhysicsData.Stable = Character.SoundTableId;
+            PhysicsData.Petable = Character.PhysicsTableId;
+            PhysicsData.CSetup = Character.ModelTableId;
+
+            ContainerCapacity = 7;
+
+            if (Character.DefaultScale != null)
+                PhysicsData.ObjScale = Character.DefaultScale;
+
+            // Hair/head
```
Mogwai-TheFurry: Model data that cannot change easily (such as is defined at character creation) will probably need to ultimately go in a separate data structure so that it can be restored if a player takes off all equipment.

Database/Base/WorldBase.sql
ogmage78: fixed

Source/ACE.Database/WorldDatabase.cs
ogmage78: fixed

Source/ACE/Entity/Player.cs
ogmage78: yea - we were discussing that the other day
Right now we reload that base each time.     we are going to have to decide if we want to store your base undressed appearance and build up from there - which I think is a good way to do it
that way logging in is behind the scenes just equipping your items
and your appearance is the addition of your armor ordered by priority

I will add a card on Trello.

PR #385, opened 2017-06-16
asynchronous-object-saving outside game loop.

Forbiddenz:
Moved DB saving into its own game and implemented a blocking collection model.
Card - asynchronous-object-saving


Source/ACE/Managers/DbManager.cs
```diff
@@ -0,0 +1,50 @@
...
+            // starts game loop.
+            running = true;
+            new Thread(Tick).Start();
+        }
+
+        private static void Tick()
+        {
+            while (running)
+            {
+                Thread.Sleep(1);
```
ddevec: The Thread.Sleep() is unneeded, as the BlockingCollection manages waiting.

Forbiddenz:
Ok, let me try to add that in, today a bit crazy.  I also would like to try figure out a a better async reporting method maybe in here..  I am just a little wary of the await function inside this thread.

Mogwai-TheFurry:
Also:
* Character saving on another thread is good.  Moving ALL saving to another thread is better.  Making characters but not items async could lead to bad race conditions and object duplication.
* There's no graceful shutdown mechanism.  The Tick loop needs to check a flag that can be set on the class for a graceful shutdown after all work is done.

Source/ACE.Database/ICommonDatabase.cs
```diff
@@ -15,5 +15,6 @@ public interface ICommonDatabase
         AceObject GetObject(uint aceObjectId);
 
         Task<bool> SaveObject(AceObject aceObject);
+        Task<bool> SaveObjectAsync(AceObject aceObject);
```
Mogwai-TheFurry: Normally, you would have this:
```
bool SaveObject(AceObject aceObject);
Task<bool> SaveObjectAsync(AceObject aceObject);
```

Source/ACE.Database/IShardDatabase.cs
```diff
@@ -34,5 +34,8 @@ public interface IShardDatabase : ICommonDatabase
         uint GetNextCharacterId();
 
         AceCharacter GetCharacter(uint id);
+
+        // call back error interfaces
+        void UnableToSaveObject(AceObject ao);
```
Mogwai-TheFurry: This isn't part of the database interface - you need a callback to the calling code, not just some other function in your current class.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -420,6 +420,35 @@ public uint RenameCharacter(string currentName, string newName)
...
+                UnableToSaveObject(aceObject);
+                result = false;
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// Hooking into this interface will catch a saveobject error.
+        /// </summary>
+        public void UnableToSaveObject(AceObject ao)
```
Mogwai-TheFurry: Take out the UnableToSaveObject stuff - your logging will do for now.

PR #386, opened 2017-06-16
Bug Fix - Delete Character Trello Card

ogmage78:
* Trello card task - delete character crashes server.   Fixed this issue.
* Added event code for database to do the actual house keeping to flip the flag once the hour restore period has expired.
* Minor code cleanup.
* Removed update directory for old character database.
* Fixed pickup and drop item.   Location was protected and not able to be set for loot (WorldObjects)   I temp set this to allow set.   Once we refactor physics data out this can go away.


Database/Updates/Shard/01-06-16-2017-SetCharacterDeletedFlag.sql
```diff
@@ -0,0 +1,18 @@
+DELIMITER $$
+
+CREATE DEFINER=`root`@`localhost` EVENT `SetCharacterDeletedFlag` ON SCHEDULE EVERY 15 MINUTE STARTS '2017-06-16 16:15:35' ON COMPLETION NOT PRESERVE ENABLE DO BEGIN
```
LtRipley36706: Can change to `EVERY 1 MINUTE` and the net effect is the character is deleted from the character list on time instead of ~15 minutes after deleteTime

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -173,9 +174,15 @@ public Task RemoveAllFriends(uint characterId)
...
 
-        public void DeleteOrRestore(ulong unixTime, uint id)
+        public async Task<bool> DeleteOrRestore(ulong unixTime, uint aceObjectId)
         {
-            throw new NotImplementedException();
+            AceCharacter aceCharacter = new AceCharacter(aceObjectId);
+            LoadIntoObject(aceCharacter);
+            aceCharacter.DeleteTime = unixTime;
+
+            // aceCharacter.Deleted = true;  This is just a reminder - the DB will set this 1 hour after deletion.
```
LtRipley36706: As a protection we should put `aceCharacter.Deleted = false` which also makes this work for Restore.

PR #388, opened 2017-06-17
Fix "Player" name.

ddevec:
Cleaned up player initialization, and bugs related to player creation.  Player names now display properly.


ddevec:
After reviewing Og's comments, I should probably look at how the flags/headers work.  Right now I take all data possible from the database, apparently it holds extra data I should re-calculate.  Hopefully will address tomorrow (6-18-17)

Source/ACE/Network/GameAction/Actions/GameActionMoveToState.cs
```diff
@@ -90,7 +90,9 @@ public static void Handle(ClientMessage message, Session session)
             forcePositionTimestamp = message.Payload.ReadUInt16();
             message.Payload.ReadByte();
 
-            session.Player.UpdateLocation(position);
+            // FIXME(ddevec): The game sends invalid positions sometimes... when in portal space.  Ignore?
+            if (position.Cell != 0)
```
Mogwai-TheFurry: I'm thinking we might ought to have a Position constant that indicates portal space.  This way you don't leave the character behind somewhere too long where they can still be targeted by monsters or something after they've portaled away.

Source/ACE.Entity/AceCharacter.cs
```diff
@@ -22,6 +22,8 @@ public AceCharacter(uint id)
             // Required default properties for character login
             // FIXME(ddevec): Should we have constants for (some of) these things?
             ItemType = (uint)ObjectType.Creature;
+            AceObjectDescriptionFlags = (uint)(ObjectDescriptionFlag.Stuck | ObjectDescriptionFlag.Player | ObjectDescriptionFlag.Attackable);
+            WeenieHeaderFlags = (uint)(WeenieHeaderFlag.ItemCapacity | WeenieHeaderFlag.ContainerCapacity | WeenieHeaderFlag.Usable | WeenieHeaderFlag.RadarBlipColor | WeenieHeaderFlag.RadarBehavior);
```
Mogwai-TheFurry: This is absolutely ideal, but are we ready for this yet?  Just checking.

PR #397, opened 2017-06-19
Fix CIRand admin function.

ogmage78:
* Trello card task - Extend ORM code to support multiple getLists or implement another method to query by alternate keys - re-implement cirand
* Many thanks to ddevec - he did the work on the ORM code revision.   I just tested and implemented for my use.
* Fixed undiscovered alignment bug at hook type we were sending as dword and it is a word.   That was shifting everything below it.
* Added priority to aceObject and set the aceObject priority field as a backing object for worldObject priority.
* Created function to calculate container burden.
* added view to world database that was needed for cirand and will be used in loot generation.
* updated position table to have a faster function for location per discord discussion.
* Updated trello board to reflect current project status.


Source/ACE/Entity/Container.cs
```diff
@@ -72,14 +75,24 @@ public virtual void RemoveFromInventory(ObjectGuid inventoryItemGuid)
...
                 if (inventory.ContainsKey(inventoryItemGuid))
                     inventory.Remove(inventoryItemGuid);
             }
         }
 
+        public ushort UpdateBurden()
+        {
+            ushort calculatedBurden = 0;
+            lock (inventoryMutex)
+            foreach (KeyValuePair<ObjectGuid, WorldObject> entry in inventory)
```
Mogwai-TheFurry: can we always do explicit scoping on a lock, please?
```
lock (inventoryMutex)
{
    // locked stuff here
}
```

Database/Updates/World/01-06-19-2017-alterPostionTable.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE `ace_world`.`ace_position`   
+  DROP COLUMN `landblock` ,
+  ADD COLUMN `landblock` INT(5) UNSIGNED AS ( ((landblockraw >> 16) & 0xFFFF)) VIRTUAL,
```
Mogwai-TheFurry: so much easier to read...

Database/Updates/World/01-06-19-2017-alterPostionTable.sql
```diff
@@ -0,0 +1,8 @@
...
+  DROP COLUMN `cell` ,
+  ADD COLUMN `cell` INT(5) UNSIGNED AS ( (CONV(RIGHT(LPAD(HEX(`landblockRaw`),8,'0'),4),16,10)) ) VIRTUAL,
```
Mogwai-TheFurry: but this is so painful!  do we even have a use case for needing cell as a separate column?  if not, just drop it.

Source/ACE/Entity/Container.cs
ogmage78: Done

Database/Updates/World/01-06-19-2017-alterPostionTable.sql
ogmage78: Dropped ')

Database/Updates/World/01-06-19-2017-alterPostionTable.sql
LtRipley36706: Can also change back to smallint with this code
``landblock` smallint(5) GENERATED ALWAYS AS ((`landblockRaw` >> 16)) VIRTUAL,`
as the previous expression was forcing the change to an INT for whatever reason

PR #398, opened 2017-06-20
Landblock/Concurrency/Core Functionality Restructure (attempt #2)

ddevec:
Sorry about recreating the pull.  I accidentally borked my old branch and had to close the PR.

Note: This branch also includes Og's "MoveTo" operations.

The overall goal of this branch is to restructure our main-game logic in a manner that safely allows parallelism, without complex locking schemes.  The overall idea of this restructure can be summarized as follows:
- Physics, Movement, and Game Logic updates are done in their own phases, each phase is executed to completion before the next begins.
- Within the GameLogic phase, WorldObjects are updated in parallel, but updates lock-free operations with these mechanics:
   -- WorldObjects can only read/modify their own state on their own turn to "Act"
   -- Communication between WorldObjects is allowed by having the Landblock forward "Actions" between the WorldObjects.

To facilitate this design I've added several classes (IActions, IActors, ActionChains) that encapsulate this functionality.  A more detailed list of the behaviors is below.

Overall summary:
All work happens from UpdateWorld() thread in WorldManager
Outline of framework for physics position updates.
Added work managing infrastructure: Actions -- Remove busy-waiting/polling

Parallelism handled as follows:
- All motion (will be) calculated by the physics engine (based on requests done in the object update phase)
- After calculated, all motion is updated sequentially in the UpdateWorld function
- All landblocks update any objects within them through Actions on an ActionQueue
  -- Landblocks are acted on in parallel
  -- Objects within landblocks act sequentially
- Any action on an object may only read/modify the object its acting on 
  -- Exception - position information - guaranteed not to change by above points
  -- Any cross-object interactions must be done by queueing an action on the other object with data passed (This is currently handled by delegate/lambda functions within the action infrastructure, but could be co-routines, or dedicated per-action classes)

Other improvements/notes:
- The "DelayManager" Handles waiting, removing the need for busy waiting.
- All asyncrhonous inputs (e.g. network communication) must delegate their work to the action handlers of objects (Designated by functions prefixed with HandleAction).
- There is an infrastructure for position-based broadcasting (although issues in our landblock adjacencies need to be fixed before this will actually work cross-landblocks)
- Landblocks are also "Actors" as they must manage object pick-ups/etc.
- The movement to the "Actor/Action" interface removes any need for polling -- If we have a world with 1 billion entries, but only 2 active, it will run as fast as a world with 2 entries

This is a huge change, I tried to segment it, but the races caused by restructuring the parallelism broke the server until I restructured how incoming network events happened -- The end result was this branch.

As far as I know the current branch does everything master does, plus has some minor bugfixes, and supports broadcasts for more things (like coming out of portal space, updating position, dropping/picking up items).

I'm looking forward to hear what you think of this structure.

Notable things removed:
Landblock.GeWorldObjectById
Landblock.Broadcast (handled by players now)
Landblock.UseTime

Things that still need to be done:
More robust object tracking
Asynchronous Landblock loading/prefetching.


ddevec:
Object tracking is a known "todo".  I made a smaller range to debug items populating/disappearing.  Its trivial to increase that range, although it will just be a temporary patch.

The "Pressing J" issue is due to a malformed animation packet causing position corruption (in fact, sometimes pressing J crashes my client).  This is a bug in our current codebase that is masked by our very large object tracking range, (exposed by my changes, not fundamental to them).

I've not seen the portal-space masking bug.  If you can give me steps to recreate it, I'll look into it, that's likely one of mine.

My patch shouldn't have changed any of our character flags.  This is likely a bug that was introduced by other code, but is easier to see as we can actually see characters populate/interact now.  I'll try to put in a quick fix for it though.

Summary:
- I'll change the tracking range, no problem (although this is just masking an inefficient system).
- I'll look into the ethereal flag.
- If you can tell me how you got the purple bubbles to persist, I really want to look at that.

- I can work on fixing the bugs exposed not introduced by my branch if you really want (e.g. malformed animation packets).  I prefer more targeted commits.  The goal of this branch (currently) isn't to fix all bugs in the game, but to introduce a new core/landblock structure.

ddevec:
After further reviewing the PCAPs and experimenting with the "Ethereal" physics state, this does not appear to be the solution to the physics detection (players still run into each other when its on, and I've yet to see it set in a retail createobject for a player, but maybe my sampling was bad).  Perhaps this issue is best left to another pull request?

There was a bug in my object tracking (I wasn't squaring distance in one of my checks).  You'll now see better behavior in that regard.

I'll look into the missed physics/bubble state packet when I can reproduce it.

Source/ACE/Entity/Creature.cs
```diff
@@ -67,13 +69,21 @@ public CreatureVital Mana
         }
 
         /// <summary>
-        /// This will be false when creature is dead and waits for respawn
+        /// If we need to move, where do we need to go?   This may be replaced by a property that Mogwai was discussing
```
Mogwai-TheFurry: not that i recall :|

Source/ACE/Entity/Landblock.cs
```diff
@@ -25,23 +30,33 @@ namespace ACE.Entity
...
     {
         private static readonly ILog log = LogManager.GetLogger(System.Reflection.MethodBase.GetCurrentMethod().DeclaringType);
 
-        private const float adjacencyLoadRange = 96f;
-        private const float outDoorChatRange = 75f;
-        private const float indoorChatRange = 25f;
-        private const float maxXY = 192f;
-        private const float maxobjectRange = 20000;
-        private const float maxobjectGhostRange = 40000;
+        public const float adjacencyLoadRange = 96f;
```
Mogwai-TheFurry: No public fields, please. Private fields, public properties.  Note, private => camelCase.  public => ProperCase.

Source/ACE.Entity/Position.cs
```diff
@@ -237,6 +237,20 @@ private uint GetCellFromBase(uint baseX, uint baseY)
             return (block << 16) | (cell + 1);
         }
 
+        public static LandblockId XYToLandblock(float x, float y)
+        {
+            uint baseX = (uint)(x + 0x400);
```
Mogwai-TheFurry: Combined with how it is used in Landblock.cs (for adjacencies), this math looks wrong.  Also, landblocks in dungeons have no functional adjacencies for loading/roaming purposes so that needs to be taken into account.  I'm not sure what your intended functionality is here or I'd offer more help on how to do this, but I don't see the point of this function.  Add a constant offset to x and y float coords (resulting in the wrong landblock x/y), and calculate the cell based on x/y math?  The cell I get, but the block x/y is wrong.  You're passing in the x/y floats from WITHIN the landblock (0 to 192) when this is called from Landblock.cs.  You need that to take a stab at what cell you're in, but cells in dungeons aren't the nice simple grid as they are in the overworld.

And, I could be completely wrong.  In which case you need to create function headers to describe what these things are doing :)

Source/ACE/Entity/Landblock.cs
```diff
@@ -187,688 +207,512 @@ private void AddPlayerTracking(List<WorldObject> wolist, Player player)
...
+            while (!broadcastQueue.IsEmpty)
+            {
+                Tuple<Position, float, GameMessage> tuple;
+                bool success = broadcastQueue.TryDequeue(out tuple);
+                if (!success)
+                {
+                    log.Error("Unexpected TryDequeue Failure!");
+                    break;
                 }
+                Position pos = tuple.Item1;
```
Mogwai-TheFurry: Tuples make me want to go to VS 2017 and .Net vNext.  Ugh...

Source/ACE/Entity/Player.cs
```diff
@@ -48,7 +53,7 @@ public sealed class Player : Creature
         /// <summary>
         /// ObjectId of the currently selected Target (only players and creatures)
         /// </summary>
-        public uint SelectedTarget { get; set; }
+        private ObjectGuid SelectedTarget { get; set; }
```
Mogwai-TheFurry: private => camelCase.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -699,8 +730,61 @@ public void WriteUpdatePositionPayload(BinaryWriter writer)
             writer.Write(Sequences.GetCurrentSequence(SequenceType.ObjectForcePosition));
         }
 
-        public virtual void Tick(double tickTime)
+        protected void ForceUpdatePosition(Position newPosition)
+        {
+            ForcedLocation = newPosition;
+        }
+
+        /// FIXME(ddevec): We're probably ultimately only going to update velocity, not true position...
```
Mogwai-TheFurry: we'll have to do both.  velocity for client prediction, position for client correction.

Source/ACE/Managers/LandblockManager.cs
```diff
@@ -17,9 +19,16 @@ public static class LandblockManager
 
         private static readonly object landblockMutex = new object();
 
-        // debating during this into a dictionary keyed on LandblockId.  would be functionally equivalent
-        // and get rid of the 2D array shenanigans
-        private static readonly Landblock[,] landblocks = new Landblock[256, 256];
+        // FIXME(ddevec): Does making this volatile really make double-check locking safe?
```
Mogwai-TheFurry: safe enough, IMO.

ddevec:
Mogwai, I believe this commit addresses your concerns (I left in the double check locking fixme though, because I hate double-check locking.  You can remove the comment on commit if you really want to).

LtRipley36706:
Pulled latest changes and the range was indeed much better. We'll still need to address deleting objects as that only occurs when something is destroyed in 3d space or goes in to someone's pack from the ground for example. I'll try to look at it more again this evening. as well as seeing if I can replicate the purple state.

LtRipley36706:
One other thing to use as reference
`/teleloc 0x01F50229 50 -180 0 1 0 0 0`
This is a dungeon that is a pretty good test for indoor range issues. As of current pull, if you run to the bottom you'll see how the top level had everything spawned correctly but as you run down deeper, you'll find that doors, levers, chests are not in position. If you relog at various levels, you'll see them correctly for where you are, but as you run around and away from them, they are removed from the client and never reappear unless you relog or teleport out and back.

ddevec:
So, this issue with objects vanishing in large, indoor landblocks is an issue with our tracking system, and is present in the master branch.  I agree we need to fix this.  I created a card on Trello to address it, and I'm looking forward to fixing it, but I believe its outside the scope of this (already large) pull request.

This branch does send more deletes than our other branch did, because it has a sane tracking range (our other branch sent deletes when objects left tracking range, the range was >>> 2 landblocks, so it never sent deletes).  I'll try to address that tonight (although it is object tracking behavior, that doesn't effect the functionality of this branch, nor is it a digression from our prior branches behavior.)

If you can replicate the purple state after the changes I made yesterday, I will work very hard to fix that.

ddevec:
Apparently trying to rebase a branch that's been publicly pushed doesn't end well.  Sorry about the long commit chain, I can make a new PR with a clean rebase, if needed.

I added a fix for the many object deletes.

Source/ACE/Entity/Landblock.cs
```diff
@@ -396,26 +396,44 @@ private void UpdateStatus(int pcount)
...
+        /// <param name="distance"></param>
+        /// <returns></returns>
         private List<Landblock> GetLandblocksInRange(Position pos, float distance)
         {
             List<Landblock> inRange = new List<Landblock>();
 
             inRange.Add(this);
 
+            if (pos.Indoors)
+            {
```
Mogwai-TheFurry: FWIW, this is worth testing when you walk in a building.  I think that flag ends up set in a building in the Overworld.

PR #401, opened 2017-06-20
Equip and Unequip items

ogmage78:
* Added the ability to equip and unequip items and have it also update your visual appearance.
* We have some data work to do - to add clothing table did values, but we have added 30 or so for testing out of 150+ needed.   Thanks OptimShi for the script and the updates - also all the help on clothing!
* I am adding the script to load those testing values - they are the permanent solution just incomplete.   However, it properly goes in ace world.   The prerequisite to running the script is to have data loaded.   If you do it out of order, no harm, it just will not work.


Database/Updates/World/04-06-20-2017-alterPostionTable-stored.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE `ace_world`.`ace_position`
+  DROP COLUMN `landblock` ,
+  ADD COLUMN `landblock` INT(5) UNSIGNED AS ( ((landblockraw >> 16) & 0xFFFF)) STORED,
```
LtRipley36706: For performance reasons this line should remain VIRTUAL instead of STORED please.

It could also be changed back to a smallint since the expression used here doesn't complain about int. up to you on that one.

Database/Updates/World/05-LoadInititalClothingTableDIDs.sql
```diff
@@ -0,0 +1,72 @@
...
+# Beret
+INSERT INTO `ace_object_properties_did` (`aceObjectId`, `didPropertyId`, `propertyValue`) VALUES ('31236', '7', '268436908'); 
+# Poet's Shirt
+INSERT INTO `ace_object_properties_did` (`aceObjectId`, `didPropertyId`, `propertyValue`) VALUES ('31239', '7', '268436909'); 
+# Viamontian Leggings
+INSERT INTO `ace_object_properties_did` (`aceObjectId`, `didPropertyId`, `propertyValue`) VALUES ('31238', '7', '268436910'); 
+# Viamontian Laced Boots
+INSERT INTO `ace_object_properties_did` (`aceObjectId`, `didPropertyId`, `propertyValue`) VALUES ('31249', '7', '268436911'); 
+# Loafers
+INSERT INTO `ace_object_properties_did` (`aceObjectId`, `didPropertyId`, `propertyValue`) VALUES ('31237', '7', '268436912');
```
LtRipley36706: Please submit this data to ACE-World

Source/ACE/Entity/Player.cs
```diff
@@ -1388,7 +1391,56 @@ public override void Tick(double tickTime)
...
+
+            foreach (var w in wieldeditems)
+            {
+                var wo = w.Value;
+                ClothingTable item;
+                if (wo.ClothingBase != null)
+                    item = ClothingTable.ReadFromDat((uint)wo.ClothingBase);
+                else
+                    item = ClothingTable.ReadFromDat(0x10000002);
+                // If I don't find a match, for now you are pants!
```
LtRipley36706: Maybe change this to some kind of error catcher with a do nothing on the model + chat error like portals do when they aren't wired up, so we don't end up with some mixed results?

Database/Updates/World/04-06-20-2017-alterPostionTable-stored.sql
ogmage78: I guess you will just not respect my wishing that Mogwai review my work.   Your treatment of me is unacceptable to me and I do not wish you to review my work.   I have made my wishes very clear and I have asked you nicely on at least 10 occasions.   "Please allow Mogwai to review my work".   As to your point, for loading, it would be very easy for you to drop the column to load.   On query, virtual query on that key to load 1400 records, 4 seconds.   Stored sub 1 second.   I am not changing this as it impacts the function that supports it.   I don't want to spend my free time with someone constantly who is critical of everything I do.   If you have to review my work - then my time on this project is at an end.   I will just work on a copy in my fork.   I will not submit anything to this project if you have to review it.   I have been working for almost 40 years and I have never been treated with such disrespect, or worked with anyone that would not discuss topics.   Your default position is that you make the final call end of story.   I am not going to do that.

Database/Updates/World/05-LoadInititalClothingTableDIDs.sql
ogmage78: As I put in my notes, this does belong there and could be moved a little later.   You can't test the work I did without this.   Again, you don't consider what the free developer is trying to do or say.   This is open source.   I think the word is collaboration.

Source/ACE/Entity/Player.cs
ogmage78: That is a change I plan to make later.   Thanks

ogmage78:
That is really all I have to say.   Thanks

Database/Updates/World/04-06-20-2017-alterPostionTable-stored.sql
Mogwai-TheFurry: Was the performance hit during a big import or when querying after the import?

ogmage78: I ran the test again - it is faster stored on query (not insert) but it is not that impactful.   I am ok with changing this back if it makes sense.

ogmage78: you can test with the following:

ALTER TABLE `ace_world`.`ace_position`   
  DROP INDEX `idx_landblock`,
  DROP COLUMN `landblock` ,
  ADD COLUMN `landblock` INT(5) UNSIGNED AS ( ((landblockraw >> 16) & 0xFFFF)) VIRTUAL,
  ADD  KEY `idx_landblock` (`landblock`);

select * from ace_position order by landblock

then 

ALTER TABLE `ace_world`.`ace_position`   
  DROP INDEX `idx_landblock`,
  DROP COLUMN `landblock` ,
  ADD COLUMN `landblock` INT(5) UNSIGNED AS ( ((landblockraw >> 16) & 0xFFFF)) STORED,
  ADD  KEY `idx_landblock` (`landblock`);

select * from ace_position order by landblock

it would be easy to just drop this column during import of PCAPS as well then add at the end of the import.   it takes 22 seconds to switch the column storage and build the index.

LtRipley36706: @Mogwai-TheFurry The impact was noticed on the first in-game load in of each place I visited, for example I loaded the server, logged in Arwic, exit portal, see empty town for a few seconds then objects populated, teleport to Holtburg, see same behavior with a variance of the length of time the load in takes, eventually return to Arwic and its town is loaded instantly back to the client which would make the think there is some caching in memory going on?

Mogwai-TheFurry:
Tried to merge by hand - Container was fine.  Landblock is no-go.

ogmage78:
Going to reimplement after major overhaul

PR #404, opened 2017-06-22
Inventory saving

Forbiddenz:
Ability to save inventory, no load..
But this shows us a bigger issue in our data model.

There also a rig job in here to allow up to 20 transactions conflicts before failing but there no transaction order, so its not the greatest method and could result in saved operations happening out of order due to deadlock failure on the database.

Overall there is a bigger issue in our data model that needs addressed.
This PR is just to allow anyone to see it live if they remove the generic error handling I put in to make this not crash ACE.


Forbiddenz:
Don't Merge,  just to show the db deadlock problem

PR #405, opened 2017-06-22
Serialized saves, added caching manager for saves.

ddevec:
Goal: resolve MySql deadlock exception caused by saving too much stuff at once.

Added serialization of ShardDatabase saves.
Added save-lots debug function.
Also fix bug in DbManger on shutdown.

TODO:  combine DbManager and ShardDatabase.


ddevec:
Pull request superceded by new pull.

PR #407, opened 2017-06-23
General ShardDatabase cleanup, serialization, and caching layer

ddevec:
DbManager is now Integrated into ShardDatabase along with:

- Serial database writing.
- Caching layer to ensure latest data is always loaded
- Cleanup to ShardDatabase (replaced largely redundant functions with templated variants).
- Fix to shutdown of saving thread.
- Added multi-save debugging function save-logs (because who doesn't want to save their character 10k times right now?)


Mogwai-TheFurry:
In general I'd rather see us implement new implementations of interfaces when we're doing wrapping functions like caching or off-threading.  We could have ShardCache : IShardDatabase that has a constructor that takes an IShardDatabase in it's constructor.

This eliminates the tight coupling of a class doing 2 things: storage and caching.  Having those 2 components is a historically bad combination.

PR #408, opened 2017-06-23
Quick fix for Portals to become functional

LtRipley36706:
Code provided by @ogmage78 successfully re-enables portals to work again on double click provided they have PositionType.Destination data in the ACE-World database. Latest release of database includes all previously submitted portal destination data.

Also corrected ObjectType issue in CollidableObject. It was not being set from data stored in database leading to aberrant behavior in OnUse in game.

### DIFF
--- a/Source/ACE.Database/WorldDatabase.cs
+++ b/Source/ACE.Database/WorldDatabase.cs
@@ -150,7 +150,11 @@ namespace ACE.Database
                 o.TextureOverrides = GetAceObjectTextureMaps(o.AceObjectId);
                 o.AnimationOverrides = GetAceObjectAnimations(o.AceObjectId);
                 o.PaletteOverrides = GetAceObjectPalettes(o.AceObjectId);
-                o.AceObjectPropertiesPositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                // o.AceObjectPropertiesPositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                // The following three lines is code drafted by Og II for the purpose of linking a weenie's position data to an instance's position data.
+                Dictionary<PositionType, Position> instancePositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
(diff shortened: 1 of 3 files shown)


Source/ACE.Database/WorldDatabase.cs
```diff
@@ -150,7 +150,11 @@ public List<AceObject> GetObjectsByLandblock(ushort landblock)
...
                 o.AnimationOverrides = GetAceObjectAnimations(o.AceObjectId);
                 o.PaletteOverrides = GetAceObjectPalettes(o.AceObjectId);
-                o.AceObjectPropertiesPositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                // o.AceObjectPropertiesPositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                // The following three lines is code drafted by Og II for the purpose of linking a weenie's position data to an instance's position data.
+                Dictionary<PositionType, Position> instancePositions = GetAceObjectPositions(o.AceObjectId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                Dictionary<PositionType, Position> weeniePositions = GetAceObjectPositions(o.WeenieClassId).ToDictionary(x => (PositionType)x.DbPositionType, x => new Position(x));
+                o.AceObjectPropertiesPositions = instancePositions.Union(instancePositions).Union(weeniePositions).ToDictionary(k => k.Key, v => v.Value); // not deduped ?
```
Mogwai-TheFurry: I thought about this some more overnight.

I would be more a fan of having a "WeeniePositions" property of ace object (also a Dictionary just like the current "Positions" property).  The Portal class would then look at WeeniePositions[PositionType.Destination] to figure out where to send someone.

I'm really not a fan of making our data fungible and function in different ways based on the class it gets loaded into.

LtRipley36706: @Mogwai-TheFurry what about instead of just weeniepositions we were to come up with a way to access the full weenie data object in separate properties but still accessible through each instance of an object?

LtRipley36706: there will be several use cases outside of just positions that this could be useful, vendors will be another example of that

Mogwai-TheFurry: i'd be ok with a global weenie cache implemented in the IWorldDatabase.

LtRipley36706:
Changes merged in #419

PR #410, opened 2017-06-23
Monster generator refactored

Lidefeath:
Remove the old creatures spawning and unused generator stuff and implement a new generator logic with the new data model. Use the included sql script to import testdata 'Drudgeskulkergen' near Holtburg LS.

The spawned drudge doesn't show up on radar yet, this has to be fixed still.


LtRipley36706:
Please also review #414 in connection to this PR for further thoughts about how the generator system should be deployed

PR #411, opened 2017-06-24
Pack Rat: Ready for review and merge.

ogmage78:
This has been extensively tested.   I am sure I have not followed all of the new action chain methodology correctly.   I would like to get this in as I think it moves us down the field.   I think as I get a better understanding of the action chain - I can refactor this with a minimum impact.   

* Added the ability to equip / unequip items and it will also update your visual appearance.
* We have some data work to do - finish add clothing table did values - a lot of these are done - thanks OptimShi and Ripley
* Tested - pack to pack, slot to slot, pack to ground, ground to pack, ground to wielded, wielded to ground
* move packs, move items. 
* Fixed save character options - this will need to be refactored - our character object needs some work.   
* UI changes still blows up - this fix just addressed the options1 and options2
* Fixed crash on lock or unlock UI
* TODO
1 clothing priority does not look to be right - somethings overlap incorrectly.
2 refactor for correct action chain pattern.
3 fix save for UI changes
4 refactor character save and load.   Talk to Mogwai on best way to do this.
5 UI positions still not saving - need to fix that next.

To test:  
@ci 136 (pack) or find a pack of your choice ( you can create several to test pack to pack and pack moves)
@ci 127 is a pair of pants

Extended testing
@cirand 2 (armor)
@cirand 4 (clothes)

Test items, if you get the message no clothing entry just drop and keep going until you have a nice selection to test.
Have fun.


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1001,8 +1001,26 @@ public static void HandleCI(Session session, params string[] parameters)
...
-            var loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
-            session.Player.HandleAddToInventory(loot);
+
+            // TODO: we have to be able to identify containers - you can create a wo and see if it has
+            // an item capacity but that seems a bit wasteful.   TBD on best method.   Limited number so
+
+            if (weenieId == 136)
+            {
+                var loot = LootGenerationFactory.CreateTestContainerObject(session.Player, weenieId);
+                session.Player.AddToInventory(loot);
```
ddevec: You shouldn't modify a player directly from the command/network based handler -- it races with the player accessing/modifying itself.  This is what actions are for.  Instead, enqueue an action on the player to deal with their own data.

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1001,8 +1001,26 @@ public static void HandleCI(Session session, params string[] parameters)
...
+                session.Player.TrackObject(loot);
+                session.Network.EnqueueSend(new GameMessagePutObjectInContainer(session, session.Player, loot, 0),
+                    new GameMessageUpdateInstanceId(loot.Guid, session.Player.Guid));
+            }
+            else
+            {
+                var loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
+                session.Player.AddToInventory(loot);
```
ddevec: See above -- don't modify players directly from network handlers.

Source/ACE/Entity/Landblock.cs
```diff
@@ -729,6 +733,13 @@ public void ScheduleItemTransfer(ActionChain chain, ObjectGuid wo, ObjectGuid co
             }
         }
 
+        public bool ItemExistsInWorld(ObjectGuid wo)
```
ddevec: I don't like this function.  Its dangerous to access the landblock, as items can be added to or removed from it arbitrarily.  If its only called in certain circumstances it may be safe, but in general its dangerous.

Source/ACE/Entity/Player.cs
```diff
@@ -1558,7 +1563,37 @@ protected override void SendUpdatePosition()
             Session.Network.EnqueueSend(msg);
         }
 
-        public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid containerGuid)
+        private void HandleUnequip(WorldObject item, uint location)
```
ddevec: Unequip is a fine name -- keep the prefix though HandleAction alerts the caller that this is known callable from outside of the players action queue (it will enqueue any action it needs to run safely).

Source/ACE/Entity/Player.cs
```diff
@@ -1644,9 +1672,78 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
+                    LandblockManager.RemoveObject(inventoryItem);
+                inventoryItem.PhysicsData.PhysicsDescriptionFlag &= ~PhysicsDescriptionFlag.Position;
+                inventoryItem.PositionFlag = UpdatePositionFlag.None;
+                inventoryItem.PhysicsData.Position = null;
+                inventoryItem.Location = null;
+                inventoryItem.WeenieFlags = inventoryItem.SetWeenieHeaderFlag();
+            }
+        }
+
+        public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid containerGuid, uint location = 0)
```
ddevec: This function looks at the player.  To do so safely it needs to run its operations as part of the player's actions.  (e.g. using ActionChains).  Its also prefixed with HandleAction indicating its safe to call from outside the player's action context (so it must use action chains).

Source/ACE/Entity/Player.cs
```diff
@@ -1823,7 +1918,56 @@ public void HandleActionSmiteSelected()
                 }
             }).EnqueueChain();
         }
-        
+
+        public void EquipUnequipItem(uint itemId, uint slot)
```
ddevec: This is public, and called from outside the player's action context (from a game message).  It should be HandleActionEquipUnequip, and use an ActionChain.

Source/ACE/Command/Handlers/AdminCommands.cs
ogmage78: I don't object, but I am not sure I understand.  I thought we were not multi-threaded - for the main game loop anymore.   How can I race modifying myself?   Not arguing - just trying to understand.   Thanks a ton by the way for the assist.   I am going to try to get my bug fixed first then I will try to make these other changes.

Source/ACE/Entity/Landblock.cs
ogmage78: I added that - when you get a message to put something in pack Put Item In Container from the client F7B1 0019 - you get 3 bits of data.   Item Guid, Container Guild and Location (this is overloaded) if the item is a container (or a foci)  going into a container, then location is the side slots for your character.   The determination of the item - container, foci or non container - is needed for our response in the F7B0 0022 where we send back 4 bits of information, the same 3 plus that identification of the item that is needed by the client.   It is called is_item_container and is always 0, 1 or 2 - 0 non container, 1 container and 2 foci.   I have created a property on world object to classify each world object as one of the 3.   

That was a long explanation, but basically when I get the message to put something in a pack, I have no way to know am I going internal pack to pack, from wielded to pack, pack to same pack different slot or ground to main pack, ground to slot in a side pack.   (another player or vendor handing us something is the same but without the ability to specify location, that action always uses slot 0)   So - why did I make that function?   To try and see which scenario I am dealing with.   I can kill that and just do the reverse.   I created a pack aware player inventory search so if I don't have it it must be from the world.   However, I have not seen a specific pcap of a player being handed something.   If it is the same message, that would be an issue for my current understanding on how this works.   I just made a post requesting a pcap.

ogmage78: I just found an example - it is a different message with the same response from the server.   The client sends us a F7B1 00CD   that will make it all work.

Source/ACE/Command/Handlers/AdminCommands.cs
ddevec: The main game loop has been staged.  There is still paralleism within the system.  For instance, network packed are handled asyncrhonously from the main game loop.  The commands all race with the worldobjects, so we handle that asynchrony by passing messages (in the form of actions).

If this winds up being a big issue, I may move command handlers into the main thread, but for now I would like to try it this way.

Source/ACE/Entity/Landblock.cs
ddevec: For reference (I don't understand your last post).

When we move items between containers we can identify if it is on the landblock or not by examine ourselves (or items we have exclusive control over).  Starting with a simple scenario:  the only items that can be put in our container are those on the world.

In this case, we're either moving an item between our containers (we can check this safely locally), or it comes from the world, no need to check its on the world to know how to handle it (you can check if its on yourself).

There are weird scenarios where we're grabbing an item from another container (e.g. your home chest, or a corpse), but we've gained exclusive access to that container ahead of time (by opening it), and therefore know to check it as well.

This make sense?

Source/ACE/Entity/Container.cs
```diff
@@ -26,48 +29,111 @@ public Container(ObjectType type, ObjectGuid guid, string name, ushort weenieCla
         public Container(AceObject baseObject)
             : base(baseObject) { }
 
+        public Container(ObjectGuid guid, AceObject baseAceObject)
```
Mogwai-TheFurry: I'm a bit puzzled why this constructor is the best solution.  The vast majority of these properties sets are setting properties on a default AceObject of the base class.  It seems that there would have been better solutions than doing explicit copies of everything (perhaps calling the other base constructor with a clone of the baseAceObject, then overriding the Guid?)

Source/ACE/Entity/Container.cs
```diff
@@ -26,48 +29,111 @@ public Container(ObjectType type, ObjectGuid guid, string name, ushort weenieCla
...
-            inventoryItem.PhysicsData.PhysicsDescriptionFlag = inventoryItem.PhysicsData.SetPhysicsDescriptionFlag(inventoryItem);
-            inventoryItem.ContainerId = null;
-            inventoryItem.Wielder = null;
-            inventoryItem.WeenieFlags = inventoryItem.SetWeenieHeaderFlag();
             if (inventory.ContainsKey(inventoryItemGuid))
             {
                 inventory.Remove(inventoryItemGuid);
+                Burden = UpdateBurden();
+            }
+            else
```
Mogwai-TheFurry: I've seen enough of this inventory-ownership-check to think that we should have a method for it.  Something along the lines of "HasItem(uint itemGuid)".  Especially now that it's 18 lines of code instead of 2.

Source/ACE/Entity/Player.cs
```diff
@@ -1558,7 +1573,42 @@ protected override void SendUpdatePosition()
...
+            UpdateAppearance(container, item.Guid.Full);
+            Session.Network.EnqueueSend(new GameMessagePutObjectInContainer(Session, container.Guid, item, location),
+                new GameMessageUpdateInstanceId(item.Guid, container.Guid, PropertyInstanceId.Container),
+                new GameMessageObjDescEvent(this),
+                new GameMessageSound(Guid, Sound.UnwieldObject, (float)1.0));
+            CurrentLandblock.EnqueueBroadcast(Location, Landblock.MaxObjectRange, new GameMessageObjDescEvent(this));
+        }
+
+        private void HandleMove(WorldObject item, Container container, uint location)
+        {
```
Mogwai-TheFurry: "HandleMove" in the Player class is a bit misleading going by the name alone.  Perhaps "MoveInventoryItem" or "HandleInventoryMove".

Source/ACE/Entity/Player.cs
```diff
@@ -1558,7 +1573,42 @@ protected override void SendUpdatePosition()
...
+                if (previousContainer != null)
+                    previousContainer.RemoveFromInventory(item.Guid);
+                container.AddToInventory(item);
+                item.ContainerId = container.Guid.Full;
+            }
+            Session.Network.EnqueueSend(new GameMessagePutObjectInContainer(Session, container.Guid, item, location),
+                                        new GameMessageUpdateInstanceId(item.Guid, container.Guid, PropertyInstanceId.Container));
+        }
+
+        private void HandlePickup(Container container, ObjectGuid itemGuid, uint location, PropertyInstanceId iidPropertyId)
```
Mogwai-TheFurry: NitPick: "HandlePickup _What_"

Source/ACE/Entity/Player.cs
```diff
@@ -1633,20 +1692,157 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
                 }
             });
 
             // Set chain to run
-            putItemInContainerChain.EnqueueChain();
+            pickUpItemChain.EnqueueChain();
+        }
+
+        private void SetInventoryForWorld(ObjectGuid containerGuid, WorldObject inventoryItem)
+        {
```
Mogwai-TheFurry: This code should definitely live in WorldObject.  Also note, containerGuid isn't used at all - remove it or use it.

Source/ACE/Entity/Player.cs
```diff
@@ -1633,20 +1692,157 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
+                                         | UpdatePositionFlag.ZeroQx;
+
+            inventoryItem.PhysicsData.PhysicsDescriptionFlag = inventoryItem.PhysicsData.SetPhysicsDescriptionFlag(inventoryItem);
+            inventoryItem.ContainerId = null;
+            inventoryItem.Wielder = null;
+            inventoryItem.WeenieFlags = inventoryItem.SetWeenieHeaderFlag();
+        }
+
+        private static void SetInventoryForOffWorld(ObjectGuid containerGuid, WorldObject inventoryItem)
+        {
```
Mogwai-TheFurry: This method also belongs in WorldObject.  And it's more like "PutInContainer".  Also, containerGuid is not used.  Remove it or use it.

Source/ACE/Entity/Player.cs
```diff
@@ -1633,20 +1692,157 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
+                SetupModel baseSetup = SetupModel.ReadFromDat((uint)PhysicsData.CSetup);
+                for (byte i = 0; i < baseSetup.SubObjectIds.Count; i++)
+                {
+                    if (!coverage.Contains(i) && i != 0x10) // Don't add body parts for those that are already covered. Also don't add the head, that was already covered by AddCharacterBaseModelData()
+                        ModelData.AddModel(i, baseSetup.SubObjectIds[i]);
+                }
+            }
+        }
+
+        public void HandleActionWieldItem(Container container, uint itemId, uint location)
```
ddevec: You've changed the name of this to HandleAction... but you haven't moved the work onto an action on the player's action queue.  This method is called from a network handler (network thread), and therefore runs concurrently with the player's actions.  This isn't allowed, as it modifies player's state.  To fix it you must delegate the work to an action.  Something like:

ActionChain chain = new ActionChain(this, () =>
{
work here
});
chain.EnqueueAction().

Look at all of the other HandleAction... functions for examples of how to do this.

Source/ACE/Entity/Player.cs
```diff
@@ -1633,20 +1692,157 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
+                return;
+            UpdateAppearance(container, itemId);
+            Session.Network.EnqueueSend(new GameMessageSound(Guid, Sound.WieldObject, (float)1.0),
+                                        new GameMessageObjDescEvent(this),
+                                        new GameMessageUpdateInstanceId(container.Guid, itemGuid, PropertyInstanceId.Wielder),
+                                        new GameEventWieldItem(Session, itemGuid.Full, location));
+            CurrentLandblock.EnqueueBroadcast(Location, Landblock.MaxObjectRange, new GameMessageObjDescEvent(this));
+        }
+
+        public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid containerGuid, uint location = 0)
```
ddevec: Same as above -- move the work into an action, or this will race and crash the server once we start getting a larger workload.

Source/ACE/Entity/Container.cs
ogmage78: I probably got confused.   It happens :)     That debug object - I don't understand why we have it.   I am not saying it is wrong.   It can't be used for containers like backpacks and it was causing me some real grief.    I will fix it.   Might be later this afternoon before I can work on it.

Source/ACE/Entity/Container.cs
ogmage78: I can do that.   I had to make that container aware.   We could do it another way.   I could take that logic out and add a dictionary on container that was a key value pair of objectguid and container guid.   Then, just ask do I have it from that - fast call.   If I need it, then it gives me the explicit container it is in and I can call the simple 2 line add or remove.   Let me know if you have a preference.

Source/ACE/Entity/Player.cs
ogmage78: Yes - I can fix that.

Source/ACE/Entity/Player.cs
ogmage78: Nit will be picked ;)   I will address

Source/ACE/Entity/Player.cs
ogmage78: Will do

Source/ACE/Entity/Player.cs
ogmage78: Will do

Source/ACE/Entity/Container.cs
Mogwai-TheFurry: Have the function also accept a "check sub-containers" flag as well?

Source/ACE/Entity/Container.cs
ogmage78: On looking at this more - the issue is I never finished moving physics data and model data into world object.   If I finish the move, then those properties are expressed using the aceobject as it's backing object.   Good news is - it is simple to finish the move.  The bad news is, I will touch or change 50% of the files to do it.   I think that needs to be a clean - that is the only thing I do PR.   Let me know what you want me to do.     physicsData.cSetup goes away and this shows up in world object for example.

public uint? CSetup
        {
            get { return AceObject.ModelTableId; }
            set { AceObject.ModelTableId = value; }
        }

Mogwai-TheFurry: Willing to accept as is provided you leave a // TODO tag.

PR #414, opened 2017-06-25
For Review: Expanded Generator Code/Scripting

LtRipley36706:
Please review for discussion and brainstorming how best we implement going forward


LtRipley36706:
It should be noted that this exposes some issues with items in the world. Clients don't seem to see them disappear when one picks an item up, or when put down after being picked up. Generators don't seem to have any respawn or item status tracking yet. Can't smite creatures so dunno if it it tracks them

Unrelated i'm also seeing crashes still when firing off an emote. not consistent but does still occur.

Lidefeath:
As already discussed in discord, I'd prefer to not spawn generators as items in the world. As the use case for that admin-only stuff is small. Besides it can get crowdy in places once linked or hierarchical - however you want to call them - generators come in: i.e. in a drudge camp with 3 different drudges and a campfire, you suddenly have 5 generators sitting around too (1 campgen + 3 drudgegens + 1 campfiregen)

LtRipley36706:
I think I understand what you're saying but to be sure let me check: you see the generators as a 1 to 1 for each wcid. A generator on the landscape spawns 3 drudges + 1 camp fire, each drudge has its own generator + a generator for the fire? Would it be possible to instead organize it so that a generator can spawn and manage multiple objects? That should cut down on the gen spam and I think replicates how they worked in retail in cases.

Lidefeath:
Everything you asked would be possible. I had a better - unmerged - generator code version where a generator could spawn 1 or many weenies (objects) and/or other generators that could spawn 1 or many weenies.
After discussing it with you , you wanted it more aligned with the current generator weenies that we know. 
After looking at those I proposed this model where a generator spawns a single object by its weenie but also can spawn other generators which then could spawn weenies. Theoretically you could let those generators spawn other generators again, but I am not sure if we want that. Looking at all the discovered weenies it's possible that this generator chain was several elements deep.
On the other hand we need a plan on how to populate these structures too: If we don't have the data for the complex solution then we don't need the complex solution.

If I were to design the next step I 'd introduce one or max 2 tiers of sub generators for now. But limit the spawning to 1 weenie per generator.

Mogwai-TheFurry:
My thoughts:
1) I know way less than you guys about this stuff, so feel free to slap me and say I'm wrong.

2) Based on what I remember of game play, I have always thought that there would be 1 top-level generator for a given camp / spawn point.  That top-level generator would pick a spawn set from a table and just spawn it all.  That is, if it picked drudges and a camp fire, the campfire would spawn on the generator, a drudge a little to the west (facing east), another drudge to the east (facing west), etc.

To clarify... 1 generator -> 1 (server side) spawn table.  1 spawn table -> N (server side) weenies with 1 relative location each.

3) Recursive generators may have uses.  I like the theoretical use case of them, but I'm not sure how practical it would be.

Mogwai-TheFurry:
Dependency / commit chain got messed up so this conflicts with the #410 predecessor.

Lidefeath:
Imho, we should have 1-2 recursions with generators for the following scenarios:

Scenario A - random spawns on the landblock 
1. One recursion: You have one top-level generator per landblock for random spawns. It can spawn a random number of different camp-generators: i.e. the tumeroks or carenzis or bandits on Marae Lessel. And these generators could now directly spawn the different weenies for such a camp.

2. Two recursions: That same top-level generator spawns different camp-generators and those spawn the single generators for single monsters and those spawn the 1 weenie they are made for.

Scenario B - fixed spawn position in the landblock 
1. One recursion: You have a camp-generator at the specific location that spawns single generators for each weenie. And those spawn the one weenie directly.

2. No recursion: You have a camp-generator directly spawn the different weenies at this position.

3. No recursion: You have a single generator spawning one specific weenie, I.e. the cow in Holtburg

Conclusion:
Looking at those scenarios I'd vote for implementing 1 recursion so we have a solution for scenario A1 and B2 and B3 (or even B1 if we wanted)

LtRipley36706:
See #420

PR #419, opened 2017-06-27
Weenie caching to fix portals

Mogwai-TheFurry:


Source/ACE.Database/CachingWorldDatabase.cs
```diff
@@ -0,0 +1,95 @@
...
+        private object _weenieCacheMutex = new object();
+
+        public CachingWorldDatabase(IWorldDatabase wrappedDatabase)
+        {
+            _wrappedDatabase = wrappedDatabase;
+        }
+
+        public AceObject GetAceObjectByWeenie(uint weenieClassId)
+        {
+            if (!_weenieCache.ContainsKey(weenieClassId))
```
ddevec: This is most likely unsafe.  Unless ContainsKey is guaranteed atomic w.r.t. _weenieCache (if Dictionaries are hash maps under the hood its highly unlikely), then this will eventually crash the code.  Double-check locking relies on a strictly atomic and ordered pre-check (volatility guarantees ordering).

Source/ACE.Database/CachingWorldDatabase.cs
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+
+            return (AceObject)_weenieCache[weenieClassId].Clone();
+        }
+
+        public AceObject GetObject(uint aceObjectId)
+        {
+            // if they're asking for a weenie, just give them the weenie.
+            if (_weenieCache.ContainsKey(aceObjectId))
```
ddevec: Also unsafe, unless ContainsKey is atomic w.r.t. _weenieCache.

Mogwai-TheFurry:
switched to a ConcurrentDictionary.  blah.

PR #420, opened 2017-06-27
Add Generator object

LtRipley36706:
Included test data for training dungeon part 1 which should provide a few objects to help with system build out, such as inventory, respawning, object tracking between clients...

### DIFF
--- a/Source/ACE/Entity/Generator.cs
+++ b/Source/ACE/Entity/Generator.cs
@@ -6,13 +6,8 @@ using System.Diagnostics;
 
 namespace ACE.Entity
 {
-    public class Generator : CollidableObject
+    public class Generator : WorldObject
     {
-        public Generator(ObjectType type, ObjectGuid guid, string name, ushort weenieClassId, ObjectDescriptionFlag descriptionFlag, WeenieHeaderFlag weenieFlag, Position position)
-            : base(type, guid, name, weenieClassId, descriptionFlag, weenieFlag, position)
-        {
-        }
-
         public Generator(ObjectGuid guid, AceObject baseAceObject)
             : base((ObjectType)baseAceObject.ItemType, guid)
         {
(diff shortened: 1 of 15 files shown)


Source/ACE/Entity/Generator.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Diagnostics;
+
+namespace ACE.Entity
+{
+    public class Generator : WorldObject
+    {
+        public Generator(ObjectGuid guid, AceObject baseAceObject)
+            : base((ObjectType)baseAceObject.ItemType, guid)
+        {
+            Name = baseAceObject.Name ?? "NULL";
```
Mogwai-TheFurry: why all the commented code in a new class?  Can you document why all this isn't necessary and just nuke it?

Source/ACE/Entity/Monster.cs
```diff
@@ -23,6 +25,74 @@ public Monster(AceObject aceO)
             else
                 this.WeenieClassid = (ushort)(aceO.WeenieClassId - 0x8000);
```
Mogwai-TheFurry: Why are we manually copying all this data in a derived type?  Seems like this is a symptom of a bad constructor chain.

LtRipley36706: TBH, I just noticed that there was an issue with monsters showing up on radar, so I c&p'd from DebugObject which corrected that. Didn't go beyond that. I don't get what the weenieclassid ushort stuff is either.

Source/ACE/Entity/Generator.cs
LtRipley36706: Started from DebugObject and then hard coded propertied I'm assuming to be immutable between generators and likely could be left out of the database in the future entirely. Left it commented for easier undo, can remove in the future when certain of the path. I'm also making the assumption that code from GeneratorFactory eventually ends up in this class instead and we make use of actionchaining to spawn/despawn, etc

PR #421, opened 2017-06-27
Master Preload

Forbiddenz:
Added ability to preload all the landblocks in the game.
This is a massive process, but its a great tool for testing or even benchmarking in the near future.


Forbiddenz:
Made Changes.

PR #423, opened 2017-06-27
Review and possible merge: Fix portal requirements

kdmxp256:
The code works during my functional test, but I don't know if it was the correct way to fix it. Also, there seems to be a slight delay when using a another portal, after having failed a different portal's activation requirements. Example, with a character that is lower than 80, try to use the Eastwatch portal in the TN Annex. Player gets denied because of being too low a level. Next use the Danby's Output portal next to the Eastwatch one. There is a slight delay before teleport. Subsequent portals that don't block do not have a delay, until the next time the player doesn't pass the requirements of another portal.

### DIFF
--- a/Source/ACE/Entity/Portal.cs
+++ b/Source/ACE/Entity/Portal.cs
@@ -123,12 +123,20 @@ namespace ACE.Entity
 
         public uint MinimumLevel
         {
-            get { return AceObject.GetIntProperty(PropertyInt.MinLevel) ?? 0; }
+            get
+            {
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
+                return weenie.GetIntProperty(PropertyInt.MinLevel) ?? 0;
+            }
         }
 
         public uint MaximumLevel
         {
-            get { return AceObject.GetIntProperty(PropertyInt.MaxLevel) ?? 0; }
+            get
+            {
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
+                return weenie.GetIntProperty(PropertyInt.MaxLevel) ?? 0;
+            }
         }
 
         public uint SocietyId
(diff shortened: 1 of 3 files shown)


Source/ACE/Entity/Portal.cs
```diff
@@ -138,23 +146,33 @@ public uint SocietyId
...
-            get { return ((AceObject.GetIntProperty(PropertyInt.PortalBitmask) ?? 0) & (uint)PortalBitmask.NoSummon) == 0; }
+            get
+            {
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
+                return ((weenie.GetIntProperty(PropertyInt.PortalBitmask) ?? 0) & (uint)PortalBitmask.NoSummon) == 0;
+            }
             set
             {
-                uint current = AceObject.GetIntProperty(PropertyInt.PortalBitmask) ?? 0;
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
```
Mogwai-TheFurry: No need for a setter here at all, IMO.

Source/ACE/Entity/Portal.cs
```diff
@@ -138,23 +146,33 @@ public uint SocietyId
...
-            get { return ((AceObject.GetIntProperty(PropertyInt.PortalBitmask) ?? 0) & (uint)PortalBitmask.NoRecall) == 0; }
+            get
+            {
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
+                return ((AceObject.GetIntProperty(PropertyInt.PortalBitmask) ?? 0) & (uint)PortalBitmask.NoRecall) == 0;
+            }
             set
             {
-                uint current = AceObject.GetIntProperty(PropertyInt.PortalBitmask) ?? 0;
+                var weenie = Database.DatabaseManager.World.GetAceObjectByWeenie(AceObject.WeenieClassId);
```
Mogwai-TheFurry: No need for a setter here at all, IMO.

kdmxp256:
I also added some debugging messages that show up in the chat window.  They demonstrate that delays experienced before the teleport with some portals happens at the player.Session.Player.Teleport(portalDest) command.

PR #424, opened 2017-06-27
quickfix

Forbiddenz:
Quick fix for 0 index bug.


Forbiddenz:
quick fix for missing 200 bug ripley found .. typo on the range.. 65225

PR #428, opened 2017-06-28
Load alb quick fix / diag fix

Forbiddenz:
Found small bug in diag tool + bug in load landblock.
should work now... hopefully..

Not the best at landblock understanding here.
But after talking with others.. it appears the landblock in ac are from 0 to 254 ?
I am still confused by this... and Honestly I am just tired of dealing with this... yet trying to fix it..

At this point.. someone else who is understands the landblock system better can patch it if its not right.


Mogwai-TheFurry:
This isn't the right fix, for what it's worth.

PR #429, opened 2017-06-28
Inventory save

Forbiddenz:
Ability to save inventory, code works but is not enabled due to deadlock issue with our main DB code.
I had some hack code to get around deadlock and it did save the items to the db correctly as far as I can tell.  We still will need load code of course..


Mogwai-TheFurry:
Overall, not how it needs to be done.  We'll get there with the DB synchronization update.

PR #430, opened 2017-06-28
Review: Extend generators to spawn linked generators

Lidefeath:
This introduces linked generators: install the sql script to be able to play with it.
With this you can spawn single monsters, spawn monster camps and have random spawns on a landblock.


Lidefeath:
Fixed the merge conflict.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
```diff
@@ -0,0 +1,122 @@
+USE `ace_world`;
+
+/* Create new generator link table */
+CREATE TABLE IF NOT EXISTS `ace_object_generator_link` (
+  `aceObjectId` int(10) unsigned NOT NULL,
+  `index` tinyint(3) unsigned NOT NULL,
+  `generatorId` int(10) unsigned NOT NULL,
```
LtRipley36706: `generatorId` column refers to the wcid of the generator to spawn or the generator spawning the aceObjectId? This reads like it also should have a FK cascade delete attached to it so that if either the generator or whatever generator is spawning is deleted, it is removed from the spawn table/links.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
```diff
@@ -0,0 +1,122 @@
...
+	(aceObjectId,
+	`index`,
+	generatorId,
+	generatorWeight)
+VALUES
+	(1996, 1, 4172, 100),
+	(4172, 1, 893, 60),
+	(4172, 2, 954, 40);
+	
+INSERT INTO ace_position
```
LtRipley36706: Not many generators would be one offs with unique positions I'm currently thinking, especially the ones you've setup here, they'd be instanced all over the world so I'm not sure the weenies would have positions.

See #437 and [ACE-World:#23](https://github.com/ACEmulator/ACE-World/pull/23) for what I'm proposing for how to setup the various instance clones of generator wcids. Changes are needed in the database to allow for proper inserting which ACE hadn't really tried to do yet.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
Lidefeath: Yes the column refers to the wcid of the generator to spawn as the linked generator to spawn will get its position  - either absolute or relative (= random in that landblock) - from its top level generator.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
Lidefeath: I'll have a look at your PR again. So far my thinking was that all static spawns in the world would have camp-generators I.e. Drudgecampgen as well as quest related generators. Futhermore each landblock would have a top-level generator that specifies which monster camps can spawn in that landblock.
I wouldn't see the positions as a position of the weenie, but as a position where the instance of that weenie is placed.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
Mogwai-TheFurry: generatorWcId or generatorWeenieClassId, then, please.

Database/Updates/World/06-06-27-2017-generator-chains-testdata.sql
LtRipley36706: I can't think of any reason a weenie should have a posititontype=1 in based on our current setup. For every weenie we have, there's at least two objects in ace_object if the weenie was spawned in world, the weenie object and the instance object.

Following your lead, lowaaluviangen could be an example of a landscape generator that was place all over the outskirts of holtburg right? I'd think potentially 1 for each landblock. lets just say there are 5 landblocks surrounding holtburg, So thats the weenie object (1996) with no position, then 5 instances of that weenie with a position somewhere within the landblock, perhaps the center of it just to help with whatever location randomization code we go with. The instances would end up having generated aceObjectIds from the insert scripts to prevent collisions.

drudgecampgen could be nested with options to spawn campfires, drudgesneakergens and drudgeskulkergens, or even just the wcids of those drudges themselves instead of generators right? really sort of up to whomever is setting up the generator I guess and assumes the responsibility of not screwing up a spawn table that ripples across the world ;)

drudgecampgen could be a spawn option of the lowaaluviangen so technically it needs no positions although of course there would be instances of the drudgecampgen as well as drudgesneakergen and drudgeskulkergen because they are such basic generators that had to have been widely used in the game

Lidefeath: In my line of thinking we'd only have the one lowaaluviangen in the world database as all the rest is randomly generated at runtime and doesn't need to be saved/persisted to the database. I have no problem using another positiontype than 1 if that helps.

LtRipley36706: the way landblock spawning works now is it needs an instanced object to place, how would we define the parameters of where a single weenie can be randomly spawned across the land other than using hand place or even script generated instances we shove into ace_world.

Lidefeath: Sorry, I don't get the problem. If you try out my code you'll see that the lowaaluvian generator which is in the DB - and the in the world - will spawn 10 camp generators in the world at random positions which then will create the appropriate single monster generators - which are not shown in the world to avoid clutter - and these spawn the monsters in the world. 
I wasn't convinced that spawning generators at all in the world is needed, but here you got them. So what's still missing or is wrong?

LtRipley36706:
Closing due to this being merged in with #442

PR #434, opened 2017-06-29
Appraise Items (Updated 7/2) READY FOR MERGE

ogmage78:
I switched tracks as I needed appraisal to work.   I have appraisal about 90% finished - it is functional and a good bit ahead of where we were.   I have put in the foundation for the correct combat stance - but I have more work to do as well as refactoring the code for stance.   However, this is a lot and a delayed merge could cause issues.   I would have separated them but I messed up and worked too much in the wrong branch.   I don't think anything I have done will regress us in anyway.


Source/ACE/Entity/Player.cs
```diff
@@ -1570,14 +1570,44 @@ protected override void SendUpdatePosition()
...
+            {
+                item.Parent = null;
+                item.ParentLocation = null;
+                item.CurrentWieldedLocation = null;
+                item.Location = null;
+                item.ContainerId = container.Guid.Full;
+                item.SetPhysicsDescriptionFlag(item);
+            }
+
+            inContainerChain.AddDelaySeconds(0.001);
```
ddevec: Not sure I understand this delay.  Weird code like this should at-minimum be commented.

Source/ACE/Entity/Player.cs
```diff
@@ -1763,33 +1794,106 @@ public void HandleActionWieldItem(Container container, uint itemId, uint locatio
...
+                                        break;
+                                    }
+                                default:
+                                    {
+                                        placementId = 1;
+                                        childLocation = 1;
+                                        break;
+                                    }
+                            }
+                            Children.Add(new EquippedItem(itemGuid.Full, (EquipMask)location));
```
ddevec: There are actually some interesting questions here about ownership.  We're making the item landblock visible.  The item should be on the landblock (the landblock needs a lookup ability if someone goes to examine it), but ultimately shouldn't the player be responsible for it?  E.g. who is responsible as the item's parent (e.g. calling its "Act" function when it had work to do -- like using item mana).  Who is responsible for updating/tracking the item's motion?

Source/ACE/Entity/Player.cs
```diff
@@ -1763,33 +1794,106 @@ public void HandleActionWieldItem(Container container, uint itemId, uint locatio
                             HandlePickupItem(container, itemGuid, location, PropertyInstanceId.Wielder);
                             return;
                         }
+
                         // Ok do you have this - if not bail
-                        if (GetInventoryItem(itemGuid) == null) return;
+                        var item = GetInventoryItem(itemGuid);
+                        if (item == null)
+                            return;
+                        if ((EquipMask)location > EquipMask.RingLeft)
```
ddevec: These nested conditionals are getting messy.  We should think if there is a more elegant solution to this problem, or they should be commented very well.

Source/ACE/Network/GameMessages/Messages/GameMessageparentEvent.cs
```diff
@@ -0,0 +1,19 @@
+﻿using ACE.Entity;
+using ACE.Network.Sequence;
+
+namespace ACE.Network.GameMessages.Messages
+{
+    public class GameMessageParentEvent : GameMessage
```
ddevec: Filename = GameMessageparentEvent, class name = GameMessageParentEvent.  Please fix filename.

Source/ACE/Entity/Player.cs
ogmage78: This is something I don't quite understand from looking at the live server - but I have some idea why it is the way it is.   Only wielded shields and weapons are selectable on both other players and on monsters.   In live, unlike other wielded items like armor, jewelry or clothes - these items and only these items had both a wielder id AND a position in the world.   I can only speculate that to make that selectable and visible to be cast upon, it had to be known to the landblock.   That does bring up this question.   You could only have two items a weapon / caster and optionally a shield.   Both of those items had a parent_id set which was the wielder.   Now all wielded items have wielder set but only those with a position that are selectable have the parent_id set.   I wonder if when we get positions updates or (are calculating them from physics) if we update the position of the message sender then update on the child objects the position = to the parent position.   They are always the same.   Not sure there, we will need to figure that out.

ogmage78:
Continued work. Added intProperty that was not being set (41) default combat style. Added to aceObject and surfaced on WO. Starting to use that data to determine appropriate combat stance.   This is way more complex that it looks on the surface.   Any review and help is appreciated.

Source/ACE/Entity/Player.cs
ogmage78: I am going to refactor that out.   I was trying to learn the action chain.

Source/ACE.Entity/AceObjectPropertiesSpell.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using ACE.Common;
+using MySql.Data.MySqlClient;
+namespace ACE.Entity
+{
+    [DbTable("ace_object_properties_spell")]
```
ddevec: I'm not sure I understand why we need a new table to represent spells -- an integer property of an object -- when we already have an int_properties table.  I'm not saying its wrong, I just don't immediately see why its needed

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1010,7 +1008,7 @@ public static void HandleCI(Session session, params string[] parameters)
             WorldObject testContainer = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
             if (testContainer.ItemCapacity > 1)
             {
-                var loot = LootGenerationFactory.CreateTestContainerObject(session.Player, weenieId);
+                Container loot = LootGenerationFactory.CreateTestContainerObject(session.Player, weenieId);
                 loot.ContainerId = session.Player.Guid.Full;
                 session.Player.AddToInventory(loot);
```
ddevec: Just noticed this -- AddToInventory modifies the player -- therefore it must be run within the player's actions (e.g. add this part to an ActionQueue action, then run that)

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1019,7 +1017,24 @@ public static void HandleCI(Session session, params string[] parameters)
             }
             else
             {
-                var loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
+                WorldObject loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
+                loot.ContainerId = session.Player.Guid.Full;
+                session.Player.AddToInventory(loot);
```
ddevec: Same as above -- must add on Player's action.

Source/ACE/Entity/Landblock.cs
```diff
@@ -321,16 +321,14 @@ private Position GetWorldObjectPosition(ObjectGuid objectId)
...
 
-            wo = worldObjects.ContainsKey(objectId) ? worldObjects[objectId] : null;
-            if (wo != null)
-            {
-                var csetup = SetupModel.ReadFromDat(wo.SetupTableId.Value);
-                return (float)Math.Pow((wo.UseRadius.Value + csetup.Radius + 1.5), 2);
-            }
-            return 0.00f;
+            WorldObject wo = worldObjects.ContainsKey(objectId) ? worldObjects[objectId] : null;
+            if (wo?.SetupTableId == null) return 0.00f;
```
ddevec: I'm pretty sure we have a no conditional-on-same-line policy.  (Personally, I strongly believe in always using curly braces as well, but I wont enforce that)

Source/ACE/Entity/WorldObject.cs
```diff
@@ -606,21 +638,91 @@ protected WorldObject(ObjectGuid guid, AceObject aceObject)
             PaletteGuid = aceObject.PaletteId;
         }
 
-        public void SetCombatMode(CombatMode newCombatMode)
+        public void SetCombatMode(CombatMode newCombatMode, Player player)
```
ddevec: This confuses me -- Why is SetCombatMode a function of WorldObject and not player?  Additionally, why is Player passed into it?  What context is it called from?  (e.g. is it safe to access the player from this context? -- probably not).

This smells funny to me, and I would like to understand it more.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -606,21 +638,91 @@ protected WorldObject(ObjectGuid guid, AceObject aceObject)
...
-                    var gm = new UniversalMotion(MotionStance.UANoShieldAttack);
-                    gm.MovementData.CurrentStyle = (ushort)MotionStance.UANoShieldAttack;
-                    SetMotionState(gm);
+                    UniversalMotion gm = null;
+                    WorldObject meleeWeapon = null;
+
+                    // Let's see what if anything is equipped and if we have a shield.
+                    EquippedItem mEquipedMelee = player.Children.Find(s => s.EquipMask == EquipMask.MeleeWeapon);
+                    if (mEquipedMelee != null)
+                        meleeWeapon = player.GetInventoryItem(new ObjectGuid(mEquipedMelee.Guid));
```
ddevec: Are you sure you can access Player here (and other places within this function) safely (from a concurrency prospective)?

Source/ACE/Network/GameAction/Actions/GameActionChangeCombatMode.cs
```diff
@@ -9,7 +10,7 @@ public static class GameActionChangeCombatMode
         public static void Handle(ClientMessage message, Session session)
         {
             var newCombatMode = message.Payload.ReadUInt32();
-            session.Player.SetCombatMode((CombatMode)newCombatMode);
+            session.Player.SetCombatMode((CombatMode)newCombatMode, session.Player);
```
ddevec: This is not going to be safe -- need a HandleActionSetCombatMode function for concurrency correctness -- really that whole function construction looks hacky to me -- Why is it in WorldObject and not Player.  Why is Player passed in (as opposed to making it virtual -- if it really should be defined in WorldObject, etc).

Source/ACE/Entity/Player.cs
```diff
@@ -1796,33 +1827,106 @@ public void HandleActionWieldItem(Container container, uint itemId, uint locatio
...
+                                    new GameMessageUpdateInstanceId(container.Guid, new ObjectGuid(0),
+                                        PropertyInstanceId.Container),
+                                    new GameMessageUpdateInstanceId(container.Guid, itemGuid,
+                                        PropertyInstanceId.Wielder),
+                                    new GameMessagePrivateUpdatePropertyInt(Session, PropertyInt.CurrentWieldedLocation,
+                                        location));
+                                CurrentLandblock.EnqueueBroadcast(
+                                    Location,
+                                    Landblock.MaxObjectRange,
+                                    new GameMessageCreateObject(item),
```
ddevec: You've got some duplicated broadcasts here -- e.g. GameMessageCreateObject is sent directly to the client -- then again on broadcast.

Source/ACE.Entity/AceObjectPropertiesSpell.cs
ogmage78: That is not a new table - it has always been in ace_world and is populated by the main data load.   It was just never wired up.   All I did was wire it up and load the values.   It is used as part of the magic system for items to store what enchantments are permanently part of the item.   I needed to implement it for appraisal event to work correctly.   It is one of the tables just like ace_properties_int - it was just never wired up before now.

Source/ACE/Command/Handlers/AdminCommands.cs
ogmage78: I did not go back and retrofit old code.   That has been in there forever.   I can put it in an actionQueue.

Source/ACE/Entity/Landblock.cs
ogmage78: Yea that just slipped by me - I will fix it.

Source/ACE/Entity/WorldObject.cs
ogmage78: It will need to be moved to creature - they go into combat mode as well.   I just made a mistake and put this in the wrong object.

Source/ACE/Entity/WorldObject.cs
ogmage78: probably not.   We were not doing concurrency before and it is taking me a lot longer to come up to speed on that than I would like.

Source/ACE.Entity/AceObjectPropertiesSpell.cs
LtRipley36706: @ddevec an AceObject can have many spells on it, these are different than the primary Spell which comes from DID table. As is currently the case, PropertyInts can only have a single entry per.

Source/ACE/Network/GameAction/Actions/GameActionChangeCombatMode.cs
ogmage78: probably just a lack of understanding on my part.   I am not sure, with all of these issues - should I just close this PR?

Source/ACE.Entity/AceObjectPropertiesSpell.cs
ddevec: No problem.  I didn't realize the table had already existed.  Wanted to make sure we weren't making unnecessary DB changes.  This is fine.

Source/ACE/Command/Handlers/AdminCommands.cs
ddevec: Yeah, I only noticed it because of the lines added below.  It should be an easy fix though.

Source/ACE/Entity/WorldObject.cs
ddevec: If you make my "HandleActionChangeCombatMode" suggestion below, and move this function into Creature, this should be fine (I think).

Source/ACE/Network/GameAction/Actions/GameActionChangeCombatMode.cs
ddevec: No, your close to correct -- just this one part needs to be fixed up.

GameAction messages come in from the network (and are therefore run in network threads).  In order for the network threads to interact with game state safely, we need to add some synchronization.

By convention I've done this in Player.cs by creating functions named HandleAction<thing>, which use an ActionChain to move the work onto the player.  That should be an acceptable solution here to solve your concurrency issues (if you move the logic in the ChangeCombatMode to the appropriate place it should take care of your issues there too).

Source/ACE/Entity/Player.cs
ogmage78: Removed

ddevec:
Accepting -- but adding note that we need to fix concurrency on GameActionChagneCombatMode

PR #437, opened 2017-06-30
Of Guids and Generators...

LtRipley36706:
* Changed WorldBase to allow for auto incrementing of aceObjectId and weenieClassId
* Changed WEENIE_MAX to be 199999 to account for ACE (re)created weenies starting at 100000
* Several changes to GuidManager and redirected CommonObjectFactory to use it instead of its own list
* Fixed some constructor issues related to Guids. Not sure if it is the right way to go about it but does seem to work in testing.
* Altered GameEventIdentifyObjectResponse to output debug stuff for more objects temporarily and fixed its output of Guids in decimal form.


Source/ACE.Entity/AceObject.cs
```diff
@@ -11,7 +11,7 @@ namespace ACE.Entity
     [DbTable("ace_object")]
     public class AceObject : ICreatureStats, ICloneable
     {
-        public const ushort WEENIE_MAX = 0xFFFF;
+        public const uint WEENIE_MAX = 199999;
```
ddevec: 199999 is a somewhat arbitrary number.  Why this number?  What is the actual maximum we can set it to?  Before we chose 0xFFFF because it fit a ushort.  Are we sure weenies don't have to be constrained to a ushort for some (likely client related) reason?

Source/ACE/Managers/GuidManager.cs
```diff
@@ -14,25 +14,33 @@ public static class GuidManager
         // players
         private const uint playerMin = 0x50000001;
         private const uint playerMax = 0x59999999;
-        private static uint player = 0x50000001;
+        private static uint player   = 0x50000001;
+
+        // admins
+        private const uint adminMin = 0x80000001;
+        private const uint adminMax = 0x89999999;
```
ddevec: 0x89999999 is an arbitrary number.  Did you mean 0x8FFFFFFF, or perhaps 0x88000000?

Source/ACE/Managers/GuidManager.cs
```diff
@@ -14,25 +14,33 @@ public static class GuidManager
...
 
+        // Orginally found in CommonObjectFactory
+        // private static uint nextObjectId = 0x80000001;
+
         // Npc / Doors / Portals / Etc
-        private const uint staticObjextMin = 0x6000001;
-        private const uint staticObjextMax = 0x6999999;
-        private static uint staticObject = 0x6000001;
+        private const uint staticObjectMin = 0x6000001;
+        private const uint staticObjectMax = 0x6999999;
```
ddevec: Same as above 9999 vs ffff

Source/ACE/Managers/GuidManager.cs
```diff
@@ -14,25 +14,33 @@ public static class GuidManager
...
+        private static uint staticObject   = 0x6000001;
 
         // Monsters /
-        private const uint monsterMin = 0x8000001;
-        private const uint monsterMax = 0x8999999;
+        private const uint monsterMin     = 0x8000001;
```
ddevec: This overlaps with the admin range.  That can't be a good thing.

Source/ACE/Managers/GuidManager.cs
```diff
@@ -45,6 +53,16 @@ public static uint NewPlayerGuid()
         }
 
         /// <summary>
+        /// Returns New Admin Guid
+        /// </summary>
+        /// <returns></returns>
+        public static uint NewAdminGuid()
```
ddevec: This function is clearly not thread-safe, but will likely be used (or try to be used) from a threaded context.  Either make it thread safe with a lock (recommended) or (if your feeling fancy) atomic increment (https://msdn.microsoft.com/en-us/library/dd78zt0c(v=vs.110).aspx, note you'll have to do casting to make this work), or put in the header the exact contexts it can be called from.

Source/ACE.Entity/AceObject.cs
LtRipley36706: In theory, the maximum it could be set to would be what ever number you choose to start actual spawned instances of ace_objects at minus 1. For every weenie you have there is an entry in both the ace_weenie_class table and ace_object both ids being used for their index, so to prevent overlap of an instance a ACE generated instance or pulled from PCAP instance you want to pick a limit.

Likely we could change 199999 to 19999999 to match the true amount of space I'm building in ACE-World for any newly generated weenies, but of course its highly unlikely we'd use anywhere near that much.

The number is "compressed" by a function call so while the client always reads it as a ushort as I understand it, it gets converted because otherwise the client was always limited to less than 32K weenies and we know they went to the 50K+ range.

Source/ACE/Managers/GuidManager.cs
LtRipley36706: Copied from player, changed from 5 to 8 based on what I'd seen in used in the past.

Source/ACE/Managers/GuidManager.cs
LtRipley36706: Function is never used TBH

Source/ACE/Managers/GuidManager.cs
LtRipley36706: does it? 0x8000001 is the same as 0x80000001 ?

ddevec: Mis-counted my 0's.  This is fine.

Source/ACE/Managers/GuidManager.cs
LtRipley36706: No doubt here and I also wonder if GuidManager is generating the ids for what eventually is stored in ace_shard when saving inventory or whatever, if so.. how does it know which ids are taken up (it doesn't)

I just made sure everything that generates ids is using the same function now. I would most certainly need yours or others assistance to make this thing function as it appears to be meant for

ogmage78: I was looking at that earlier - I don't think that code works as none of the updated values are persisted.   I think it was put in more as a stub until we started saving items because every server restart resets those values (i think) I have not spent a lot of time looking at it tbh.

LtRipley36706: in the sense of spawning things in world, it works, but you're absolutely correct it needs to be looked over and fleshed out so that it can be used to generate correct guids for new items to be saved in ace_shard, and persist so that it doesn't reuse incorrectly. gotta get everything on the same page first and then we have a fighting chance :)

PR #439, opened 2017-06-30
Fixed serialization bug.

ogmage78:
* Fixed issue with the wrong data size being used in the weenie data serialization.
* Structure, Max Structure, StackSize and MaxStackSize were being sent as uint and they are ushort.  This caused intermittent alignment issues.

### DIFF
--- a/Source/ACE/Entity/WorldObject.cs
+++ b/Source/ACE/Entity/WorldObject.cs
@@ -843,16 +843,16 @@ namespace ACE.Entity
                 writer.Write((sbyte?)CombatUse ?? 0);
 
             if ((WeenieFlags & WeenieHeaderFlag.Structure) != 0)
-                writer.Write(Structure ?? 0);
+                writer.Write((ushort?)Structure ?? 0);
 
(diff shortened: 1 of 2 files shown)


Source/ACE/Entity/WorldObject.cs
```diff
@@ -843,16 +843,16 @@ public virtual void SerializeCreateObject(BinaryWriter writer)
                 writer.Write((sbyte?)CombatUse ?? 0);
 
             if ((WeenieFlags & WeenieHeaderFlag.Structure) != 0)
-                writer.Write(Structure ?? 0u);
+                writer.Write(Structure ?? 0);
```
Mogwai-TheFurry: Just plain "0" will default to a signed int.  You need to specify "ushort" in an explicit cast if that's what you need.

ogmage78: Made requested changes

Mogwai-TheFurry:
Casting has a higher order of operation than the null coalesce:
http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-334.pdf page 151

Now you're casting the variable and not the 0.

PR #440, opened 2017-06-30
cleaned up compiler warnings, minor tweaks to shard db interface

Mogwai-TheFurry:


Source/ACE/Entity/Player.cs
```diff
@@ -1076,47 +1076,50 @@ public void AddFriend(string friendName)
...
-
-            // Remove from character object
-            Character.RemoveFriend(friendId.Low);
-
-            // Send packet
-            Session.Network.EnqueueSend(new GameEventFriendsListUpdate(Session, GameEventFriendsListUpdate.FriendsUpdateTypeFlag.FriendRemoved, friendToRemove));
+            DatabaseManager.Shard.DeleteFriend(Guid.Low, friendId.Low, (() =>
+            {
+                // Remove from character object
+                Character.RemoveFriend(friendId.Low);
```
ddevec: I don't believe that the AceCharacter "Friend" functions are thread safe.  Your callback is running from outside of the session's context.  These are likely racy accesses.

Source/ACE/Entity/Player.cs
```diff
@@ -1076,47 +1076,50 @@ public void AddFriend(string friendName)
...
 
                 // Save to DB, assume success
-                DatabaseManager.Shard.AddFriend(Guid.Low, newFriend.Id.Low);
-
-                // Add to character object
-                Character.AddFriend(newFriend);
+                DatabaseManager.Shard.AddFriend(Guid.Low, newFriend.Id.Low, (() =>
+                {
+                    // Add to character object
+                    Character.AddFriend(newFriend);
```
ddevec: I don't believe that the AceCharacter "Friend" functions are thread safe. Your callback is running from outside of the session's context. These are likely racy accesses.

Mogwai-TheFurry: This was already in the Task handler for GetObjectInfoByName.  This is not an issue introduced by this PR.

PR #441, opened 2017-06-30
SimpleCrudeInventorySave

Forbiddenz:
Seems to work, saves fine now with no deadlocks..

Only concern now is, how to do the checking and only update properties that are dirty.
Some guidance would be appreciated .

### DIFF
--- a/Source/ACE/Entity/Container.cs
+++ b/Source/ACE/Entity/Container.cs
@@ -1,8 +1,9 @@
-using System;
+﻿using System;
 using ACE.Entity.Enum;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


Source/ACE/Entity/Container.cs
```diff
@@ -41,6 +42,20 @@ public Container(AceObject aceObject, ObjectGuid guid)
         }
 
         // Inventory Management Functions
+        public virtual void SaveInventory()
```
Mogwai-TheFurry: I don't think Container is the place for this.  The containers that save to the database are special cases.  Namely, housing storage and players.  It should go in those places.

PR #447, opened 2017-07-04
Fix Concurrency Issues and move items from admin to debug

ogmage78:
Moved items out of admin that were not part of turbine admin functions over to debug commands.   Added actions to action chains for concurrency safety.


Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -705,8 +709,65 @@ public static void HandleListPlayers(Session session, params string[] parameters
...
+                    numItems = Convert.ToByte(parameters[1]);
+                }
+                catch (Exception)
+                {
+                    ChatPacket.SendServerMessage(session, "Not a valid number - must be a number between 0 - 50", ChatMessageType.Broadcast);
+                    return;
+                }
+            }
+            ActionChain chain = new ActionChain();
+            chain.AddAction(session.Player, () => LootGenerationFactory.CreateRandomTestWorldObjects(session.Player, typeId, numItems));
```
ddevec: Same as above

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -705,8 +709,65 @@ public static void HandleListPlayers(Session session, params string[] parameters
...
+        public static void HandleWeapons(Session session, params string[] parameters)
+        {
+            HashSet<uint> weaponsTest = new HashSet<uint>() { 93, 148, 300, 307, 311, 326, 338, 348, 350, 12748, 12463, 31812 };
+            ActionChain chain = new ActionChain();
+
+            chain.AddAction(session.Player, () =>
+            {
+                foreach (uint weenieId in weaponsTest)
+                {
+                    WorldObject loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
```
ddevec: I recognize this function is not introduced by your PR, so I wont reject, but...

I'm not a big fan of the LootGenerationFactory modifying the player in the player's context -- logically the Player should modify itself in its own context. IMO: The job of a LootGenerationFactory is to generate loot, not modify players.

This interface is just asking to be abused.  By accepting a player argument it hints that it will safely modifies the player, which it does not.

PR #448, opened 2017-07-04
Barber

OptimShi:
Added "@barbershop" debug command and corresponding functionality to load the interface, and save and apply settings.
Changed some CharGen types that were ushort to uints. (They were short for legacy reasons no longer needed)

Not sure if the HandleActionFinishBarber() function is the "correct" method, so please feel free to point out what I may not be doing the "proper way".

### DIFF
--- a/Source/ACE/Entity/Player.cs
+++ b/Source/ACE/Entity/Player.cs
@@ -201,6 +201,7 @@ namespace ACE.Entity
             Sequences.AddOrSetSequence(SequenceType.PrivateUpdatePropertyInt64, new ByteSequence(false));
             Sequences.AddOrSetSequence(SequenceType.PrivateUpdatePropertyDouble, new ByteSequence(false));
             Sequences.AddOrSetSequence(SequenceType.PrivateUpdatePropertyString, new ByteSequence(false));
+            Sequences.AddOrSetSequence(SequenceType.PrivateUpdatePropertyDataID, new ByteSequence(false));
 
             // This is the default send upon log in and the most common.   Anything with a velocity will need to add that flag.
             PositionFlag |= UpdatePositionFlag.ZeroQx | UpdatePositionFlag.ZeroQy | UpdatePositionFlag.Contact | UpdatePositionFlag.Placement;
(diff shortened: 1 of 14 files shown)


RoboticOxygen:
tried it out, very cool

Source/ACE/Entity/Player.cs
```diff
@@ -1577,6 +1578,73 @@ private void HandleActionTeleToMarketplaceInternal()
             DoMotion(motionMarketplaceRecall);
         }
 
+        public void HandleActionFinishBarber(ClientMessage message)
+        {
+            // Read the payload sent from the client...
+            PaletteGuid = message.Payload.ReadUInt32();
+            Character.HeadObject = message.Payload.ReadUInt32();
```
ddevec: These operations modify the Player (through the character).  Therefore they need to be run as part of the Player's Action's.  Do this through an ActionChain (just put the entire thing inside an ActionChain).

Source/ACE/Network/GameEvent/Events/GameEventStartBarber.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        public GameEventStartBarber(Session session)
+            : base(GameEventType.StartBarber, GameMessageGroup.Group09, session)
+        {
+            // These are the motion tables for Empyrean float and not-float (one for each gender). They are hard-coded into the client.
+            const uint EmpyreanMaleMotionDID = 0x0900020Eu;
+            const uint EmpyreanFemaleMotionDID = 0x0900020Du;
+
+            // We will use this function to get the current player's appearance values.
+            AceCharacter characterClone = (AceCharacter)Session.Player.GetAceObject();
```
ddevec: This accesses the player.  If the player is modified while its being cloned it may crash the server.  Put this inside an ActionChain to mitigate that issue.

OptimShi:
I moved the Barber functions into an ActionChain, as requested.

Source/ACE/Network/GameAction/Actions/GameActionFinishBarber.cs
```diff
@@ -13,7 +7,9 @@ public static class GameActionFinishBarber
         [GameAction(GameActionType.FinishBarber)]
         public static void Handle(ClientMessage message, Session session)
         {
-            session.Player.HandleActionFinishBarber(message);
+            ActionChain chain = new ActionChain();
+            chain.AddAction(session.Player, () => session.Player.HandleActionFinishBarber(message));
```
ddevec: I would prefer that the action chain is within HandleActionFInishBarber (HandleAction indicates that the function is safe to call from outside of the player's action context).  I wont block the PR over it.

OptimShi: I will adjust this, then. Because no one, including myself, will ever go back and do it later! Hold off on pushing, please.

PR #451, opened 2017-07-06
Wastin' away again in Splitsville 

ogmage78:
* Initial work on splitting a stackable item.   This is working and in an action chain.   Right now, it is only working for items in your possession.   I need to extend it to work with items in chests and off the ground.
* There is some new type of sequence that we need to research and implement.   I put a big todo in the code.
* I made some posts about getting guid manager going.   We are at a point where we need it.   I modified some of the ranges to reflect the current allocations .   This makes really poor use of our sequence space but resolution is TBD.
* I added in the method that @ddevec  wrote for player that we can propagate pattern for the other ranges.

* TODO - find out about sequence used in split message. Next implement stack which is the complement function from split.

### DIFF
--- a/Source/ACE.Database/ShardDatabase.cs
+++ b/Source/ACE.Database/ShardDatabase.cs
@@ -1,10 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading.Tasks;
 using ACE.Entity;
 using ACE.Entity.Enum;
-using ACE.Entity.Enum.Properties;
 using log4net;
 // ReSharper disable InconsistentNaming
 
@@ -111,7 +109,7 @@ namespace ACE.Database
             DeleteAceObjectPropertyAttribute = 155,
             DeleteAceObjectPropertyAttribute2nd = 156,
             DeleteAceObjectPropertySkill = 157,
-            // XXX(ddevec): @Og II -- use these to query max values
+            // Used to get max values from DB
             GetMaxPlayerId = 200,
         }
 
@@ -542,9 +540,9 @@ namespace ACE.Database
         public bool SaveObject(AceObject aceObject)
         {
             DatabaseTransaction transaction = BeginTransaction();
-            
+
             SaveObjectInternal(transaction, aceObject);
-            
+
             return transaction.Commit().Result;
         }
 
@@ -580,7 +578,7 @@ namespace ACE.Database
         {
             // Update the character table -- save the AceObject to ace_object.
             SaveAceObjectBase(transaction, aceObject);
-            
+
             SaveAceObjectPropertiesInt(transaction, aceObject.AceObjectId, aceObject.IntProperties);
             SaveAceObjectPropertiesBigInt(transaction, aceObject.AceObjectId, aceObject.Int64Properties);
             SaveAceObjectPropertiesBool(transaction, aceObject.AceObjectId, aceObject.BoolProperties);
@@ -894,7 +892,7 @@ namespace ACE.Database
             attribs.ForEach(a => a.AceObjectId = aceObjectId);
 
             var theDirtyOnes = attribs.Where(x => x.IsDirty).ToList();
-            
+
             foreach (var prop in theDirtyOnes)
             {
                 if (prop.HasEverBeenSavedToDatabase)
@@ -921,7 +919,7 @@ namespace ACE.Database
 
             // calling ToList forces the enumerable evaluation so that we can call .Remove from within the loop
             var theDirtyOnes = tempSkills.Where(x => x.IsDirty).ToList();
-            
+
             foreach (var prop in theDirtyOnes)
             {
                 if (prop.HasEverBeenSavedToDatabase)
--- a/Source/ACE.Database/ShardDatabase.cs
+++ b/Source/ACE.Database/ShardDatabase.cs
@@ -85,7 +85,9 @@ namespace ACE.Database
             InsertAceObjectPropertiesPositions = 132,
             InsertAceObjectPropertiesAttributes = 133,
             InsertAceObjectPropertiesAttributes2nd = 134,
-            InsertAceObjectPropertiesSkills = 135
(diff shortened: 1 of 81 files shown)


ogmage78:
https://www.youtube.com/watch?v=CICf8xoLyG8
Going to leave this here just 'cause

Source/ACE.Entity/AceObject.cs
```diff
@@ -973,6 +973,34 @@ public object Clone()
             return ret;
         }
 
+        public object CopyClone(uint guid)
```
ddevec: I don't understand the distinction between CopyClone and clone.  Why do we need two separate functions, and what is the difference?

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -738,6 +738,32 @@ public static void HandleWeapons(Session session, params string[] parameters)
...
         }
 
+        [CommandHandler("splits", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0,
+            "Creates some stackable items in your inventory for testing.")]
+        public static void HandleSplits(Session session, params string[] parameters)
+        {
+            HashSet<uint> splitsTest = new HashSet<uint>() { 237, 300, 690, 20630, 20631, 37155, 31198 };
+            ActionChain chain = new ActionChain();
+
+            chain.AddAction(session.Player, () =>
```
ddevec: Actions on the first try.  Good man.

Source/ACE/Managers/GuidManager.cs
```diff
@@ -1,46 +1,66 @@
...
-        private static uint player   = 0x50000001;
+        // Running server is guid master - database only read as startup to get current max per range.
+        // weenie class templates Max 65,535 - took Turbine 17 years to get to 10K
+        // these will be added by developers and not in game.
+        // Nothing in this range is persisted by the game.   Only developers or content creators can create them to be persisted.
+        // Fragmentation: None
+        private const uint weenieMin = 0x00000001;
+        private const uint weenieMax = 0x0000FFFF;
+        // At Server startup read current max from the DB
+        private static uint weenie   = 0x00001344;
```
ddevec: I'm hesitant to include hard-coded values for these (all initial values here).  They should be read from the DB at startup.

Source/ACE.Entity/AceObject.cs
ogmage78: Fixed - I had a brain freeze.   I need to be able to assign a new aceobjectid but I could do that with no duplicate code - just a new signature.

Source/ACE/Managers/GuidManager.cs
ogmage78: I made extensive posts in our dev channel.   Until we actually save nothing is persisted.   Also, as I put in the comments - if everyone is ok with this, we will remove the hard coded current values and they will be read in at server start up.   This is really just POC so we can see if we want to take this path.

ddevec: I really don't like this -- its asking for future bugs.

It doesn't seem that hard to me to read the appropriate DB tables and find the MAX(id) for each appropriate range.  I would prefer someone else's opinion on this before merge.

ogmage78: That is fine - this can just sit.  We have nothing now and I was not going to move ahead with it until we got some consensus and feedback which no one responded either way.  I really can't move forward on this for the other 4 use cases until this is merged.

Mogwai-TheFurry: I have no objections to the hard-coding the weenie ID ranges.  If it makes you feel better, we can even put contraints on the database to mirror these values.  IIRC, these are hard coded in the client as well.

The "current max" does seem a bit... unnecessary.  Ctrl+F suggests nothing is calling NewWeenieGuid( ) - what is it intended for?

ogmage78:
@Mogwai-TheFurry  I removed that method.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -85,7 +85,9 @@ private enum ShardPreparedStatement
             InsertAceObjectPropertiesPositions = 132,
             InsertAceObjectPropertiesAttributes = 133,
             InsertAceObjectPropertiesAttributes2nd = 134,
-            InsertAceObjectPropertiesSkills = 135
+            InsertAceObjectPropertiesSkills = 135,
+            // XXX(ddevec): @Og II -- use these to query max values
```
ddevec: Please remove this comment

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -156,6 +164,21 @@ protected override void InitializePreparedStatements()
             ConstructStatement(ShardPreparedStatement.InsertTextureOverridesByObject, typeof(TextureMapOverride), ConstructedStatementType.InsertList);
             ConstructStatement(ShardPreparedStatement.InsertPaletteOverridesByObject, typeof(PaletteOverride), ConstructedStatementType.InsertList);
             ConstructStatement(ShardPreparedStatement.InsertAnimationOverridesByObject, typeof(AnimationOverride), ConstructedStatementType.InsertList);
+
+            // FIXME(ddevec): Use max/min values defined in factory -- this is just for demonstration purposes
```
ddevec: I really dislike this FIXME.  I guess we can keep it for now and make a trello card about actually doing Guid manager allocation properly

Source/ACE/Managers/GuidManager.cs
```diff
@@ -47,29 +59,63 @@ public static class GuidManager
...
+                {
+                    Monitor.Wait(ShardLock);
+                }
+            }
+
+            // NOTE: + 1 -- we don't want to reassign our current guid
+            return ret + 1;
+        }
+
+        // FIXME(ddevec): @Og -- hopefully you can use this
```
ddevec: Please change this FIXME to address that we need to do this for all initial ranges

ogmage78:
Comments updated and I am adding the trello card now.

PR #452, opened 2017-07-07
Character list will automatically select your last played character.

OptimShi:


OptimShi:
Note that this requires an updated `vw_ace_character` view. SQL is in the "Database/Updates/Shard" folder.

PR #456, opened 2017-07-08
Motion State Flags

ogmage78:
* Fixed issue with MotionStateFlag - setting now by the existence of data like we do with all other flags.   This prevents data / flag mismatch


Source/ACE/Entity/MovementData.cs
```diff
@@ -250,6 +169,29 @@ public MovementData ConvertToClientAccepted(uint holdKey, CreatureSkill run)
             return md;
         }
 
+        public void SetMovementStateFlag()
+        {
```
ddevec: I'm going to be nit-picky here -- I don't like having to remember/know to call this function before we get the MovmentStateFlag.  Can we make the accessor (get) of MovementStateFlag do this computation?

ogmage78: Flags are only used when we serialize - I can move it to the serialize method and call it there?

ogmage78: I made the changes you requested.

ogmage78: It was working the way I had it, but it is not working in accessor.   I am trying to debug now.

ogmage78: I have tried to do it the way you requested - but it does not seem to work.   It seems that it should I must have something obviously wrong.

ogmage78: I found the bug in this.   You can merge it if you want to.

PR #464, opened 2017-07-12
Enum Reshuffle

LtRipley36706:


Source/ACE.Entity/Enum/IdentifyResponseFlags.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace ACE.Network.Enum
+{
+    [Flags]
+    public enum IdentifyResponseFlags
```
ddevec: Is the an entity or network enum?  I agree most of these should be in entity.  I'm not 100% sure all of them belong in there.

Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
```diff
@@ -282,5 +285,177 @@ private string DebugOutputString(System.Type type, WorldObject obj)
 
             return debugOutput;
         }
+
+        private void WriteIdentifyObjectHeader(BinaryWriter writer, IdentifyResponseFlags flags, bool success)
```
ddevec: These functions could be private static void -- they don't use any object local data -- they are purely functional.

Source/ACE.Entity/Enum/IdentifyResponseFlags.cs
LtRipley36706: Yeah forgot I fixed this but didn't save

PR #465, opened 2017-07-13
Restructuring AceObject and other objects

LtRipley36706:
Please review changes to objects and note any concerns with continuing down this road...

The idea is to clean up and strip down AceObject to let the property get/set functions handle most everything not DbField and start to move, keep stuff in their appropriate object classes.

I'd like to move the ICreatureStats out as well but not totally sure how to get that out and in to AceCharacter, as not every worldobject or aceobject needs those ability fields


Source/ACE/Entity/WorldObject.cs
```diff
@@ -193,166 +188,199 @@ public ContainerType ContainerType
...
         public ushort? MaxStackSize
         {
-            get { return AceObject.MaxStackSize; }
-            set { AceObject.MaxStackSize = value; }
+            get { return (ushort?)AceObject.GetIntProperty(PropertyInt.MaxStackSize); }
+            set { AceObject.SetIntProperty(PropertyInt.MaxStackSize, value); }
         }
 
-        public uint? ContainerId
+        public uint? Container
```
Mogwai-TheFurry: "Container" is an object, this is an identifier.  ContainerId is strongly preferred.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -414,92 +442,92 @@ public float Workmanship
...
-            get { return AceObject.PhysicsTableId; }
-            set { AceObject.PhysicsTableId = value; }
+            get { return AceObject.GetDataIdProperty(PropertyDataId.PhysicsEffectTable); }
+            set { AceObject.SetDataIdProperty(PropertyDataId.PhysicsEffectTable, value); }
         }
 
         /// <summary>
         /// This is used for equiped items that are selectable.   Weapons or shields only.   Max 2
         /// </summary>
         public uint? Parent
```
Mogwai-TheFurry: I realize this isn't your change, but similar to Container, this should be ParentId.

Source/ACE/Factories/GeneratorFactory.cs
```diff
@@ -65,34 +66,34 @@ public static List<WorldObject> CreateWorldObjectsFromGenerator(AceObject genera
...
-            for (int i = 0; i < generator.MaxGeneratedObjects; i++)
+            for (int i = 0; i < generator.GetIntProperty(PropertyInt.MaxGeneratedObjects); i++)
             {
-                switch (generator.GeneratorType)
+                switch (generator.GetIntProperty(PropertyInt.GeneratorType))
                 {
                     // Use the position of the generator as a static position
                     case (int)GeneratorType.Absolute:
-                        pos = generator.Location.InFrontOf(2.0);
+                        pos = generator.GetPosition(PositionType.Location).InFrontOf(2.0);
```
Mogwai-TheFurry: You're doing InFrontOf, but the comment says "use the position of the generator".  I know, you didn't change that, but at least add a "// TODO" comment if you don't want to fix it.

LtRipley36706: the position of the generator object is moved backwards 2.0 prior to this code line, so it's just setting the object's position to the correct spot.

LtRipley36706:
Published some work on Door.

The thought behind this was that AceObject is supposed to be the object that goes in to or comes out of the database, everything that then goes in world is either a WorldObject or something that derives(?) from a WorldObject. All the generic or universal properties/fields are set up in it, much like was already halfway occuring, while some stuff was in AceObject.. beyond that then from WorldObjects, other objects expose the properties they would use.

Source/ACE/Entity/Door.cs
```diff
@@ -21,30 +24,120 @@ public Door(AceObject aceO)
...
-                ActionChain autoCloseTimer = new ActionChain();
-                autoCloseTimer.AddDelaySeconds(autoCloseTime);
-                autoCloseTimer.AddAction(this, () => Close());
-                autoCloseTimer.EnqueueChain();
-            }
-            else
-            {
-                Close();
-            }
+        public bool IsLocked
```
saday-cimpress: I'm betting we don't need to save "IsLocked" in the database.  If something defaults to locked, we spawn it locked until somebody unlocks it.

ogmage78:
These are my thoughts on AceObject.    I admit, I am very new to the whole ORM so take it for what it is worth.   I like the simplicity of using AceObject to encapsulate our persistence.   If it exists as part of the aceObject set of tables, we load it.   That makes it's use very hard to screw up.   I am always in favor of that.   Because of the way our tables are designed, items that do not apply do not exist in the database.   For example, making a call to load attributes on an object that does not have any incurs very little overhead.   

I think we should leave AceObject as is.   Where we can make some beneficial changes might be to look at world object and other derived classes to see if moving them to more appropriate properties backed by AceObject are appropriate.   AceObject should be almost invisible to us.   We really should not have to think about it or interact with it.   If we are having to think about it, we are probably doing something wrong.   Just my 2 pyreal's worth.

LtRipley36706:
@ddevec If I understand what you mean by breaking encapsulation that means for every PropertyX you want to use, you have to put the get/setter in AceObject and then do additional one in whatever subclass of AceObject you want to make use of it?

what's the point of having to specificly put each field in AceObject when the code doesn't need it?

LtRipley36706:
I'm likely just being dense on this point, I don't see the difference in something doing AceObject.Named = x vs AceObject.SetPropertyThing(PropertyThing.Named) even moreso if it was WorldObject.Named which is really just doing WorldObject.Named = AceObject.Named = GetPropertyThing(PropertyThing.Named) in AceObject

you still should and have to know where it is expected to end up in the database because you have to write the get/setter in AceObject.

AceObject will continue to grow quite large with hundreds of get/setters most of which will be specific use cases so largely irrelevant to the majority of objects loaded

that being said, if coders understand that they're putting the get/setters in AceObject then putting an additional set of get/setters in the subclasses or whatever the term is.. i'll move on from this cleanup. I don't agree that we're making things easier to follow, efficient to build out going forward or debug objects

Source/ACE.Entity/AceCharacter.cs
```diff
@@ -20,20 +20,22 @@ public AceCharacter(uint id)
...
             WeenieHeaderFlags = (uint)(WeenieHeaderFlag.ItemsCapacity | WeenieHeaderFlag.ContainersCapacity | WeenieHeaderFlag.Usable | WeenieHeaderFlag.RadarBehavior);
+            WeenieClassId = 1;
+            SetDataIdProperty(PropertyDataId.Icon, 100667446);
+            SetIntProperty(PropertyInt.ItemType, (uint)Enum.ItemType.Creature);
             SetIntProperty(PropertyInt.ItemsCapacity, 102);
             SetIntProperty(PropertyInt.ContainersCapacity, 7);
-            WeenieClassId = 1;
-            Deleted = false;
-            DeleteTime = 0;
+            SetBoolProperty(PropertyBool.IsDeleted, false);
```
ddevec: These should operate on properties directly as opposed to calling the setproperty function.

Source/ACE.Entity/AceCharacter.cs
```diff
@@ -134,15 +136,15 @@ public uint CharacterOptions2Mapping
             set { SetIntProperty(PropertyInt.CharacterOptions2, value); }
         }
 
-        public uint TotalLogins
+        public new uint TotalLogins
```
ddevec: Why the "new" keyword here (and on the other properties)?

Source/ACE.Entity/AceObject.cs
```diff
@@ -1018,25 +1394,7 @@ public void SetStringProperty(PropertyString propertyId, string value)
...
-        }
-
-        public Position Instantiation
-        {
-            get { return GetPosition(PositionType.Instantiation); }
-            set { SetPosition(PositionType.Instantiation, value); }
-        }
-
-        protected Position GetPosition(PositionType positionType)
+        public Position GetPosition(PositionType positionType)
```
ddevec: Why is this (and other get/set property/attribute/position functions) public?  They should be protected, there should be no reason to access them outside of the AceObject (or its subclasses).

Source/ACE/Entity/Door.cs
```diff
@@ -21,30 +24,120 @@ public Door(AceObject aceO)
...
             movementOpen.ForwardCommand = (ushort)MotionCommand.On;
             movementClosed.ForwardCommand = (ushort)MotionCommand.Off;
         }
 
-        public override void OnUse(ObjectGuid playerId)
+        public bool Ethereal
         {
-            // TODO: check if door is locked, send locked soundfx if locked and fail to open.
-            float autoCloseTime = 30.0f;
+            get { return AceObject.GetBoolProperty(PropertyBool.Ethereal) ?? false; }
```
ddevec: This should be encapsulated by AceObject, or one of its subclasses, not here.

Source/ACE/Factories/GeneratorFactory.cs
```diff
@@ -65,34 +66,34 @@ public static List<WorldObject> CreateWorldObjectsFromGenerator(AceObject genera
...
                     // Generate a random position inside the landblock
                     case (int)GeneratorType.Relative:
-                        pos = GetRandomLocInLandblock(random, generator.Location.Cell);
+                        pos = GetRandomLocInLandblock(random, generator.GetPosition(PositionType.Location).Cell);
                         break;
                 }
 
                 // If this generator has linked generators use those for spawning objects
-                if (generator.ActivationCreateClass == 0)
+                if (generator.GetIntProperty(PropertyInt.ActivationCreateClass) == 0)
```
ddevec: You shouldn't call GetIntProperty directly, this breaks encapsulation.

Source/ACE/Factories/GenericObjectFactory.cs
```diff
@@ -14,15 +15,15 @@ public static List<WorldObject> CreateWorldObjects(List<AceObject> sourceObjects
...
-                var oType = (ItemType)aceO.ItemType;
+                var oType = (ItemType)aceO.GetIntProperty(PropertyInt.ItemType);
                 var oDescFlag = (ObjectDescriptionFlag)aceO.AceObjectDescriptionFlags;
 
-                if (aceO.GeneratorStatus)  // Generator
+                if (aceO.GetBoolProperty(PropertyBool.GeneratorStatus) ?? false)  // Generator
                 {
-                    aceO.Location = aceO.Location.InFrontOf(-2.0);
-                    aceO.Location.PositionZ = aceO.Location.PositionZ - 0.5f;
+                    aceO.SetPosition(PositionType.Location, aceO.GetPosition(PositionType.Location).InFrontOf(-2.0));
```
ddevec: Get/setproperty/position again.

Source/ACE.Entity/ObjectGuid.cs
```diff
@@ -5,14 +5,71 @@ namespace ACE.Entity
...
+        Item
     }
 
     public struct ObjectGuid
     {
         public static readonly ObjectGuid Invalid = new ObjectGuid(0);
 
+        public static uint WeenieMin { get; } = 0x00000001;
+        public static uint WeenieMax { get; } = 0x000F423F;
+
```
ddevec: I'm not a huge fan of all of these ranges being tight.  The point of reading out of the DB is that we define a set of "acceptable" ranges, and then figure out what exists in the DB.  Its okay if the DB changes in the future, as long as everything is within acceptable ranges.   Tight ranges mean any change to the DB require changing this code.  I'm not super opposed to it, but I dislike having to change this every time we change the weenies in the DB.

Source/ACE/Entity/Door.cs
```diff
@@ -104,22 +105,25 @@ public float ResetInterval
             private set;
         }
 
+        private double? resetTimestamp;
         public double? ResetTimestamp
         {
-            get;
-            private set;
+            get { return resetTimestamp; }
```
ddevec: Why are these values public?  Reading them is probably racy, so its dangerous making them public without explicit purpose.

Source/ACE/Entity/Door.cs
```diff
@@ -242,35 +246,42 @@ public override void OnUse(ObjectGuid playerId)
...
         private void Open(ObjectGuid opener = new ObjectGuid())
         {
-            CurrentLandblock.EnqueueBroadcastMotion(this, motionOpen);
-            CurrentMotionState = motionStateOpen;
-            PhysicsState |= PhysicsState.Ethereal;
-            Ethereal = true;
-            IsOpen = true;
-            if (opener.Full > 0)
-                UseTimestamp++;
+            ActionChain chain = new ActionChain();
```
ddevec: This is closer -- but your still checking IsLocked and IsOpen in the Player's scope, not the doors (in OnUse).

Source/ACE/Managers/GuidManager.cs
```diff
@@ -127,22 +127,26 @@ public uint Alloc()
         private static uint staticObject = staticObjectMax;
         */
 
+        public static uint GeneratorMin { get; } = ObjectGuid.GeneratorMin;
```
ddevec: If we're moving these to ObjectGuid (which I'm okay with), then I would remove the constants from here, and just use the ObjectGuid constants instead.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -293,7 +293,7 @@ public uint GetNextCharacterId()
         {
             uint maxId = ExecuteConstructedGetAggregateStatement<ShardPreparedStatement, CachedCharacter, uint>(ShardPreparedStatement.GetNextCharacterId);
 
-            ObjectGuid nextGuid = new ObjectGuid(maxId + 1, GuidType.Player);
+            ObjectGuid nextGuid = new ObjectGuid(maxId + 1);
```
ddevec: Since you changed to using the GuidManager for PlayerId's this function should be deleted entirely.

Source/ACE/Entity/Door.cs
LtRipley36706: fields are public so that you can read them for debugging in game via assessment.

Source/ACE.Entity/ObjectGuid.cs
LtRipley36706: the weenie range is probably reasonably good to stay between 1 and 999,999. Turbine only had at most 60,000 weenies in the entire run.

The generator range I added is altogether very likely way to small. It needs to account for generators added to the database as well as cloned/spawned generator copies as we add camp spawning, etc.. I think at this point, it's fine to just start with a smaller range to start until we get a plan in place for reassignment as something like this is pool usage is going to burn through ids as landscape spawns come and go.

As is now, we only have about 20 generators in our DB and they aren't near finished in design yet

Source/ACE/Entity/Door.cs
```diff
@@ -215,9 +215,6 @@ public override void OnUse(ObjectGuid playerId)
                 }
                 else
                 {
-                    bool isLocked;
-                    bool isOpen;
-
                     ActionChain checkDoorChain = new ActionChain();
                     CurrentLandblock.ChainOnObject(checkDoorChain, Guid, (WorldObject dwo) =>
```
ddevec: You should be able to do an action chain with "this", like you did with moveToDoorChain above, instead of using the CurrentLandblock.ChainOnObject (no need to bring the landblock into this when unneeded).

Otherwise this looks better now.

Source/ACE.Entity/ObjectGuid.cs
ddevec: Ahh, I didn't realize 0xF423F was 999,999.  Would it be possible to choose our ranges based on hex values (e.g. 0x100000) instead of decimal -- mixing decimal and hex here obfuscates the meaning of the code (e.g. I thought it was a tight range, when in reality it was an arbitrary decimal value converted to hex).  If not, can you comment it, instead of having some large, arbitrary magic number there?

LtRipley36706:
AppVeyor and nuget packages seem to be having issues..

LtRipley36706:
Changed the fields in Door to private.. not thrilled about being unable to read them at all in game for debugging purposes at present.

Source/ACE/Entity/Door.cs
```diff
@@ -320,15 +291,20 @@ private void Close(ObjectGuid closer = new ObjectGuid())
 
         private void Reset()
         {
-            if ((Time.GetTimestamp() - UseTimestamp) < ResetInterval)
-                return;
+            ActionChain chain = new ActionChain();
```
ddevec: Reset is already called from an ActionChain on "this", this action chain is redundant.

Source/ACE/Entity/Door.cs
```diff
@@ -283,7 +254,7 @@ public override void OnUse(ObjectGuid playerId)
         private void Open(ObjectGuid opener = new ObjectGuid())
         {
             ActionChain chain = new ActionChain();
-            CurrentLandblock.ChainOnObject(chain, Guid, (WorldObject wo) =>
+            chain.AddAction(this, () =>
```
ddevec: Open is now called exclusively from an ActionChain on the Door, this chain is redundant.

Source/ACE/Entity/Door.cs
```diff
@@ -302,7 +273,7 @@ private void Open(ObjectGuid opener = new ObjectGuid())
         private void Close(ObjectGuid closer = new ObjectGuid())
         {
             ActionChain chain = new ActionChain();
-            CurrentLandblock.ChainOnObject(chain, Guid, (WorldObject wo) =>
+            chain.AddAction(this, () =>
```
ddevec: Close is already called from an ActionChain on the door.  This chain is redundant.

PR #466, opened 2017-07-15
Fight Club - Initial working code - ready for review and merge. 

ogmage78:
* Initial work on combat mode.   System now will move into and out of the correct combat mode depending on what items you have equipped.
* TODO: There are some edge cases I still need to implement and additional testing.   I would like to get this in so that I do not run into huge merge issues later on.
* Testing - you can log in - run @weapons after log in.   This will get you various weapons based on weapon type.   You can go into and out of combat mode
* Melee, missile - both bow and xbow, thrown, sword, UA, two handed as well as dual wield.    I also implemented if you remove your weapon while in combat mode, 
* you will drop into unarmed.   If you equip an item and you are not in peace mode, you will assume the correct combat mode.   There may be some tweaking needed 
* for changing combat mode in combat mode - I think it did some slower animation - but the basics are in and I can tweak the sequence once I talk to players or If
* I can find a pcap with that behavior.


Source/ACE/Entity/Creature.cs
```diff
@@ -251,115 +251,227 @@ protected virtual void VitalTickInternal(CreatureVital vital)
...
             }
+        }
 
-            CombatMode = newCombatMode;
-            switch (CombatMode)
+        public void HandleSwitchToPeaceMode(CombatMode oldCombatMode, bool isAutonomous = false)
+        {
+            HandleUnloadMissileAmmo(oldCombatMode);
+
+            // TODO: <this is a hack for now to be removed.> Placement has an issue we have not figured out.   It has to do with animation frame. Og II
```
ddevec: FIXME instead of todo?

Source/ACE/Entity/Creature.cs
```diff
@@ -251,115 +251,227 @@ protected virtual void VitalTickInternal(CreatureVital vital)
...
+                    CurrentLandblock.EnqueueBroadcast(Location, Landblock.MaxObjectRange, new GameMessageUpdatePosition(this));
+                    SetMotionState(this, mm);
+                    mm.MovementData.ForwardCommand = (ushort)MotionCommand.Reload;
+                    SetMotionState(this, mm);
+                    CurrentLandblock.EnqueueBroadcast(Location, Landblock.MaxObjectRange, new GameMessageUpdatePosition(this));
+                    combatModeChain.AddDelaySeconds(0.25);
+                    // System.Threading.Thread.Sleep(250); // used for debugging
+                    mm.MovementData.ForwardCommand = (ushort)MotionCommand.Invalid;
+                    SetMotionState(this, mm);
+                    combatModeChain.AddDelaySeconds(0.40);
```
ddevec: This fix probably shouldn't be part of this patch, but I really don't like these random delays -- we should put in a stub function for waiting for animation changes (which will eventually be read from our dat files)...

Source/ACE/Entity/Creature.cs
ogmage78: I did not have that token defined.   I added it and changed it.   Updated.

Source/ACE/Entity/Creature.cs
ogmage78: I put in FIXME's on both of these until we can get the delay times.   Without these delays, it does not function correctly.   However, just FYI these are about the correct delays - it will be interesting to see what the actual delays are.

Source/ACE/Entity/Creature.cs
```diff
@@ -251,115 +251,227 @@ protected virtual void VitalTickInternal(CreatureVital vital)
...
+            else
+                log.InfoFormat("Changing combat mode for {0} - wielded item {1} has not be assigned a default combat style", Guid, mEquipedMelee?.Guid ?? mEquipedTwoHanded?.Guid);
+        }
+
+        public void SetCombatMode(CombatMode newCombatMode)
+        {
+            log.InfoFormat("Changing combat mode for {0} to {1}", Guid, newCombatMode);
+
+            CombatMode oldCombatMode = CombatMode;
+            CombatMode = newCombatMode;
```
ogmage78: Missed that completely.   I have moved this inside the action chain.   Fixed.

PR #467, opened 2017-07-16
Added npcLooksLikeObject option in appraisal response

OptimShi:
* Added npcLooksLikeObject option in appraisal response (Note that no objects in ACE_World are currently set with this bool property)
* Replaced some "unknown" bitfields with proper variable names in PlayerDescription.

### DIFF
--- a/Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
+++ b/Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
@@ -76,7 +76,7 @@ namespace ACE.Network.GameEvent.Events
             }
 
             var propertiesString = obj.PropertiesString.Where(x => x.PropertyId < 9000).ToList();
-
+/*
 #if DEBUG
             // TODO: This needs to come out - only in while we are testing.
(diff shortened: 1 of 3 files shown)


Source/ACE/Network/GameEvent/Events/GameEventIdentifyObjectResponse.cs
```diff
@@ -144,7 +144,9 @@ public GameEventIdentifyObjectResponse(Session session, WorldObject obj)
                 flags |= IdentifyResponseFlags.StringStatsTable;
             }
 
-            if (obj.ItemType == ItemType.Creature)
+            // if PropertyPool.NpcLooksLikeObject == true, do not treat as creature data
+            var npcLooksLikeObject = propertiesBool.FirstOrDefault(x => x.PropertyId == (uint)PropertyBool.NpcLooksLikeObject)?.PropertyValue;
```
ddevec: Could you make this a property of WorldObject/AceObject, instead of reading it directly from propertiesBool?  In general we don't want to mess with properties outside of those classes (with rare exceptions -- like when we have to send them to the client).

PR #468, opened 2017-07-20
WIP: WorldObject Reorganization + more

LtRipley36706:
* Reorganized folder structure for WorldObject and sub objects.
* Move chat to be handled by Player.
* Set up local chat to be landblock ranged.
* Wired up emotes and soul emotes, also landblock ranged.
* Moved ID functionality to be handled within WorldObject.
* Added PK Altar object.

@OptimShi There's some issues with the SetupModel and PK Altar objects. Would like your input on what is going on.. I swapped in a lifestone ID to let me progress through the code to test further but if you bring the real id back in, you'll find it crashes on reading.


Mogwai-TheFurry:
Also, in general, this feels like it should have been multiple PRs.  You're jamming a lot of work into 1 PR rather unnecessarily, IMO.

PR #469, opened 2017-07-20
Fixes to raising attributes

Lidefeath:
Primary and secondary attributes can be raised correctly with xp again. Existing characters will have coord and quick exchanged as the enums for those abilities in the server are now aligned with the client.
Vital ticking wasn't touched yet, so it only starts during Player.Load

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,12 @@
 # ACEmulator Change Log
 
+### 2017-07-20
+[Lidefeath]
+* Fix raising primary and secondary attributes.
+* Aligned the enums PropertyAttribute.Quickness and PropertyAttribute.Coordination with the client.
+* Due to this alignment old chars will have wrong coord and quick, so make new characters.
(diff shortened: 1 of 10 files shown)


changelog.md
```diff
@@ -1,5 +1,12 @@
 # ACEmulator Change Log
 
+### 2017-07-20
+[Lidefeath]
+* Fix raising primary and secondary attributes.
+* Aligned the enums PropertyAttribute.Quickness and PropertyAttribute.Coordination with the client.
+* Due to this alignment old chars will have wrong coord and quick, so make new characters.
+* I didn't touch the trigger of VitalTicks yet, so vital ticking is only enabled during Player.Load().
```
ddevec: Nice work.  Just an FYI for when you (or whoever) does look at vital ticking on attribute changes.  The invariant on starting a vital tick is that the vital was at max, and after the change it is not (you can't start ticking on /any/ change, this will cause the vital to tick at too high a rate).

PR #470, opened 2017-07-21
Fix for SetupModel Crash

LtRipley36706:


OptimShi:
I approve this. (Except I can't actually "approve" it.) Confirmed it works and fixes the issue we were running into.

PR #471, opened 2017-07-21
Broadcast Local Chat, Emotes and Soul Emotes

LtRipley36706:


Source/ACE/Entity/Player.cs
```diff
@@ -2523,5 +2524,79 @@ protected override void UpdateVitalInternal(CreatureVital vital, uint newVal)
...
+        public void HandleActionTalk(string message)
+        {
+            ActionChain chain = new ActionChain();
+            chain.AddAction(this, () => DoTalk(message));
+            chain.EnqueueChain();
+        }
+
+        public void DoTalk(string message)
+        {
+            if (message.StartsWith("@"))
```
ddevec: As a style note, I would rather the player agnostic message parsing go in the Message Handler, than in the Player's do talk functionality (is the parsing part really part of "DoTalk"?).  I'm not offended by this implementation though

PR #472, opened 2017-07-21
Changed Workmanship

LtRipley36706:
Changed the way Workmanship is set to not always be 0.


LtRipley36706:
Without this Workmanship is always set to 0 for all objects.. Workmanship should only be set if ItemWorkmanship is set to something and then if ItemWorkmanship AND structure are set, some math is done to get a combined number and that is sent as Workmanship instead

ddevec:
I still don't get how returning 0 for Workmanship (I believe 0 is an invalid workmanship), is problematic and needs to be changed.  (NOTE: this does not necessarily cause 0 to be stored in the database, only the in-game representation.)  

This change necessarily complicates any code paths that need to use workmanship (as a null value must now be checked), so the change should only be included if there is a problem it is solving.

LtRipley36706:
It should not be set to 0 by default. we we're not returning 0 for anything else like this, why is this an exception?

ddevec:
We used to default several things to 0 at different layers.  I didn't realize that had been removed.  If this is making Workmanship work consistently with all of the other properties, I'm fine with it.

PR #473, opened 2017-07-21
Moved Identify to WorldObject

LtRipley36706:


Source/ACE/Entity/WorldObject.cs
```diff
@@ -694,6 +695,433 @@ public void Examine(Session examiner)
...
+            WriteIdentifyObjectBoolProperties(writer, flags, propertiesBool);
+            WriteIdentifyObjectDoubleProperties(writer, flags, propertiesDouble);
+            WriteIdentifyObjectStringsProperties(writer, flags, propertiesString);
+            WriteIdentifyObjectDidProperties(writer, flags, propertiesDid);
+            WriteIdentifyObjectSpellIdProperties(writer, flags, propertiesSpellId);
+            WriteIdentifyObjectArmorProfile(writer, flags, propertiesArmor);
+            // TODO: There are probably other checks that need to be made here
+            if (ItemType == ItemType.Creature && GetType().Name != "DebugObject")
+            {
+                WriteIdentifyObjectCreatureProfile(writer, (Creature)this);
```
ddevec: This should be handled in the "Creature" class via virtual method override, not here, as it is a creature exclusive profile.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -694,6 +695,433 @@ public void Examine(Session examiner)
...
+            // Ok Down to business - let's identify all of this stuff.
+            WriteIdentifyObjectHeader(writer, flags, successfulId);
+            WriteIdentifyObjectIntProperties(writer, flags, propertiesInt);
+            WriteIdentifyObjectInt64Properties(writer, flags, propertiesInt64);
+            WriteIdentifyObjectBoolProperties(writer, flags, propertiesBool);
+            WriteIdentifyObjectDoubleProperties(writer, flags, propertiesDouble);
+            WriteIdentifyObjectStringsProperties(writer, flags, propertiesString);
+            WriteIdentifyObjectDidProperties(writer, flags, propertiesDid);
+            WriteIdentifyObjectSpellIdProperties(writer, flags, propertiesSpellId);
+            WriteIdentifyObjectArmorProfile(writer, flags, propertiesArmor);
```
ddevec: Eventually this should probably be moved to an "Armor" subclass

Source/ACE/Entity/WorldObject.cs
```diff
@@ -694,6 +695,433 @@ public void Examine(Session examiner)
...
+            // TODO: There are probably other checks that need to be made here
+            if (ItemType == ItemType.Creature && GetType().Name != "DebugObject")
+            {
+                WriteIdentifyObjectCreatureProfile(writer, (Creature)this);
+            }
+            WriteIdentifyObjectWeaponsProfile(writer, flags, propertiesWeaponsD, propertiesWeaponsI);
```
ddevec: Similar to "Armor", this will need to go into a "Weapon" subclass.

LtRipley36706:
@ddevec This closer to what you're expecting?

It will break most of the creature assessment panels as 99% of the objects in the world are not Creatures yet.

Source/ACE/Entity/Creature.cs
```diff
@@ -495,5 +497,162 @@ protected virtual void UpdateVitalInternal(CreatureVital vital, uint newVal)
                 new ActionChain(this, () => VitalTickInternal(vital)).EnqueueChain();
             }
         }
+
+        public override void SerializeIdentifyObjectResponse(BinaryWriter writer)
```
ddevec: This has a signficiant amount of duplicated code with WorldObject's SerializeIdentifyObjectResponse.  Consider calling "base.SerializeIdentifyObjectResponse", or if that is inappropriate, refactoring such that you deduplicate the code between the two calls.

Source/ACE/Entity/Creature.cs
```diff
@@ -495,5 +497,162 @@ protected virtual void UpdateVitalInternal(CreatureVital vital, uint newVal)
...
+            var propertiesArmor = PropertiesDouble.Where(x => (x.PropertyId < 9000
+                                                         && (x.PropertyId >= (uint)PropertyDouble.ArmorModVsSlash
+                                                         && x.PropertyId <= (uint)PropertyDouble.ArmorModVsElectric))
+                                                         || x.PropertyId == (uint)PropertyDouble.ArmorModVsNether).ToList();
+            if (propertiesArmor.Count > 0)
+            {
+                flags |= IdentifyResponseFlags.ArmorProfile;
+            }
+
+            // Weapons Profile
```
ddevec: Why do creatures deal with weapons and armor profiles?

Source/ACE/Entity/Creature.cs
LtRipley36706: I did no refactoring, just trying to see if this is generally what you were saying

ddevec: I think this is generally right then.  

If breaking creature identification for non-creatures DebugObjects is a problem, we may consider not doing this refactor yet.  

I would argue that there is little point returning a proper identification for an improperly implemented underlying type, but I'm not working with general WorldObject properties or their uses, and I have a limited view in this realm.

LtRipley36706: to be fair, most of the stuff returned in the creatureprofile displayed in the client are bad numbers/stats, so i'm kinda okay with breaking it in one regard.. 

i've got some thoughts on how to implement weenietype and delete debugobject going forward but it'll take some db work + thinking/coding time to propose properly, possibly looking to do that in a week or so

LtRipley36706: in short, the way we key object creation is bad and unsustainable. ObjectDescriptionFlags is more changeable than we originally suspected.

PR #476, opened 2017-07-26
WorldObject reclassification via WeenieType

LtRipley36706:
- NOTE: The following changes require ACE-World database v0.1.6 or greater...
- No longer setting WeenieHeaderFlag.Value if Value = 0.
- DebugObject, CollidableObject and UsableObject are removed.
- Moved OnUse/OnCollide to WorldObject virtuals for overriding.
- Added WeenieType to AceObject/WorldObject for weenie classification.
- Added Generic object which the default object for weenies.
- Reworked GenericObjectFactory to use WorldObject.GetObjectFromWeenieType.


LtRipley36706:
Check discord if you'd like a sample of the new data for testing, I'll be working on a new export for the next day or so for ACE-World

LtRipley36706:
Think this is good to go for now

Source/ACE/Entity/GenericObject.cs
```diff
@@ -0,0 +1,64 @@
...
+                    var sendUseDoneEvent = new GameEventUseDone(player.Session);
+
+                    player.Session.Network.EnqueueSend(errorMessage, sendUseDoneEvent);
+                });
+                chain.EnqueueChain();
+            }
+        }
+
+        public override void InternalOnUse(Session session)
+        {
```
ddevec: Its not clear what the difference between OnUse and InternalOnUse are, and why they are two separate functions (one doesn't call the other here either...).  Additionally I see the "Internal" function being called from other classes.  This is a very confusing convention...

Source/ACE/Entity/Player.cs
```diff
@@ -2345,27 +2345,27 @@ public void HandleActionUse(ObjectGuid usedItemId)
         {
             new ActionChain(this, () =>
             {
-                if (CurrentLandblock != null)
+                if (GetInventoryItem(usedItemId) != null)
                 {
-                    // Just forward our action to the appropriate user...
-                    ActionChain onUseChain = new ActionChain();
-                    CurrentLandblock.ChainOnObject(onUseChain, usedItemId, (WorldObject wo) =>
+                    WorldObject iwo = GetInventoryItem(usedItemId);
```
ddevec: Nit: GetInventoryItem is called 2x back to back.  This should be consolidated to a single call.

Source/ACE/Entity/Player.cs
```diff
@@ -2345,27 +2345,27 @@ public void HandleActionUse(ObjectGuid usedItemId)
...
+                    ////iwo = iwo.GetObjectFromWeenieType(); // is this right? is this even needed?
+                    iwo.InternalOnUse(Session);
```
ddevec: Why is Player calling InternalOnUse?  The player is clearly not internal to the item.

Source/ACE/Entity/Player.cs
```diff
@@ -2345,27 +2345,27 @@ public void HandleActionUse(ObjectGuid usedItemId)
...
-                        if (uo != null)
+                        // Just forward our action to the appropriate user...
+                        ActionChain onUseChain = new ActionChain();
+                        CurrentLandblock.ChainOnObject(onUseChain, usedItemId, (WorldObject wo) =>
                         {
-                            uo.OnUse(Guid);
-                        }
-                    });
-                    onUseChain.EnqueueChain();
+                            if (wo != null)
```
ddevec: I don't believe this null check is needed -- the landblock will just drop the chain if the object isn't found (we actually may need to give the landblock some way of nacking the chainer of failure in the future).

Source/ACE/Entity/WorldObject.cs
```diff
@@ -80,59 +82,184 @@ public List<AceObjectPropertiesSpell> PropertiesSpellId
...
 
-        public Position ForcedLocation { get; private set; }
+        public void AddModel(byte index, uint modelresourceid)
+        {
+            Model newmodel = new Model(index, modelresourceid);
+            models.Add(newmodel);
+        }
 
-        public Position RequestedLocation { get; private set; }
+        public void Clear()
```
ddevec: Clear the entire world object?  This should have a more clear name.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -1963,7 +1991,7 @@ private WeenieHeaderFlag SetWeenieHeaderFlag()
             if (AmmoType != null)
                 weenieHeaderFlag |= WeenieHeaderFlag.AmmoType;
 
-            if (Value != null)
+            if (Value != null && (Value > 0))
```
ddevec: Is it correct behavior that it is impossible to create an item with Value 0, or does the client interpret no sent value as Value 0?

Source/ACE/Entity/WorldObject.cs
```diff
@@ -2411,35 +2440,39 @@ private PhysicsDescriptionFlag SetPhysicsDescriptionFlag()
...
                 if (CurrentMotionState != null)
                 {
-                    var movementData = CurrentMotionState.GetPayload(wo.Guid, wo.Sequences);
-                    writer.Write(movementData.Length); // number of bytes in movement object
-                    writer.Write(movementData);
-                    uint autonomous = CurrentMotionState.IsAutonomous ? (ushort)1 : (ushort)0;
-                    writer.Write(autonomous);
+                    var movementData = CurrentMotionState.GetPayload(Guid, Sequences);
+                    if (movementData.Length > 0)
+                    {
```
ddevec: Look at how the client handles this (the pastebin I posted the other day), and verify this is right.  I'm not sure the movementData is being sent correctly.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -2695,5 +2726,50 @@ private void RecallAndSetPhysicsStateBools()
             if (AceObject.IsFrozen ?? false)
                 Frozen = true;
         }
+
+        public WorldObject GetObjectFromWeenieType()
```
ddevec: I don't understand this at all.  This creates a new object with the same weenie type and AceObject as the current object?  Why is this appropriate, and where is it used?

Source/ACE/Factories/GenericObjectFactory.cs
```diff
@@ -29,37 +26,14 @@ public static List<WorldObject> CreateWorldObjects(List<AceObject> sourceObjects
...
-                    default:
-                        // Use the DebugObject to assist in building proper objects for weenies
-                        // FIXME(ddevec): Some objects have null location.  This freaks out the landblock... ignore them?
-                        if (aceO.Location != null)
-                        {
-                            results.Add(new DebugObject(aceO));
-                        }
-                        break;
-#endif
+                    results.Add(wo.GetObjectFromWeenieType());
```
ddevec: I don't like how this works.  Why must a generic object be created, then another object, which shares its data be created?  This smells wrong.

Source/ACE/Entity/GenericObject.cs
LtRipley36706: Could be bad naming, how about OnUse and OnUseContained or ContainedOnUse?

The difference being OnUse is called upon something in the landblock, the other is called when the player already has the object

Source/ACE/Entity/Player.cs
LtRipley36706: see previous comment

Source/ACE/Entity/WorldObject.cs
LtRipley36706: Didn't add this, only moved it

Source/ACE/Entity/WorldObject.cs
LtRipley36706: The example of this would be doors. Many doors have a value of 0 but they are not sent with weenieflags.

You can have a value of 0, it is never sent as part of the weenie flag, only upon ID

Source/ACE/Entity/GenericObject.cs
ddevec: Perhaps "HandleActionOnUse" -- as that is the standard naming to indicate an action chain is required internally -- and an "OnUseInternal" would be acceptable?

Source/ACE/Entity/Player.cs
ddevec: How about commenting this line, then.  It has a subtle behavior which is not immediately documented.  If I know how the action chains work well, and couldn't immediately see its purpose, it could probably use a comment.

Source/ACE/Entity/WorldObject.cs
ddevec: Either fix, or comment what it does (to be later fixed then), its a deceptive name.

Source/ACE/Entity/WorldObject.cs
LtRipley36706: see next comment

Source/ACE/Entity/WorldObject.cs
ddevec: As long as we never need to send a value of 0 with the weenieheaderflag, then this is fine.

Source/ACE/Entity/WorldObject.cs
LtRipley36706: let's address that in another PR

ddevec: This can wait until another PR.  I only commented on it because you included the animation_id fix in this PR.  This is of the same nature.

Source/ACE/Factories/GenericObjectFactory.cs
LtRipley36706: `var objects = DatabaseManager.World.GetObjectsByLandblock(this.id.Landblock);
            var factoryObjects = GenericObjectFactory.CreateWorldObjects(objects);
            factoryObjects.ForEach(fo =>
            {
                if (!worldObjects.ContainsKey(fo.Guid))
                {
                    worldObjects.Add(fo.Guid, fo);
                    fo.SetParent(this);
                }
            });`

GenericObjectFactory.CreateWorldObjects is called from the code above in Landblock.cs

it wants to spit out a list of WorldObjects. We need to create the objects based on their WeenieType but the function for creating shouldn't be hidden away or need to be duplicated as I believe its going to be used several times so I put it in WorldObject as GetObjectFromWeenieType. 

I don't currently see a way to get from AceObject to final real object (Lifestone, Door, etc) at this point without going through WorldObject and since that's an abstract class, I have to create something derived from it first and then use that object to create the correct weenie object.

I'm sure I'm going about this in a more convoluted method than needed, but I don't see the smarter solution presently, if you have a better way to get from AceObject to Lifestone and be able to reuse that same weenie categorizer function with generators, or spawning items, etc, we're gonna be set.

LtRipley36706: Maybe I can play around with GenericObjectFactory itself and make that thing more useful there.. lemme look into that

ddevec: Make GetObjectFromWeenieType a static method of one of the generator classes, have it take in an AceObject and a WeenieType.

PR #477, opened 2017-07-26
Query Item Mana

Lidefeath:
Add the GameAction for QueryItemMana and the corresponding GameEvent response.


Source/ACE/Entity/WorldObject.cs
```diff
@@ -1877,6 +1877,23 @@ public void QueryHealth(Session examiner)
             examiner.Network.EnqueueSend(updateHealth);
         }
 
+        public void QueryItemMana(Session examiner)
+        {
+            float manaPercentage = 1f;
+            uint success = 0;
+
+            var currentMana = PropertiesInt.Where(x => x.PropertyId == (uint)PropertyInt.ItemCurMana).FirstOrDefault();
```
ddevec: Please don't use "PropertiesInt" directly.  Instead wrap the lookup in a field in the AceObject holding the data (see fields like Burden on how to do this).  PropertiesInt is only exposed for when we're sending bulk properties to the client (and perhaps we shouldn't expose it this way to discourage people from using it).

PR #481, opened 2017-07-30
Expecto Patronum - Spell bar managment

ogmage78:
### 2017-07-30
[Og II]
* Added spell bar management and persistence.   This continues the work for learning and persisting spells.   
* Cleaned up some of the duplication of enumerations in Game Event Player Description
* Added some debug code around universal motion.
* Added some debug asserts in Player Description 
* To test - use @learnspell to get some spells in your spell book if you do not have any.   They will persist with your character. 
* You can also unlearn them and they will be removed.
* Now, drag your spells into any configuration in your spell bars you like.   You can add them, remove them or reorder them.
* They will be saved on our periodic save, or when you log out or you can force a save with the debug command @save-now.
* Added table to allow for spell bar configuration persistence.
* Cleaned up Enumerations - Game Action Types
* That's all for this PR.


Source/ACE/Entity/Player.cs
```diff
@@ -194,6 +194,43 @@ public void HandleActionLearnSpell(uint spellId)
...
+        /// <summary>
+        /// This method implements player spell bar management for - adding a spell to a specific spell bar (0 based) at a specific slot (0 based).
+        /// </summary>
+        /// <param name="spellId"></param>
+        /// <param name="spellBarPositionId"></param>
+        /// <param name="spellBarId"></param>
+        public void HandleActionAddSpellToSpellBar(uint spellId, uint spellBarPositionId, uint spellBarId)
+        {
+            // The spell bar magic happens here. First, let's mind our race conditions....
+            // No need for action chain here - internal to us, no other object can change us.
```
ddevec: This can be read and written at the same time (You can learn a spell while deleting/casting/moving the spellbar).  Even though its internal to us, its concurrent.  In general don't bother trying to elide ActionChains, your more likely to introduce bugs that are difficult to debug later on.

Source/ACE/Entity/Player.cs
```diff
@@ -194,6 +194,43 @@ public void HandleActionLearnSpell(uint spellId)
...
+
+        /// <summary>
+        /// This method implements player spell bar management for - removing a spell to a specific spell bar (0 based)
+        /// </summary>
+        /// <param name="spellId"></param>
+        /// <param name="spellBarId"></param>
+        public void HandleActionRemoveSpellToSpellBar(uint spellId, uint spellBarId)
+        {
+            // More spell bar magic happens here. First, let's mind our race conditions....
+            // No need for action chain here - internal to us, no other object can change us.
```
ddevec: Same as above -- please put in actionchian.

Source/ACE/Entity/WorldObject.cs
```diff
@@ -2464,16 +2464,21 @@ public void SerializePhysicsData(BinaryWriter writer)
...
                 else
-                {                  
+                {
+                    log.Debug($"Our flag is set but our current motion state is null. {this.Guid.Full:X}");
                     writer.Write(0u);
                 }
             }
             else if ((PhysicsDescriptionFlag & PhysicsDescriptionFlag.AnimationFrame) != 0)
                 writer.Write((AnimationFrame ?? 0u));
+            // TODO: Keep an eye on this, are we sure the client does not just ignore it?   I would think they way it reads by buffer length that this would blow up.
```
ddevec: the else if is how the client parses the code.  Its funky.  The above movement data writing code is wrong.  I'll try to put a patch in for it this week.

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
```diff
@@ -329,7 +304,12 @@ private void WriteEventBody()
 
             // FIXME(ddevec): We have duplicated data everywhere.  There is an AceObject CharacterOption flag, and a Player.CharacterOptions...
             //    Which one is right?  I have no idea.  Right now the aceObject works...  we should probably do a refactoring once we've restored functionality
-            var optionFlags = DescriptionOptionFlag.CharacterOption2;
+
+            // TODO: Refactor this to set all of these flags based on data. Og II
+            CharacterOptionDataFlag optionFlags = CharacterOptionDataFlag.CharacterOptions2;
+            if (aceObj.SpellsInSpellBars.Exists(x => x.AceObjectId == aceObj.AceObjectId))
```
ddevec: Another accessor instead of direct access case.

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
```diff
@@ -339,28 +319,33 @@ private void WriteEventBody()
...
-                writer.Write(0u);
-                writer.Write(0u);
-                writer.Write(0u);
-                writer.Write(0u);
-                writer.Write(0u);
-            }*/
+                for (int i = 0; i <= 7; i++)
+                {
+                    var sorted =
+                        aceObj.SpellsInSpellBars.FindAll(x => x.AceObjectId == aceObj.AceObjectId && x.SpellBarId == i)
```
ddevec: Same as above. (with accessor)

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -313,7 +303,7 @@ public AceObject GetObject(uint aceObjectId)
 
         private void LoadIntoObject(AceObject aceObject)
         {
-            // this flag determines whether subequent calls to "save" trigger an insert or an update
+            // this flag determines when the subsequentnt calls to "save" trigger an insert or an update
```
ddevec: Grammar Nazi!  :-P

Source/ACE/Entity/Player.cs
```diff
@@ -194,6 +194,43 @@ public void HandleActionLearnSpell(uint spellId)
...
+            AceObject.SpellsInSpellBars.Remove(AceObject.SpellsInSpellBars.Single(x => x.SpellBarId == spellBarId && x.SpellId == spellId));
+            // Now I have to reorder
+            var sorted = AceObject.SpellsInSpellBars.FindAll(x => x.AceObjectId == AceObject.AceObjectId && x.SpellBarId == spellBarId).OrderBy(s => s.SpellBarPositionId);
```
ddevec: In general I don't like exposing the properties of AceObject outside of the AceObject (it requires the user of AceObject to know how the data is stored, which shouldn't be exposed to the game-content data).  Can you write an accessor for this (and other accesses of this type)?

Source/ACE/Entity/WorldObject.cs
ogmage78: I did not change it, only put in a debug line

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
ogmage78: This is a special case and I think in this instance should be allowed.   The server never does anything with these.   It is only a persistence function.   It saves them and it reads them to load them.  Everything else happens client side in regards to spell bars.   I can easily add another layer - but it would be for nothing I believe.

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
ogmage78: This is a special case and I think in this instance should be allowed.   The server never does anything with these.   It is only a persistence function.   It saves them and it reads them to load them.  Everything else happens client side in regards to spell bars.   I can easily add another layer - but it would be for nothing I believe.

Source/ACE/Entity/Player.cs
ogmage78: I had the action chains in and took them out.   I will add them back now.

Source/ACE/Entity/Player.cs
ogmage78: This is a special case and I think in this instance should be allowed. The server never does anything with these. It is only a persistence function. It saves them and it reads them to load them. Everything else happens client side in regards to spell bars. I can easily add another layer - but it would be for nothing I believe.

Source/ACE/Entity/WorldObject.cs
ddevec: I'm just commenting on your added todo -- we know whats going on here, and we'll fix it.   You don't need to change this code.

Source/ACE/Entity/Player.cs
ddevec: I'm pretty strongly opinionated that there should be a logical separation between the two components (database/backend, and in-game data).  I would really like to see this abstracted consistently with the remainder of the data that falls into this paradigm.  If your strongly against doing so I wont hold up this PR over it, but the inner me that loves consistency and well defined interfaces will be very sad.

ogmage78:
Added the accessors - thanks again for the code review!

Source/ACE/Entity/Player.cs
ogmage78: I got that added.   Thanks for reviewing the code for us!

PR #482, opened 2017-07-31
Add Cow object

LtRipley36706:
* NOTE: The following changes require ACE-World database v0.1.8 or newer...
* Added NpcLooksLikeObject check for Creature assessment profile fix provided by @OptimShi.
* Moved Default do nothing UseDone to WorldObject and removed it from GenericObject.
* Added Cow weenie object.
* Added Cow, Creature, and Container to the WorldObjectFactory.
* Say Hi to the cow in Holtburg.


ogmage78:
I know we discussed this a while back.   Are we going to do one of these per creature family?   By that I mean all Tuskers for example will be one object not an object for every different tusker?  I think if we do this by weenietype and not weenie - that will allow content creators to add content without having to change server code.

LtRipley36706:
No need. Cows were a unique WeenieType, all other creatures will simply be Creature.cs

ddevec:
Monster.cs?  Monsters have different logical behaviors than creatures (e.g. players don't have scripts they run to attack nearby players).

LtRipley36706:
There's no distinction between Monsters and Creatures in the client, it would seem from turbine's viewpoint , we all sprang from creature weenietype

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -314,6 +328,42 @@ private List<AceObjectPropertiesSpell> GetAceObjectPropertiesSpell(uint aceObjec
             return objects;
         }
 
+        private List<AceObjectPropertiesSkill> GetAceObjectPropertiesSkill(uint aceObjectId)
+        {
+            var criteria = new Dictionary<string, object> { { "aceObjectId", aceObjectId } };
+            var objects = ExecuteConstructedGetListStatement<WorldPreparedStatement, AceObjectPropertiesSkill>(WorldPreparedStatement.GetAceObjectPropertiesSkills, criteria);
+            ////objects.ForEach(o =>
```
Mogwai-TheFurry: no reason to include chunks of commented code in new stuff.

Source/ACE/Entity/Creature.cs
```diff
@@ -89,10 +89,10 @@ public CreatureVital Mana
 
         public double RespawnTime { get; set; }
 
-        public Creature(AceObject aceObject, ObjectGuid guid)
-            : base(aceObject, guid)
-        {
-        }
+        ////public Creature(AceObject aceObject, ObjectGuid guid)
```
Mogwai-TheFurry: Deletes are preferred over comments.  We have history.

Source/ACE/Entity/GenericObject.cs
```diff
@@ -15,46 +15,24 @@ public GenericObject(AceObject aceObject)
         {
         }
 
-        public GenericObject(ObjectGuid guid, AceObject aceObject)
-            : base(guid, aceObject)
-        {
-        }
+        ////public GenericObject(ObjectGuid guid, AceObject aceObject)
```
Mogwai-TheFurry: At this point the class serves no purpose.  It's a commented out class with a constructor.  Just remove it.

LtRipley36706:
```cs

public static WorldObject CreateWorldObject(AceObject aceO)
        {
            WeenieType objWeenieType = (WeenieType?)aceO.WeenieType ?? WeenieType.Generic;

            switch (objWeenieType)
            {
                case WeenieType.LifeStone:
                    return new Lifestone(aceO);
                case WeenieType.Door:
                    return new Door(aceO);
                case WeenieType.Portal:
                    return new Portal(aceO);
                // case WeenieType.PKModifier:
                //    return new PKModifier(aceO);
                case WeenieType.Cow:
                    return new Cow(aceO);
                case WeenieType.Creature:
                    return new Creature(aceO);
                case WeenieType.Container:
                    return new Container(aceO);
                default:
                    return new GenericObject(aceO);
            }
        }

```

Need a default object. I assume this object will eventually have some worth while code in it.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -39,7 +39,10 @@ private enum WorldPreparedStatement
             GetAceObjectPropertiesPosition = 24,
             GetAceObjectPropertiesSpell = 25,
             GetAceObjectGeneratorLinks = 26,
-            GetMaxId = 27
+            GetMaxId = 27,
+            GetAceObjectPropertiesAttributes,
```
ddevec: I don't like the inconsistency here.  Either remove the = # from all of them, or put the = # in your enum values.

Source/ACE/Entity/Cow.cs
```diff
@@ -0,0 +1,126 @@
...
+                float playerDistanceTo = player.Location.SquaredDistanceTo(Location);
+
+                ////if (playerDistanceTo >= 2500)
+                ////{
+                ////    var sendTooFarMsg = new GameEventDisplayStatusMessage(player.Session, StatusMessageType1.Enum_0037);
+                ////    player.Session.Network.EnqueueSend(sendTooFarMsg, sendUseDoneEvent);
+                ////    return;
+                ////}
+
+                if (playerDistanceTo >= radiusSquared)
```
ddevec: Not that you *have* to do it here, but this code appears common to most/all HandleActionOnUse functions.  Perhaps we should come up with some other solution than copying it all over the place?

Source/ACE/Entity/GenericObject.cs
Mogwai-TheFurry: I'd rather it echo out something to the effect of "Was asked to create a {WeenieType}, but don't know how" or something.  GenericObject is useless Object.

LtRipley36706: except, there are many WeenieType.Generic objects in the database.. its not going to be as useless as you are suggesting

LtRipley36706: Also, removing it kills every spawn-able object not explicitly handled by CreateWorldObject which does us no good in the now.

Source/ACE.Database/WorldDatabase.cs
LtRipley36706: Og's PR removes them all, I only added new ones from Shard and did not include numbers

Source/ACE/Entity/Cow.cs
LtRipley36706: a reasonable suggestion, it probably will be in at least a few more objects as well

Source/ACE.Entity/AceObject.cs
```diff
@@ -1816,24 +1816,48 @@ public object Clone(uint guid)
...
+            ret.IntProperties.ForEach(c => c.AceObjectId = guid);
+            ret.Int64Properties.ForEach(c => c.AceObjectId = guid);
+            ret.DoubleProperties.ForEach(c => c.AceObjectId = guid);
+            ret.BoolProperties.ForEach(c => c.AceObjectId = guid);
+            ret.DataIdProperties.ForEach(c => c.AceObjectId = guid);
+            ret.InstanceIdProperties.ForEach(c => c.AceObjectId = guid);
+            ret.StringProperties.ForEach(c => c.AceObjectId = guid);    
+            ret.GeneratorLinks.ForEach(c => c.AceObjectId = guid);
+            
+            // We're not seemingly changing the guids for the Dictionaries below from Clone(), is this bad?
```
ddevec: The dictionaries separate the DB layer data and the in-game data, so the guid doesn't need to be adjusted (it will be crated/adjusted on save).  This is an okay note, but the comments below it are just clutter.

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1009,28 +1009,12 @@ public static void HandleCI(Session session, params string[] parameters)
...
-            else
-            {
-                WorldObject loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
-                loot.ContainerId = session.Player.Guid.Full;
-                session.Player.AddToInventory(loot);
-                session.Player.TrackObject(loot);
-                session.Network.EnqueueSend(new GameMessagePutObjectInContainer(session, session.Player.Guid, loot, 0),
-                    new GameMessageUpdateInstanceId(loot.Guid, session.Player.Guid, PropertyInstanceId.Container));
-            }
+            WorldObject loot = WorldObjectFactory.CreateNewWorldObject(weenieId);
```
ddevec: Just noticed this modifies the player's inventory outside fo an actionchain.  This is a race.  Please throw a FIXME comment in there for me, since your modifying that code anyway (or wrap this chunk of code in an actionchain).

Source/ACE/Entity/Landblock.cs
```diff
@@ -104,10 +104,13 @@ public Landblock(LandblockId id)
             var factoryObjects = WorldObjectFactory.CreateWorldObjects(objects);
             factoryObjects.ForEach(fo =>
             {
-                if (!worldObjects.ContainsKey(fo.Guid))
+                if (fo != null)
```
ddevec: Under what instance will WorldObjectFactory.CreateWorldObjects populate its object list with a null, and why is that not a bug in WorldObjectFactory.CreateWorldObjects?

Source/ACE.Entity/AceObject.cs
ogmage78: @ddevec you and @Mogwai-TheFurry  need to talk.   When we clone a new object from a weenie, we have to do a deep reset of those values.   I don't care where it takes place, but it has to happen prior to save.   I have this in my code.   I made it part of the clone new method.   It happens behind the scenes.

ddevec: In general I dislike the fact that save-time data creeps into our live data-structures, and the smell here is associated with the presence of the redundant data.  I believe either implementing the guid update at clone time (as Ripley has done here), or at save time is sufficient.  Other than the fact I don't like this data existing at all, I think Ripley's implementation here is fine.

It looks like you and Ripley have essentially created the same code to accomplish this, and it seems to be a fine way to go about doing it to me.

Source/ACE/Entity/Landblock.cs
LtRipley36706: technically impossible in the current state of the code. I added the null check when while I was temping in the removal of GenericObject to reconfirm what I already knew. WOF.CreateWorldObjects prevents nulls in results now so the fo null check is just not possible to hit ever, and technically so is the null in CreateWorldObjects as long as we set a default object, which I see no reason no to do.

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: I assume this code moves to Player (eventually Admin) class anyway

Source/ACE.Entity/AceObject.cs
LtRipley36706: I copied his code in per other discussions just to make it clear what the guid code was doing

Source/ACE/Entity/Landblock.cs
ddevec: I think its important to decide if a correct implementation of CreateWorldObjects can or cannot introduce null objects into its list.  As it returns a list, I would imagine cannot is the correct behavior. 

I am a strong opponent of error-oblivious computing (intentionally ignoring errors), it leads to buggy, inconsistent, and difficult to debug code.  Intentionally ignoring values that should be impossible from CreateWorldObjects is a form of error-oblivious computing.

With that said, I would highly recommend removing this check, either replace it with an assertion (Debug.Assert) if your worried about null values, or remove it entirely.

As written, this code indicates CreateWorldObjects can/should return null values in its list, and that's just confusing.

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: likely most of these commands will need actionchains and relocation.. fwiw

PR #484, opened 2017-08-02
Inventory - Ready for Merge

ogmage78:
### 2017-08-01
[Og II]
##  Refactored again on Mogwai's feedback 

This is getting so large, I need to stop here - this is a good spot - adds significant functionality but with a few known bugs.   

* Inventory will save with the container and load again in the correct positions.   
* Character Description Event is now wired up for inventory.
* removed containers copy of aceObject Inventory - just make the list accessible.
* Updated Clone so it you pass a new guid, it resets all the associated child tables.
* implemented placement IntProperty 65 used to track slot in Container
* Created supporting methods to manage pack order
* Enhanced the debug log message for DBDEBUG 
* Small temp fix to stutter when attempting to shift walk or shift jump.   We still need to really understand autonomous
* Created new view  vw_ace_inventory_object to expose ability to pull by container.  
* general clean up comments and whitespace 
* Inventory loading - working
* TODO.   
* 1 Add in capacity checks
* 2 bug still in picking up items off the ground and saving to database. 
* 3 bug with reloaded items saved in a secondary side pack.


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1015,17 +1015,23 @@ public static void HandleCI(Session session, params string[] parameters)
             WorldObject testContainer = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
```
LtRipley36706:
```cs

            WorldObject newObj = WorldObjectFactory.CreateNewWorldObject(weenieId);
            
                newObj.ContainerId = session.Player.Guid.Full;
                newObj.Placement = 0;
                session.Player.AddToInventory(newObj);
                session.Player.TrackObject(newObj);
                session.Network.EnqueueSend(new GameMessagePutObjectInContainer(session, session.Player.Guid, newObj, 0),
                    new GameMessageUpdateInstanceId(newObj.Guid, session.Player.Guid, PropertyInstanceId.Container));
            }
        }

```

try this instead

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -825,8 +826,10 @@ public static void HandleWeapons(Session session, params string[] parameters)
             {
                 foreach (uint weenieId in weaponsTest)
                 {
-                    WorldObject loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
+                    AceObject lootAceObject = (AceObject)DatabaseManager.World.GetAceObjectByWeenie(weenieId).Clone(GuidManager.NewItemGuid().Full);
+                    WorldObject loot = new GenericObject(lootAceObject);
```
LtRipley36706:
```cs
WorldObject loot = WorldObjectFactory.CreateNewWorldObject(weenieId);
```
instead of AceObject + WorldObject stuff

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -849,11 +852,13 @@ public static void HandleSplits(Session session, params string[] parameters)
             {
                 foreach (uint weenieId in splitsTest)
                 {
-                    WorldObject loot = LootGenerationFactory.CreateTestWorldObject(session.Player, weenieId);
+                    AceObject lootAceObject = (AceObject)DatabaseManager.World.GetAceObjectByWeenie(weenieId).Clone(GuidManager.NewItemGuid().Full);
+                    WorldObject loot = new GenericObject(lootAceObject);
```
LtRipley36706:
```cs
WorldObject loot = WorldObjectFactory.CreateNewWorldObject(weenieId);
```
instead of AceObject + WorldObject stuff

Source/ACE/Command/Handlers/DebugCommands.cs
ogmage78: I tried that first - but it dies not look like it really works.   If you try to persist it, you will see that what you are doing behind the scenes, is making a world object which is made from an aceObject.   Then you are giving in a new guid.   The problem is that all of the aceObject contained in your new world object have the aceobjectid of the item you cloned from.   It does not matter as long as you are just spawning something that will never be persisted.   If you actually try to save that to the database, it blows up.   This actually works as coded.

Source/ACE/Command/Handlers/DebugCommands.cs
ogmage78: see above.

Source/ACE/Command/Handlers/AdminCommands.cs
ogmage78: See my comment below.   That has a bug when you try to persist as far as I can tell.   It is also a bit of in indirect to go from aceObject -> worldobject - clone for the reasons listed below.   I think it is better to just use the aceobject of the weenie clone and reset for a new object then worldobject from that new ace object.   Either will work if you fix the issue saving exposed - I think the way I am doing it is cleaner.

Source/ACE/Command/Handlers/DebugCommands.cs
Mogwai-TheFurry: I think the solution here is to make CreateNewWorldObject work.  CreateNewWorldObject is meant to be the primary tool for generating loot.  It currently only takes a weenie, but should probably be modified to accept a Loot Tier (int) as well.

This functionality should be an easy one liner, not a complicated daisy-chain of methods like Og had to implement.

Source/ACE/Entity/Container.cs
```diff
@@ -51,14 +52,54 @@ public Container(AceObject aceObject, ObjectGuid guid)
         {
         }
 
+        public virtual void ContainerPlacement(WorldObject inventoryItem)
+        {
+            var removedItem = inventory.Remove(inventoryItem.Guid);
+            var invList = inventory.OrderBy(o => o.Value.Placement);
+            uint place = 0;
+            AceObject saveableCopy;
+            foreach (var keyValuePair in invList)
```
Mogwai-TheFurry: This loop doesn't make sense to me.  Can you explain what it does?

Source/ACE.Entity/AceObject.cs
```diff
@@ -1798,6 +1808,20 @@ public object Clone(uint guid)
         {
             AceObject ret = (AceObject)Clone();
             ret.AceObjectId = guid;
+            // We are cloning a new AceObject with a new AceObjectID - need to set this to false. Og II
+            ret.HasEverBeenSavedToDatabase = false;
```
ogmage78: will do

Source/ACE/Command/Handlers/DebugCommands.cs
ogmage78: Ok, I will work on making the other method work and refactor my code to use that.

Source/ACE/Entity/Container.cs
ogmage78: When you either put something in your backpack or you just move it around to a new slot - everything in the pack with a placement >= the placement slot you are going in has to move.   Put 4 things in your pack.   They are in slots 0, 1, 2, 3   Now drag item in slot 3 to slot 0.    3 now is set to 0, 0 is set to 1 and 1 is set to 2 and 2 is set to 3.   Not every move ends up with a total inventory shift.   Only drag to the first of the list does that.   If you drag to the middle, only everything past that has to shift.   If you drag to the last open slot, nothing moves.   By default, when you pick something up off the ground if you do not drag it to a slot, it goes to position 0.   There may be a better way to do that.   This works correctly in all cases with the minimum number of moves. (updates).   More than happy to refactor if there is a better way.

ogmage78: it's a mini towers of hanoi ;)    Just kidding that is what it made me think of laying out what was happening on a piece of paper.

Source/ACE/Command/Handlers/DebugCommands.cs
LtRipley36706: [See this change in 482](https://github.com/ACEmulator/ACE/pull/482/commits/50df1daf9d03d82b8b8484c2e775380ad4418ee0)

Think it does what you're expecting, assuming your clone(guid) changes are also merged

Source/ACE/Entity/Container.cs
```diff
@@ -51,14 +52,54 @@ public Container(AceObject aceObject, ObjectGuid guid)
         {
         }
 
+        public virtual void ContainerPlacement(WorldObject inventoryItem)
```
ddevec: I'm not a really big fan of this function name.  What does ContainerPlacement mean?  I would assume it returns the container placement of the item, but that's clearly not it, it removes an inventory item, then does some other stuff. Perhaps RemoveItemFromInventory?  Or would that be improper for some reason?

Source/ACE/Entity/Container.cs
```diff
@@ -51,14 +52,54 @@ public Container(AceObject aceObject, ObjectGuid guid)
...
+                if (place == inventoryItem.Placement)
+                    place++;
+
+                if (keyValuePair.Value.Placement != place)
+                {
+                    keyValuePair.Value.Placement = place;
+                    saveableCopy = keyValuePair.Value.SnapShotOfAceObject();
+                    saveableCopy.ClearDirtyFlags();
+                    saveableCopy.Placement = place;
+                    DatabaseManager.Shard.SaveObject(saveableCopy, null);
```
ddevec: I'm still not 100% sure what this funciton does.  But shouldn't saving these objects be piggy-backed on saving the player, not done on every inventory reshuffle?

Source/ACE/Entity/Container.cs
```diff
@@ -51,14 +52,54 @@ public Container(AceObject aceObject, ObjectGuid guid)
...
+                }
+                place++;
+            }
+            if (!removedItem) return;
+            inventory.Add(inventoryItem.Guid, inventoryItem);
+            saveableCopy = inventoryItem.SnapShotOfAceObject();
+            saveableCopy.ClearDirtyFlags();
+            saveableCopy.Placement = inventoryItem.Placement;
+            DatabaseManager.Shard.SaveObject(saveableCopy, null);
```
ddevec: See above.

Source/ACE/Entity/Container.cs
Mogwai-TheFurry: I figured that's what it was, but it's coded rather oddly.  I don't think we need to be saving on every ordering alteration.  We can let the periodic save handle this sort of stuff.

Source/ACE/Entity/Container.cs
Mogwai-TheFurry: I like: PlaceItemInContainer

Source/ACE/Entity/Container.cs
Mogwai-TheFurry: Also, need to check for container limitations?  Ie, holds 24, already at capacity, etc.  Method should probably return a boolean indicator of success so the caller can act accordingly.

Source/ACE/Command/Handlers/DebugCommands.cs
Mogwai-TheFurry: It does part of it.  The DB layer should be assigning the objectGuid of the child objects, but it doesn't hurt (and is probably better) to do a deep reassignment of the guid.

Source/ACE/Entity/Container.cs
ogmage78: I knew there had to be a better way.   Thanks Mogwai!   I will also add in capacity checks as part of this work.

Source/ACE/Entity/Container.cs
ogmage78: Will do

Source/ACE/Entity/Container.cs
ogmage78: Going to put this save in a #if directive it makes it easy to debug if it persists right away.   I will hook this in with character save which is how it will work.

LtRipley36706:
@Mogwai-TheFurry basically a function call, Container.AddToInventory(worldobject) would add it to the collection of inventory, recalc the burden of the container and send all the appropriate createobject and privateupdateints with the RemoveFromInventory doing the removeobject msg instead yeah?

Source/ACE/Entity/Container.cs
```diff
@@ -41,19 +39,20 @@ public Container(AceObject aceObject, ObjectGuid guid, string name, ushort weeni
...
-            aceObject.Inventory.ForEach(i => inventory.Add(new ObjectGuid(i.AceObjectId), new GenericObject(i)));
         }
 
         public void SendInventory(Session session)
         {
-            foreach (var invItem in inventory.Values)
-               session.Network.EnqueueSend(new GameMessageCreateObject(invItem));
+            foreach (var invItem in Inventory.Values)
+            {
+                var inv = new GenericObject(invItem);
```
LtRipley36706: should switch to `var inv = CreateWorldObject(invItem);`

Source/ACE/Entity/Container.cs
```diff
@@ -62,66 +61,51 @@ public virtual void AddToInventory(WorldObject inventoryItem, uint placement = 0
...
 
         public bool HasItem(ObjectGuid inventoryItemGuid, bool includeSubContainers = true)
         {
             if (!includeSubContainers)
-                return inventory.ContainsKey(inventoryItemGuid);
+                return Inventory.ContainsKey(inventoryItemGuid);
 
-            var containers = inventory.Where(wo => wo.Value.ItemCapacity > 0).ToList();
-            return containers.Any(cnt => ((Container)cnt.Value).inventory.ContainsKey(inventoryItemGuid));
+            var containers = Inventory.Where(wo => wo.Value.ItemsCapacity > 0).ToList();
```
LtRipley36706: instead of testing containers by ItemsCapacity, test by WeenieType or even object type (Container) if the objects in Inventory are WorldObjects?

Source/ACE/Entity/Container.cs
```diff
@@ -150,23 +134,22 @@ public void UpdateWieldedItem(Container container, uint itemId)
...
-            return inventory.Where(wo => wo.Value.WielderId != null && wo.Value.CurrentWieldedLocation < EquipMask.WristWearLeft).ToList();
+            // return Inventory.Where(wo => wo.Value.IntProperties != null && wo.Value.CurrentWieldedLocation < EquipMask.WristWearLeft).ToList();
+            return null;
         }
 
         public virtual WorldObject GetInventoryItem(ObjectGuid objectGuid)
         {
-            var containers = inventory.Where(wo => wo.Value.ItemCapacity > 0).ToList();
+            if (Inventory.ContainsKey(objectGuid))
+                return new GenericObject(Inventory[objectGuid]);
```
LtRipley36706: not all objects are GenericObject, use CreateWorldObject to get proper object

Source/ACE/Entity/Container.cs
```diff
@@ -150,23 +134,22 @@ public void UpdateWieldedItem(Container container, uint itemId)
...
+
+            var containers = Inventory.Where(wo => wo.Value.ItemsCapacity > 0).ToList();
```
LtRipley36706: wo.GetType == Container or wo.WeenieType == WeenieType.Container here maybe?

ogmage78: That was legacy from when we did not have types set and everything was a debug object, I will fix.      This was a lot more involved than I thought, so I just overlooked that. Thanks for catching it!

Source/ACE/Entity/Container.cs
ogmage78: Good catch.   I will fix

Source/ACE/Entity/Container.cs
ogmage78: That was legacy from when we did not have types set and everything was a debug object, I will fix.      This was a lot more involved than I thought, so I just overlooked that. Thanks for catching it!

ogmage78:
@LtRipley36706  Made those changes - updating now.

Source/ACE.Entity/Enum/Properties/PropertyInt.cs
```diff
@@ -2,409 +2,409 @@
 {
     public enum PropertyInt : ushort
     {
-        Undef = 0,
-        ItemType = 1,
-        CreatureType = 2,
-        PaletteTemplate = 3,
-        ClothingPriority = 4,
-        EncumbranceVal = 5, // ENCUMB_VAL_INT,
```
LtRipley36706: Lost note in formatting, please restore so we can track what we changed from client/aclogview

Source/ACE.Entity/Enum/Properties/PropertyInt.cs
```diff
@@ -2,409 +2,409 @@
...
-        ItemsCapacity = 6,
-        ContainersCapacity = 7,
-        Mass = 8,
-        ValidLocations = 9, // LOCATIONS_INT,
```
LtRipley36706: Lost note in formatting, please restore so we can track what we changed from client/aclogview

Source/ACE.Entity/Enum/WeenieType.cs
```diff
@@ -75,6 +75,7 @@ public enum WeenieType : uint
         SocialManager,
         Pet,
         PetDevice,
-        CombatPet
+        CombatPet,
+        Foci
```
LtRipley36706: Foci weren't their own weenietype, they're WeenieType.Generic.

LtRipley36706: Foci did not have their own WeenieType. They were assigned to WeenieType.Generic

Source/ACE/Entity/WorldObject.cs
```diff
@@ -1460,7 +1460,8 @@ protected WorldObject(AceObject aceObject)
                 : this(new ObjectGuid(aceObject.AceObjectId))
         {
             AceObject = aceObject;
-
+            SetWeenieHeaderFlag();
+            SetWeenieHeaderFlag2();
```
LtRipley36706: these shouldn't be needed, the WeenieHeaderFlags are set when an object reads the field. If its not updating, its a bug for sure

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
```diff
@@ -361,13 +360,22 @@ private void WriteEventBody()
...
+                Writer.Write(inv.Value.AceObjectId);
+                Writer.Write((uint)ContainerType.NonContainer);
+            }
+            // Containers and foci go in side slots, they come last with their own placment order.
+            foreach (var inv in aceObj.Inventory.Where(i => i.Value.WeenieType == (uint)WeenieType.Container || i.Value.WeenieType == (uint)WeenieType.Foci))
+            {
+                Writer.Write(inv.Value.AceObjectId);
+                if ((WeenieType)inv.Value.WeenieType == WeenieType.Container)
+                    Writer.Write((uint)ContainerType.Container);
+                if ((WeenieType)inv.Value.WeenieType == WeenieType.Foci)
```
LtRipley36706: technically weenies weren't containers, as they were classified as WeenieType.Generic.. They do have RequireBackPackSlot flag on their ObjDesc however

Source/ACE/Network/GameEvent/Events/GameEventViewContents.cs
```diff
@@ -1,13 +1,35 @@
...
+            foreach (AceObject inv in container.Inventory.Values)
+            {
+                Writer.Write(inv.AceObjectId);
+                Debug.Assert(inv.WeenieType != null, "container.WeenieType != null");
+                switch ((WeenieType)inv.WeenieType)
+                {
+                    case WeenieType.Container:
+                        Writer.Write((uint)ContainerType.Container);
+                        break;
+                    case WeenieType.Foci:
```
LtRipley36706: see above comments, possible issue with catagorization

Source/ACE.Entity/Enum/WeenieType.cs
ogmage78: I know, I needed to add one

Source/ACE/Entity/WorldObject.cs
ogmage78: It was needed it was not working.

Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
ogmage78: They are passed as containers in the character description event.

Source/ACE.Entity/Enum/WeenieType.cs
ogmage78: We align the formatting on all of the other files - you want it unaligned?   I am not sure I follow?

Source/ACE.Entity/Enum/Properties/PropertyInt.cs
ogmage78: We align the formatting on all of the other files - you want it unaligned? I am not sure I follow?

Source/ACE.Entity/Enum/Properties/PropertyInt.cs
ogmage78: We align the formatting on all of the other files - you want it unaligned? I am not sure I follow?

LtRipley36706: no issues with the alignment, it was the note `// ENCUMB_VAL_INT` that was lost.. ACE calls it EncumbranceVal, client and aclogview call it ENCUMB_VAL_INT

Source/ACE.Entity/AceObject.cs
```diff
@@ -1220,6 +1220,12 @@ public string DisplayName
             set { SetBoolProperty(PropertyBool.RequiresBackpackSlot, value); }
         }
 
+        public bool UseBackpackSlot
```
LtRipley36706: Feels like this shouldn't be in AceObject and instead in WorldObject. I follow why you put it here though. GameEventPlayerDescription reads things as AceObject and not WorldObject. 

Maybe slap a few fixme/todos to correct the anomalous property/object usage so when someone gets around to bringing GameEventPlayerDescription in line with expectations this oddity gets corrected?

Source/ACE.Entity/AceObject.cs
```diff
@@ -1829,32 +1852,51 @@ public object Clone(uint guid)
             ret.BoolProperties.ForEach(c => c.AceObjectId = guid);
             ret.DataIdProperties.ForEach(c => c.AceObjectId = guid);
             ret.InstanceIdProperties.ForEach(c => c.AceObjectId = guid);
-            ret.StringProperties.ForEach(c => c.AceObjectId = guid);    
+            ret.StringProperties.ForEach(c => c.AceObjectId = guid);
             ret.GeneratorLinks.ForEach(c => c.AceObjectId = guid);
-            
-            // No need to change Dictionary guids per DDEVEC
```
LtRipley36706: bad merge? some comment got axed, some things got duplicated

Source/ACE.Entity/AceObject.cs
```diff
@@ -1817,8 +1841,7 @@ public object Clone(uint guid)
             AceObject ret = (AceObject)Clone();
             ret.AceObjectId = guid;
             // We are cloning a new AceObject with a new AceObjectID - need to set this to false. Og II
-            ret.HasEverBeenSavedToDatabase = false;
-            ret.IsDirty = true;
+            ret.SetDirtyFlags();
```
LtRipley36706: HasEverBeenSavedToDatabase possible issue in removing this..

is it possible for something to be HasEverBeenSavedToDatabase == true and IsDirty == false? I'd assume something could be dirty that already has been saved to the database, so I don't follow the change entirely.

Source/ACE.Entity/AceObject.cs
```diff
@@ -1829,32 +1852,51 @@ public object Clone(uint guid)
...
-            Int64Properties.ForEach(x => x.ClearDirtyFlags());
-            DoubleProperties.ForEach(x => x.ClearDirtyFlags());
-            BoolProperties.ForEach(x => x.ClearDirtyFlags());
-            DataIdProperties.ForEach(x => x.ClearDirtyFlags());
-            InstanceIdProperties.ForEach(x => x.ClearDirtyFlags());
-            StringProperties.ForEach(x => x.ClearDirtyFlags());
+       public void ClearDirtyFlags()
+        {
+            this.IsDirty = false;
+            this.HasEverBeenSavedToDatabase = true;
```
LtRipley36706: see above comment re HasEverBeenSavedToDatabase

Source/ACE.Entity/AceObject.cs
```diff
@@ -1829,32 +1852,51 @@ public object Clone(uint guid)
...
+            this.DataIdProperties.ForEach(x => x.ClearDirtyFlags());
+            this.InstanceIdProperties.ForEach(x => x.ClearDirtyFlags());
+            this.StringProperties.ForEach(x => x.ClearDirtyFlags());
+            this.Inventory.ToList().ForEach(x => x.Value.ClearDirtyFlags());
+        }
+
+        public void SetDirtyFlags()
+        {
+            this.IsDirty = true;
+            this.HasEverBeenSavedToDatabase = false;
```
LtRipley36706: see above comment re HasEverBeenSavedToDatabase

Source/ACE.Entity/BaseAceProperty.cs
```diff
@@ -22,5 +17,11 @@ public void ClearDirtyFlags()
             this.IsDirty = false;
             this.HasEverBeenSavedToDatabase = true;
         }
+
+        public void SetDirtyFlags()
+        {
+            this.IsDirty = true;
+            this.HasEverBeenSavedToDatabase = false;
```
LtRipley36706: See above comment re HasEverBeenSavedToDatabase

Source/ACE.Entity/CreatureAbility.cs
```diff
@@ -111,6 +111,12 @@ public void ClearDirtyFlags()
             _backer.HasEverBeenSavedToDatabase = true;
         }
 
+        public void SetDirtyFlags()
+        {
+            _backer.IsDirty = true;
+            _backer.HasEverBeenSavedToDatabase = false;
```
LtRipley36706: See above comment re HasEverBeenSavedToDatabase

Source/ACE.Entity/CreatureSkill.cs
```diff
@@ -110,6 +110,12 @@ public void ClearDirtyFlags()
             _backer.HasEverBeenSavedToDatabase = true;
         }
 
+        public void SetDirtyFlags()
+        {
+            _backer.IsDirty = true;
+            _backer.HasEverBeenSavedToDatabase = false;
```
LtRipley36706: See above comment re HasEverBeenSavedToDatabase

Source/ACE.Entity/CreatureVital.cs
```diff
@@ -173,6 +173,12 @@ public void ClearDirtyFlags()
             _backer.HasEverBeenSavedToDatabase = true;
         }
 
+        public void SetDirtyFlags()
+        {
+            _backer.IsDirty = true;
+            _backer.HasEverBeenSavedToDatabase = false;
```
LtRipley36706: See above comment re HasEverBeenSavedToDatabase

Source/ACE/Entity/Container.cs
```diff
@@ -46,63 +37,97 @@ public Container(AceObject aceObject)
...
-                }
+                        var placement = Inventory[i.Key].Placement;
+                        return placement != null && i.Value.Placement > (uint)placement;
+                    }).ToList().ForEach(i => i.Value.Placement--));
+                containers.ForEach(i => Inventory.Remove(i.Key));
                 Burden = UpdateBurden();
             }
         }
 
         public ushort UpdateBurden()
```
LtRipley36706: does nothing?

Source/ACE/Entity/Player.cs
```diff
@@ -467,12 +469,15 @@ public AceObject GetSavableCharacter()
...
 
+            // TODO: Og II this is where I need to load aceobject.inventory with the container.inventory
+
             // immediately after cloning, clear all dirty flags and HasBeenSavedToDatabase
             Character.ClearDirtyFlags();
 
             // These don't usually get saved back to the object so setting here for now.
             // Realisticly speaking, I think it will be possible to eliminate WeenieHeaderFlags and PhysicsDescriptionFlag from the datbase
             // AceObjectDescriptionFlag possibly could be eliminated as well... -Ripley
+            // actually we do use those without creating a wo - so it would be needed to keep them in the database Og II
```
LtRipley36706: understood, i think we're almost at a point where saving these flags is entirely unnecessary. Very little is used of them as is now

Source/ACE/Entity/Player.cs
```diff
@@ -1371,14 +1376,17 @@ public bool GetVirtualOnlineStatus()
 
         private void SendSelf()
```
LtRipley36706: just something I noticed in retail pcaps, but haven't investigated much... 

following the CreatePlayer and Player Description Event there is often a DeleteObject which deletes yourself followed by a CreateObject which creates yourself. The DeleteObject message isn't fully handled by aclogview so there could some something else in that packet which puts this in to better context

Not sure I get why, but worthing noting and investigating.

Source/ACE/Entity/WorldObject.cs
LtRipley36706: maybe worth a todo/fixme.. we shouldn't have to think too much about these

Source/ACE/Factories/LootGenerationFactory.cs
```diff
@@ -29,30 +26,14 @@ public static ActionChain GetSpawnChain(WorldObject inventoryItem, Position posi
...
-
         public static void CreateRandomTestWorldObjects(Player player, uint typeId, uint numItems)
         {
             var weenieList = DatabaseManager.World.GetRandomWeeniesOfType(typeId, numItems);
             for (int i = 0; i < numItems; i++)
             {
-                WorldObject wo = CreateTestWorldObject(player, weenieList[i].WeenieClassId);
+                WorldObject wo = WorldObjectFactory.CreateNewWorldObject(weenieList[i].WeenieClassId);
+                if (wo.WeenieType == WeenieType.Container)
+                    player.Session.Network.EnqueueSend(new GameEventViewContents(player.Session, wo.SnapShotOfAceObject()));
```
LtRipley36706: any time a container is created in the world if you are in range you are shown the contents before opening?

Source/ACE/Network/GameEvent/Events/GameEventViewContents.cs
```diff
@@ -1,13 +1,27 @@
...
+using ACE.Entity.Enum;
+
+namespace ACE.Network.GameEvent.Events
 {
+    using System.Diagnostics;
+
     public class GameEventViewContents : GameEventMessage
     {
-        public GameEventViewContents(Session session, uint objectId)
+        public GameEventViewContents(Session session, AceObject container)
```
LtRipley36706: feels like this should be a WorldObject.. AceObject is behind the scenes usually

changelog.md
```diff
@@ -1,5 +1,25 @@
 # ACEmulator Change Log
 
+
```
LtRipley36706: bad merge in here

Source/ACE.Entity/AceObject.cs
ogmage78: It really has to be in both places because to do what Mogwai wanted, I don't make inventory a world object until it is used or wielded.   It is only ready only and makes the code much simpler to follow.

Source/ACE.Entity/AceObject.cs
ogmage78: it is not enough to set the flag on the top level.   This looked like it was working because until this work, we never really saved items to the database.   What I put in - I tested and it works - you have to do a deep set the setDirtyFlags that primes it to save to the database for the first time from a clone.   The flags are maintained automatically after that first save as items are changed.   It is a good system Mogwai put in, we just needed to debug the first real saves.   This does that.

Source/ACE.Entity/AceObject.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE.Entity/AceObject.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE.Entity/BaseAceProperty.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE.Entity/CreatureAbility.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE.Entity/CreatureSkill.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE.Entity/CreatureVital.cs
ogmage78: this is correct.   This call is only to be used on first save after clone.

Source/ACE/Entity/Container.cs
ogmage78: yes it was broken - I took out the borked code.

Source/ACE/Entity/Player.cs
ogmage78: We have to store the flags because we are not always working from the world object.   Sometimes, we are just at AceObject when we need to know.   if we don't keep that, then introduce a crap load of complexity where you have to instantiate a world object just to see what it's flag is.   That is very limiting IMO.

Source/ACE/Factories/LootGenerationFactory.cs
ogmage78: it is only send to the player that has the container

Source/ACE/Network/GameEvent/Events/GameEventViewContents.cs
ogmage78: no, mogway made me change inventory to only make world objects when needed.

ogmage78:
As I put in the change log, there are some known bugs, but I think this is a good start.   I am sure there are some better ways to program this.   I would love to learn them, but as this time, I will be taking a step back from the project.   I thought this added value so I am reopening this.   If you want to merge it so someone else can pick this up - great.   If not, then just close it without merge.  I had hoped I would be able to keep working on the project, but that does not seem likely at this time.   A big thank you to @Mogwai-TheFurry  for taking the time to help a very out of date programmer.   I wish you guys the very best.

Source/ACE.Entity/AceObject.cs
Mogwai-TheFurry: Bingo.

Source/ACE/Entity/Landblock.cs
```diff
@@ -792,14 +794,14 @@ public void ScheduleItemTransferInContainer(ActionChain chain, ObjectGuid wo, Co
             }
         }
 
-        public void ScheduleItemTransfer(ActionChain chain, ObjectGuid wo, ObjectGuid container)
+        public void ScheduleItemTransfer(ActionChain chain, ObjectGuid wo, ObjectGuid container, uint placement = 0)
```
Mogwai-TheFurry: I think the terminology we've been using as a semi-standard is "QueueItemTransfer".  No biggie.

Source/ACE/Entity/Landblock.cs
```diff
@@ -819,19 +821,22 @@ private void ItemTransferContainerInternal(ObjectGuid woGuid, Container containe
...
             Container container = GetObject(containerGuid) as Container;
 
-            ItemTransferContainerInternal(woGuid, container);
+            ItemTransferContainerInternal(woGuid, container, placement);
         }
 
         private void Log(string message)
         {
-            log.Debug($"LB {id.Landblock.ToString("X")}: {message}");
+            log.Debug($"LB {id.Landblock:X}: {message}");
```
Mogwai-TheFurry: The ToString was right - it would have printed the landblock Id in Hex.  You're now only logging the "X" component.

changelog.md
```diff
@@ -18,6 +39,7 @@
 ### 2017-08-06
 [Ripley]
 * Added Scroll weenie object.
+>>>>>>> 7d22ad27bd89e4462abb1f358c37f0891ea3d9a3
```
Mogwai-TheFurry: merge artifact

ogmage78: Fixed

Source/ACE/Entity/Landblock.cs
ogmage78: Fixed - damn resharper lead me astray .... ;)

Source/ACE/Entity/Landblock.cs
ogmage78: Renamed.

PR #486, opened 2017-08-06
Add Scroll Weenie

LtRipley36706:
* Added scrolls to WorldObjectFactory
* Spawn any scroll in the database to test
* Skill check is active for any scroll level 2-6
* Removed SpellObjectFactory and cleaned up some things relating to it until it can be reintroduced or refactored


Source/ACE/Entity/Player.cs
```diff
@@ -2746,5 +2753,35 @@ public void DoMoveTo(WorldObject wo)
 
             moveToObjectChain.EnqueueChain();
         }
+
+        public bool CheckMagicSkillLevel(MagicSchool school, uint power)
+        {
+            bool ret = false;
+            CreatureSkill creatureSkill = Skills[Skill.Run];
```
Mogwai-TheFurry: Run?

Source/ACE/Entity/Player.cs
```diff
@@ -2746,5 +2753,35 @@ public void DoMoveTo(WorldObject wo)
...
+                    break;
+                case MagicSchool.LifeMagic:
+                    creatureSkill = Skills[Skill.LifeMagic];
+                    break;
+                case MagicSchool.VoidMagic:
+                    creatureSkill = Skills[Skill.VoidMagic];
+                    break;
+            }
+
+            if (creatureSkill.Status >= SkillStatus.Trained && creatureSkill.ActiveValue >= (power - 50))
```
Mogwai-TheFurry: move "50" to a constant please.

Source/ACE/Entity/Scroll.cs
```diff
@@ -0,0 +1,157 @@
...
+            Power = table.Spells[SpellId].Power;
+            School = table.Spells[SpellId].School;
+        }
+
+        private uint SpellId
+        {
+            get { return (uint)Spell.Value; }
+        }
+
+        private uint Power
```
Mogwai-TheFurry: As I understand things, this is more accurately described as "MinimumSkillLevel".  I'd prefer a rename, but at least document it.

Source/ACE/Entity/Scroll.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public override void OnUse(Session session)
+        {
+            bool success = true;
+            string failReason = "You are unable to read the scroll.";
+
+            switch (Power)
+            {
+                // research: http://asheron.wikia.com/wiki/Announcements_-_2002/06_-_Castling
+                case 50:  // Level 2
```
Mogwai-TheFurry: constants please.

Source/ACE/Entity/Scroll.cs
```diff
@@ -0,0 +1,157 @@
...
+                    break;
+                default: // Level 1 or Level 7+ never fail
+                    success = true;
+                    break;
+            }
+
+            if (!session.Player.UnknownSpell(SpellId))
+            {
+                success = false;
+                failReason = "You already know the spell inscribed upon this scroll.";
```
Mogwai-TheFurry: should we not return out at this point?  no reason to run anything past this if they already know it.

Source/ACE/Entity/Scroll.cs
```diff
@@ -0,0 +1,157 @@
...
+                readScrollChain.AddAction(session.Player, () => session.Network.EnqueueSend(failMessage));
+            }
+            var sendUseDoneEvent = new GameEventUseDone(session.Player.Session);
+            readScrollChain.AddAction(session.Player, () => session.Network.EnqueueSend(sendUseDoneEvent));
+            readScrollChain.EnqueueChain();
+        }
+
+        public override void SerializeIdentifyObjectResponse(BinaryWriter writer, bool success, IdentifyResponseFlags flags = IdentifyResponseFlags.None)
+        {          
+            var propertiesInt = PropertiesInt.Where(x => x.PropertyId == (uint)PropertyInt.Value
```
Mogwai-TheFurry: a lot of this logic should be cached so we don't have to redo these lists every time.  An easy place to put this is the constructor, but you could inline it as well (but you'd have to make it threadsafe).

Source/ACE/Network/Enum/Spell.cs
```diff
@@ -8,10 +8,6347 @@ namespace ACE.Network.Enum
 {
     public enum Spell : uint
     {
-        FlameBolt           = 0x1B,
-        FrostBolt           = 0x1C,
-        LightningBolt       = 0x4B,
-        ForceBolt           = 0x56,
-        WhirlingBlade       = 0x5C,
+        Undef,
```
Mogwai-TheFurry: Was it not possible to export these with values?  Just curious. Would have made finding stuff by value a lot easier.

Source/ACE/Entity/Scroll.cs
```diff
@@ -0,0 +1,157 @@
...
+                    break;
+            }
+
+            if (!session.Player.UnknownSpell(SpellId))
+            {
+                success = false;
+                failReason = "You already know the spell inscribed upon this scroll.";
+            }
+
+            ActionChain readScrollChain = new ActionChain();
```
Mogwai-TheFurry: nice use of chains. :)

Source/ACE/Entity/Player.cs
LtRipley36706: poor coding is at play here.. i needed to define creatureSkill but the switch changes it to the right thing, vstudio complained it was unassigned. It shouldn't be possible to ever actually stay as run, but there probably a smarter way to do what I'm doing

Source/ACE/Entity/Scroll.cs
LtRipley36706: not how I recall it. If you knew the spell, you still "read" the scroll then get the "you already know this error"

I don't have screens of the actual fail messages, so I'm just going off of memory. could use a pcap or two as well of failures but, needle in a haystack

Source/ACE/Network/Enum/Spell.cs
LtRipley36706: I pulled the list from ACLogView, you'd have to check with @OptimShi

Source/ACE/Entity/Scroll.cs
LtRipley36706: Minimum Skill Level for 50% fizzle rate

Source/ACE/Network/Enum/Spell.cs
OptimShi: The values are equal to what is in the SpellTable. The enum'd names come from the client, up to the PDB version we have. (The rest were approximated to what they likely would've been, so at least they were visible in aclogview). Either we can read the SpellTable directly if you want them, or I've got a CSV dump of it if you want to enum it somehow other than this.

PR #487, opened 2017-08-07
Wired up World Broadcast Commands

LtRipley36706:
* Wired up World Broadcast commands.
* Cleaned up a few other commands that were not needed or found in other categories.


Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -905,24 +904,38 @@ public static void HandleUnlock(Session session, params string[] parameters)
...
-            // TODO: output
+            session.Player.HandleActionWorldBroadcast($"Broadcast from {session.Player.Name}> {string.Join(" ", parameters)}", ChatMessageType.WorldBroadcast);
         }
 
         // add <spell>
-        [CommandHandler("add", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 1)]
+        [CommandHandler("add", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 1,
+            "Adds the specified spell to your own spellbook.",
+            "<spellid>")]
         public static void HandleAdd(Session session, params string[] parameters)
```
Mogwai-TheFurry: I know it's not what you were doing, but it would have been nice to rename this "addSpell" while you were here.

Source/ACE/Command/Handlers/AdminCommands.cs
```diff
@@ -1193,24 +1204,35 @@ public static void HandleHousekeep(Session session, params string[] parameters)
         }
 
         // idlist
-        [CommandHandler("idlist", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0)]
+        [CommandHandler("idlist", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0,
+            "Shows the next ID that will be allocated from SQL.")]
         public static void HandleIDlist(Session session, params string[] parameters)
```
Mogwai-TheFurry: I'm not sure this is even valid.  The values certainly aren't.

Source/ACE/Entity/Landblock.cs
```diff
@@ -509,6 +510,25 @@ public void EnqueueBroadcast(Position pos, float distance, params GameMessage[]
...
+        /// Enqueues a message for global broadcast, thread safe
+        /// </summary>
+        /// <param name="msg"></param>
+        public void EnqueueGlobalBroadcast(params GameMessage[] msgs)
+        {
+            // Atomically checks and sets the broadcastQueued bit --
+            //    guarantees that if we need a broadcast it will be enqueued in the world-managers broadcast queue exactly once
+            if (Interlocked.CompareExchange(ref broadcastQueued, 1, 0) == 0)
+            {
+                WorldManager.BroadcastQueue.EnqueueAction(new ActionEventDelegate(() => SendGlobalBroadcasts()));
```
Mogwai-TheFurry: I would be inclined to think that it is safer to add to the queue prior to firing off the delegate.  Thoughts?

LtRipley36706: this is pretty much a copy of EnqueueBroadcast.. @ddevec would be best to answer. It just made sense to me that due to the global nature of the messages, a different queue should be used here. I could be mistaken

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: the command, the handler name or both?

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: if "SQL" is changed to GuidManager is that more clear?

Source/ACE/Entity/Landblock.cs
ddevec: Mogwai, I don't have time to verify for sure, but for the EnqueueBroadcast function you're probably right, that the order should be reversed.

However, I don't understand why this function was copy/pasted at all.  Why does this work need to be delayed?  For the location based broadcasts, its to ensure that two objects are not moving while updating.  For this, I don't there is a purpose to delaying the work.

LtRipley36706: @ddevec i don't follow, probably because I'm out of my depth. Feel free to pull the PR and make any changes needed. The goal is that a object in world can trigger a global broadcast. I probably should move the handlers to WorldObject or maybe Creature.. Not sure if there were any objects outside of the Creature class that could generate a world broadcast.. such as when you kill something or enlighten, or in the case of an admin, issue a global message

LtRipley36706: The only thing for certain is EnqueueBroadcast was not the right way to do a global as it does positional checks and only broadcasts to the landblock + allowed range the originator of the message was on.

ddevec: The reason that EnqueueBroadcasts has a "BroadcastQueue" is because it must wait until it knows the positions of objects in the world are not changing before doing its position-based broadcast checks, otherwise, it will race when trying to send position based broadcasts.

As you said your doing a broadcast to all things in the world, regardless of position.  You don't have the requirement which dictates the work-queue, yet you still chose to queue up the operation to send later.  This decision doesn't make sense to me.   Why don't you just make a "BroadcastToAllPlayers" function in whatever code manages our active sessions (grabbing a lock on this shouldn't be a big deal, as it s rarely changed), or do we not maintain a list of active sessions for some reason?

LtRipley36706:
@ddevec commit b486b9b22902fa03750c1b00c092ef865b88bc8b better?

Source/ACE/Managers/GuidManager.cs
```diff
@@ -164,5 +169,23 @@ public static ObjectGuid NewItemGuid()
         {
             return new ObjectGuid(itemAlloc.Alloc());
         }
+
+        /// <summary>
+        /// Returns Most Recently assigned Guid for Items / Player Items, to be used only for informational purposes.
+        /// </summary>
+        /// <returns></returns>
+        public static ObjectGuid CurrentItemGuid()
```
LtRipley36706: Ensuring these functions are not "accidentally" used incorrectly is going to be up to all who reviews the PRs. The summary clearly states they are to be only used for informational purposes. Outside of the idlist command, the functions will in all likelihood not be referenced again.

Source/ACE/Command/Handlers/AdminCommands.cs
Mogwai-TheFurry: Both.  No biggie.

Source/ACE/Command/Handlers/AdminCommands.cs
Mogwai-TheFurry: No, it is just adding 1 to the current player's Id:
`ObjectGuid nextPlayerGuid = new ObjectGuid(currentPlayerGuid.Full + 1);`

It is only correct if you're on the last character the server made.  As is, it is logically incorrect with the stated purpose from the documentation.  Even if it were correct, I don't see how this is useful.

Source/ACE/Entity/Landblock.cs
Mogwai-TheFurry: We have a list of sessions, yes.  It is only changed when people log on/off, and should be ok to lock under normal circumstances (just to grab a copy of the container list).

At a minimum, you need to swap the order of the "enqueue" vs "broadcast" logic.  You wouldn't want to have your thread paused after you triggered the broadcast but before you enqueued.  Ideally, the handler for doing a true server-wide broadcast should just do it and not queue it.

Source/ACE/Managers/GuidManager.cs
Mogwai-TheFurry: And I'm not a fan of the "idlist" functions either.  As I commented earlier, they are not implemented correctly and of questionable value.

LtRipley36706: Doing the +1 to GuidManager.Current was wrong as that is mislabled, it is +1'd on Alloc after passing the current value. Fixed my incorrect math and renamed functions to fall into line with what they are doing.

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: you're reading the function wrong. It's not the current player's guid. The +1 is wrong as I didn't realize current was already set to +1 when Alloc is called. I've fixed the +1 issue

Source/ACE/Command/Handlers/AdminCommands.cs
LtRipley36706: I don't like the idea of deviating from known commands in AdminCommands. I'll add an addspell command in DebugCommands which calls this function.

PR #488, opened 2017-08-09
Books

OptimShi:
Read Only book/page/note/etc support.

Requires the Shard/World Database update scripts and ACE-World 0.1.9 or later for DB data.


Mogwai-TheFurry:
Ping me in Discord when you can.  There's a lot I'm not understanding and I don't want to make a big deal out of it here.

PR #489, opened 2017-08-10
Added MotionTable parsing and related classes

OptimShi:


Source/ACE.DatLoader/Entity/AnimData.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace ACE.DatLoader.Entity
+{
+    public class AnimData
+    {
+        public uint AnimId { get; set; }
+        public uint LowFrame { get; set; }
+        public uint HighFrame { get; set; }
+        // Negative framerates play animation in reverse
+        public float Framerate { get; set; }
```
Mogwai-TheFurry: As a comment for future reference, if you add a full summary tag like this:
`/// <summary>Negative framerates play animation in reverse</summary>`

When you mouse hover over AnimData.Framerate as you're using it, you'll get hover-text with the comment.  Much more useful :)

Source/ACE.DatLoader/Entity/MotionData.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ACE.DatLoader.Entity
+{
+    public class MotionData
+    {
+        public byte Bitfield { get; set; }
```
Mogwai-TheFurry: Seems like Bitfield is not used?  Perhaps comment `/// <summary>Unknown usage</summary>`
Similarly, Bitfield2 is used, perhaps make an enumeration for it and give it a name?

PR #492, opened 2017-08-11
Refactor several Weenies

LtRipley36706:
* Made changes to Cow, Door, Lifestone, Portal weenies.
* Added plethora of timestamps for eventual use in WorldObject. 
  (Some of these likely will/can be move to more appropriate places after more weenie objects are built out.)

### DIFF
--- a/Source/ACE/Entity/Lifestone.cs
+++ b/Source/ACE/Entity/Lifestone.cs
@@ -37,19 +37,19 @@ namespace ACE.Entity
                 {
                     ActionChain useObjectChain = new ActionChain();
 
-                    useObjectChain.AddAction(this, () =>
+                    useObjectChain.AddAction(player, () =>
                     {
                         Activate(player);
                     });
 
                     useObjectChain.AddDelaySeconds(5); // TODO: get animation frames length and put that delay here
 
-                    useObjectChain.AddAction(this, () => player.SetCharacterPosition(PositionType.Sanctuary, player.Location));
+                    useObjectChain.AddAction(player, () => player.SetCharacterPosition(PositionType.Sanctuary, player.Location));
 
-                    useObjectChain.AddAction(this, () => player.Session.Network.EnqueueSend(new GameMessageSystemChat("You have attuned your spirit to this Lifestone. You will resurrect here after you die.", ChatMessageType.Magic)));
+                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(new GameMessageSystemChat("You have attuned your spirit to this Lifestone. You will resurrect here after you die.", ChatMessageType.Magic)));
 
                     var sendUseDoneEvent = new GameEventUseDone(player.Session);
-                    useObjectChain.AddAction(this, () => player.Session.Network.EnqueueSend(sendUseDoneEvent));
+                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(sendUseDoneEvent));
 
                     useObjectChain.EnqueueChain();
                 }
--- a/Source/ACE/Entity/Lifestone.cs
+++ b/Source/ACE/Entity/Lifestone.cs
@@ -56,12 +56,21 @@ namespace ACE.Entity
 
                     useObjectChain.AddDelaySeconds(5); // TODO: get animation frames length and put that delay here
 
-                    useObjectChain.AddAction(player, () => player.SetCharacterPosition(PositionType.Sanctuary, player.Location));
-
-                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(new GameMessageSystemChat("You have attuned your spirit to this Lifestone. You will resurrect here after you die.", ChatMessageType.Magic)));
-
-                    var sendUseDoneEvent = new GameEventUseDone(player.Session);
-                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(sendUseDoneEvent));
+                    useObjectChain.AddAction(player, () =>
+                    {
+                        var sendUseDoneEvent = new GameEventUseDone(player.Session);
+                        if (player.IsWithinUseRadiusOf(this))
+                        {
+                            player.SetCharacterPosition(PositionType.Sanctuary, player.Location);
+                            var msg = new GameMessageSystemChat("You have attuned your spirit to this Lifestone. You will resurrect here after you die.", ChatMessageType.Magic);
+                            player.Session.Network.EnqueueSend(msg, sendUseDoneEvent);
+                        }
+                        else
+                        {
+                            var failMsg = new GameMessageSystemChat("You wandered too far to attune with the Lifestone!", ChatMessageType.Magic);
+                            player.Session.Network.EnqueueSend(failMsg, sendUseDoneEvent);
+                        }
+                    });
 
                     useObjectChain.EnqueueChain();
                 }
--- a/Source/ACE/Entity/Lifestone.cs
+++ b/Source/ACE/Entity/Lifestone.cs
@@ -1,63 +1,70 @@
-﻿using ACE.Network.GameEvent.Events;
+﻿// WeenieType.Lifestone
+
+using ACE.Common;
+using ACE.Entity.Actions;
+using ACE.Entity.Enum;
+using ACE.Network.GameEvent.Events;
 using ACE.Network.GameMessages.Messages;
 using ACE.Network.Motion;
-using ACE.Entity.Enum;
-using ACE.Entity.Actions;
-using ACE.Network.Enum;
-using ACE.DatLoader.FileTypes;
 
 namespace ACE.Entity
 {
     public class Lifestone : WorldObject
     {
+        private static readonly UniversalMotion motionSanctuary = new UniversalMotion(MotionStance.Standing, new MotionItem(MotionCommand.Sanctuary));
+
         public Lifestone(AceObject aceO)
             : base(aceO)
         {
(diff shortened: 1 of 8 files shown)


Source/ACE/Entity/Lifestone.cs
```diff
@@ -1,63 +1,88 @@
...
         }
+
+        private void Activate(Player activator)
+        {
+            CurrentLandblock.EnqueueBroadcastMotion(activator, motionSanctuary);
+
+            CurrentLandblock.EnqueueBroadcastSound(this, Sound.LifestoneOn); // This event was present for a pcap in the training dungeon.. Why? The sound already comes with animationEvent...
+
+            if (activator.Guid.Full > 0)
+                UseTimestamp++;
```
Mogwai-TheFurry: Incrementing a double doesn't feel right.  Shouldn't this be setting it to something analogous to "now"?

Source/ACE/Entity/Player.cs
```diff
@@ -1433,6 +1436,7 @@ private ActionChain GetTeleportChain(Position newPosition)
             teleportChain.AddAction(this, () =>
             {
                 InWorld = true;
+                LastPortalTeleportTimestamp++;
```
Mogwai-TheFurry: Ditto the double increment comment.

Source/ACE/Entity/Player.cs
```diff
@@ -1450,6 +1454,7 @@ private void TeleportInternal(Position newPosition)
             ExternalUpdatePosition(newPosition);
             InWorld = false;
 
+            LastTeleportStartTimestamp++;
```
Mogwai-TheFurry: Ditto the double increment comment.

Source/ACE/Entity/Portal.cs
```diff
@@ -300,6 +297,8 @@ public override void HandleActionOnUse(ObjectGuid playerId)
                             // always send useDone event
                             var sendUseDoneEvent = new GameEventUseDone(player.Session);
                             player.Session.Network.EnqueueSend(sendUseDoneEvent);
+
+                            UseTimestamp++;
```
Mogwai-TheFurry: Ditto the double increment comment.

Source/ACE/Entity/Lifestone.cs
ddevec: UseTimestamp = WorldManager.PortalYearTicks (or something of that nature) is probably more appropriate.

Why does  a lifestone track its use timestamp?

Source/ACE/Entity/Player.cs
```diff
@@ -2782,6 +2830,17 @@ public void DoMoveTo(WorldObject wo)
             moveToObjectChain.EnqueueChain();
         }
 
+        public void DoCollideWith(WorldObject wo)
```
ddevec: HandleActionOnCollide?

Source/ACE/Entity/Lifestone.cs
```diff
@@ -1,63 +1,88 @@
...
-                var sendUseDoneEvent = new GameEventUseDone(player.Session);
-                player.Session.Network.EnqueueSend(lifestoneBindMessage, sendUseDoneEvent);
+                    useObjectChain.AddDelaySeconds(5); // TODO: get animation frames length and put that delay here
+
+                    useObjectChain.AddAction(player, () => player.SetCharacterPosition(PositionType.Sanctuary, player.Location));
+
+                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(new GameMessageSystemChat("You have attuned your spirit to this Lifestone. You will resurrect here after you die.", ChatMessageType.Magic)));
+
+                    var sendUseDoneEvent = new GameEventUseDone(player.Session);
+                    useObjectChain.AddAction(player, () => player.Session.Network.EnqueueSend(sendUseDoneEvent));
```
ddevec: These three actions on player should all be in a single event, not 3.

Source/ACE/Entity/Lifestone.cs
```diff
@@ -1,63 +1,88 @@
...
+                    ActionChain useObjectChain = new ActionChain();
 
-                    // create the outbound server message
-                    serverMessage = "You have attuned your spirit to this Lifestone. You will resurrect here after you die.";
-                    player.HandleActionMotion(motionSanctuary);
-                    player.Session.Network.EnqueueSend(soundEvent);
-                }
+                    useObjectChain.AddAction(player, () =>
+                    {
+                        Activate(player);
```
ddevec: Two things:

First, why are you making a new action chain for "Activate", you're already in the context of the player.

Second, its really weird that the Lifestone private function Activate assumes its run from the context of the player.  This should be commented (in both the function summary, and here).

Source/ACE/Entity/Player.cs
```diff
@@ -2516,6 +2521,49 @@ public ActionChain CreateMoveToChain(ObjectGuid target, float distance)
             return moveToChain;
         }
 
+        public ActionChain CreateCollideWithChain(ObjectGuid target)
```
ddevec: How is this a "CollideWith" chain?  It doesn't appear to handle a collision.

Source/ACE/Entity/Lifestone.cs
LtRipley36706: its not really ++ing it. The setter is doing the actual timestamp is throwing away the increment.

Source/ACE/Entity/Player.cs
LtRipley36706: the purpose is to force the player to run into or collide with the portal. Portal's use radius is -0.1. So it needs a MoveToPosition command not a MoveToObject command.

Source/ACE/Entity/Lifestone.cs
LtRipley36706: what's the difference?

Source/ACE/Entity/Player.cs
LtRipley36706: see above

Source/ACE/Entity/Lifestone.cs
LtRipley36706: was I wrong to assume the context should be run on the thing its modifying? When you use a lifestone it changes your sanctuary position, the lifestone itself is totally static

Source/ACE/Entity/Lifestone.cs
ddevec: If your not doing an increment, why would you make the code look like your doing an increment?  This is poorly designed and makes no sense.

Source/ACE/Entity/Lifestone.cs
ddevec: Your right that the context should be run on the thing it is modifying.  When you create this action chain to call Activate, your already in the context of the Player, and there is no delay in the action chain before calling activate.

Its just a redundant action chaining.  Its not "wrong", but its confusing.  Your explicitly delaying work, but with no reason.  The act of delaying work indicates there is a reason.

Source/ACE/Entity/Lifestone.cs
LtRipley36706: its quicker to code `UseTimestamp++;` than `UseTimestamp = Time.GetTimestamp();`

```cs
private double? useTimestamp;
        public double? UseTimestamp
        {
            get { return useTimestamp; }
            set { useTimestamp = Time.GetTimestamp(); }
        }
```
this also ensures that regardless of what is passed into the field, the timestamp is always set the same way.

Source/ACE/Entity/Lifestone.cs
LtRipley36706: the delay is on the next line

Source/ACE/Entity/Player.cs
ddevec: If I understand correctly, this large chunk of copy/pasted code is largely a hack to work around portal distances being -0.1.

This seems like a very convoluted way to solve that problem. Why can't you do something simpler, like change the distance calculations in MoveTo to check against a radius instead of radius squared? Or switch portals onuse distance to 0?

I still don't understand why you call HandleActionOnCollide -- that seems wrong.

Source/ACE/Entity/Lifestone.cs
ddevec: This is just abusing the syntax of properties.  If you want functionality like this, make timestamp updated with a function -- UpdateTimestamp(), not as part of the setter in a property.

LtRipley36706: two different methods with the same exact outcome.

Source/ACE/Entity/Player.cs
LtRipley36706: its also a change from MoveToObject to MoveToPosition in movementTypes.

the code works, it might be more convoluted but it does function and until physics are in its fine for now.

Source/ACE/Entity/Lifestone.cs
Mogwai-TheFurry: One is explicit and obvious about what it does.  One is deceptive and misleading about what it does and leaves potential for misuse in the future.  This isn't debatable.

LtRipley36706: actually, it is in fact. Clearly it can be done at least two different ways. It can only be considered misleading if one simply looks at the line itself and never looks at the definition nor the implementation. 

The silly thing about this PR is its mostly reused code already in the project, yet it's picked apart like its brand new never before seen code. 24 discussion points on something so simple is just absurd.

A lot of "Perfect is the enemy of good" situations feels like these days.

PR #493, opened 2017-08-11
Spell Formula

OptimShi:
Added MotionTable parsing not to changelog (was left out of previous PR)
Added GetSpellFormula() function to SpellTable. Will return component id's based on players account name for any given spell.
Fixed some logic issues in the SpellComponentsTable


Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -876,5 +876,33 @@ public static void BarberShop(Session session, params string[] parameters)
             chain.AddAction(session.Player, () => session.Network.EnqueueSend(new GameEventStartBarber(session)));
             chain.EnqueueChain();
         }
+        /// <summary>
+        /// Debug console command to test the GetSpellFormula function.
+        /// </summary>
+        [CommandHandler("spell", AccessLevel.Admin, CommandHandlerFlag.ConsoleInvoke, 0, "Tests spell formula calculation")]
```
Mogwai-TheFurry: This feels like a developer command more than an Admin command, and the name is a bit ambiguous.  Perhaps "getSpellFormula" or something over just "spell" (unless this is a legacy command, in which case just tell me so).

OptimShi: That's why it is in debug commands--so I can debug the function. I've left it in here so I can debug the issue I noted.

Mogwai-TheFurry: my concern was more with "AccessLevel.Admin", but point taken.  Clear up the merge conflict and I'm happy to merge.

OptimShi: The AccessLevel.Admin was just a holdover since I'd originally had this be an in-game function and moved it. Was more practical to move to console (so I didn't need to log in). Adjusted that AccessLevel (doubt it's even required for a console command) to be consistent with the rest and renamed the function.

OptimShi:
I believe I got that bug with the spells with extended characters in their names/description sorted out, too.

PR #495, opened 2017-08-16
Added Animation parsing from the client_portal.dat and all relevant hooks and properties.

OptimShi:


Source/ACE.DatLoader/Entity/AnimationFrame.cs
```diff
@@ -0,0 +1,44 @@
...
+                p.RotationX = datReader.ReadUInt32();
+                p.RotationY = datReader.ReadUInt32();
+                p.RotationZ = datReader.ReadUInt32();
+                a.Locations.Add(p);
+            }
+
+            uint numHooks = datReader.ReadUInt32();
+            for (uint i = 0; i < numHooks; i++)
+            {
+                AnimationHook ah = new AnimationHook();
```
Mogwai-TheFurry: the "new AnimationHook()" isn't necessary.  You could shorten these 3 lines to:
`a.Hooks.Add(AnimationHook.Read(datReader));`

Source/ACE.DatLoader/Entity/AnimationHook.cs
```diff
@@ -0,0 +1,101 @@
...
+using ACE.Entity.Enum;
+
+namespace ACE.DatLoader.Entity
+{
+    public class AnimationHook
+    {
+        public AnimationHookType HookType { get; set; }
+        public uint Direction { get; set; }
+        private IHook _hook = null;
+        public IHook Hook { get { return _hook; } set { _hook = value; } }
```
Mogwai-TheFurry: this should probably be "private set".  The reason being that this class should be the only place that hook value is ever _set_, and everybody else strictly reads it.

Source/ACE.DatLoader/Entity/AnimationHooks/CallPESHook.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ACE.DatLoader.Entity.AnimationHooks
+{
+    public class CallPESHook : IHook
+    {
+        public uint PES { get; set; }
+        public float Pause { get; set; }
+
```
Mogwai-TheFurry: both of these should be "private set".  That's going to apply to almost all of these hook sub-types so I won't spam you with comments about it.  Reason being is that no other class should be changing these values.

Source/ACE.DatLoader/Entity/AnimationHooks/ReplaceObjectHook.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class ReplaceObjectHook : IHook
+    {
+        public AnimationPartChange APChange { get; set; }
+
+        public static ReplaceObjectHook ReadHookType(DatReader datReader)
+        {
+            ReplaceObjectHook r = new ReplaceObjectHook();
+
+            AnimationPartChange ao = new AnimationPartChange();
+            ao.PartIndex = (byte)datReader.ReadUInt16();
```
Mogwai-TheFurry: Do you think it would be worth adding a "Read(DatReader datReader)" method to AnimationPartChange here?  Seems like it would make sense.

Source/ACE.DatLoader/Entity/AnimationHooks/CreateParticleHook.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public static CreateParticleHook ReadHookType(DatReader datReader)
+        {
+            CreateParticleHook hook = new CreateParticleHook();
+            hook.EmitterInfoId = datReader.ReadUInt32();
+
+            hook.PartIndex = datReader.ReadUInt32();
+
+            Position p = new Position();
+            p.PositionX = datReader.ReadSingle();
```
Mogwai-TheFurry: Consider making an extension method for Position.Read(DatReader datReader) to reduce redundancy in reading these 7 singles.  You can put this class in the DatLoader project somewhere, and then call it like you would any other method on Position.

```cs
public static class PositionExtensions
{
    public static Position ReadPosition(DatReader datReader)
    {
        Position p = new Position();
        p.Read(datReader);
        return p;
    }

    public static void Read(this Position p, DatReader datReader)
    {
        p.PositionX = datReader.ReadSingle();
        p.PositionY = datReader.ReadSingle();
        p.PositionZ = datReader.ReadSingle();
        p.RotationW = datReader.ReadSingle();
        p.RotationX = datReader.ReadSingle();
        p.RotationY = datReader.ReadSingle();
        p.RotationZ = datReader.ReadSingle();
    }
}
```

You could then call:
```cs
hook.Offset = PositionExtensions.ReadPosition(datReader);
```

Source/ACE.DatLoader/Entity/AnimationHooks/SetOmegaHook.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class SetOmegaHook : IHook
+    {
+        public AceVector3 Axis { get; set; }
+
+        public static SetOmegaHook ReadHookType(DatReader datReader)
+        {
+            SetOmegaHook so = new SetOmegaHook();
+            AceVector3 v = new AceVector3(datReader.ReadSingle(), datReader.ReadSingle(), datReader.ReadSingle());
+            so.Axis = v;
```
Mogwai-TheFurry: so.Axis = new AceVector3(...) would be easier to read.  No biggie.

Source/ACE.DatLoader/Entity/AnimationHooks/AttackHook.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class AttackHook : IHook
+    {
+        public AttackCone AttackCone { get; set; }
+
+        public static AttackHook ReadHookType(DatReader datReader)
+        {
+            AttackHook a = new AttackHook();
+            AttackCone ac = new AttackCone();
+
+            ac.PartIndex = datReader.ReadUInt32();
```
Mogwai-TheFurry: Consider giving AttackCode a "Read(DatReader datReader)" method, and then making AttackCone's property setters private.

Source/ACE.DatLoader/FileTypes/Animation.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                a.NumParts = datReader.ReadUInt32();
+                a.NumFrames = datReader.ReadUInt32();
+
+                bool hasPosFrames = ((flags & 1) > 0);
+                if (hasPosFrames && a.NumFrames > 0)
+                { 
+                    for (uint i = 0; i < a.NumFrames; i++)
+                    {
+                        Position p = new Position();
```
Mogwai-TheFurry: You would be able to reuse that Position code block I did here as well.
```cs
a.PosFrames.Add(PositionExtensions.Read(datReader));
```

ogmage78:
@OptimShi dude!   Seriously great work!

Source/ACE.DatLoader/Entity/AnimationHooks/ReplaceObjectHook.cs
OptimShi: I didn't bother, since long term the goal is to remove this class and replace with the AnimationOverride object. I've got that on my to-do list (there's a few places this is being used currently that can be replaced so we just have the one object).

Perhaps an extension, like you suggested with the Positions, would a better solution.

OptimShi:
All feedback implemented except the one I noted. Thanks for the help with this, I've learned a little bit about some C# workings--which is always good!

PR #496, opened 2017-08-16
Fix multifragment packetization logic

Zegeger:
* Rewrote packetizer code to fix incorrect multi-fragment messages (basically we shouldn't send optional headers when a single fragment fills the full packet)  
* In general this made the packetizer much cleaner and more organized  
* In the process I added support for preemptive tail sending.  Essentially for multifragment packets, the last fragment will likely be much smaller then the full packet size, so it could fit in an earlier message among other fragments.  We see this can occur in production network traces (eg the last index arrives as the first fragment of the message). Previously we did not do this, but now we can.


OptimShi:
I ran this and everything appeared to work as expected. Fixed the intermittent "garbled" book data we were experiencing, as well.

PR #497, opened 2017-08-20
Some naive, simple fellowship code. Creating and leaving fellow works.

rtmruczek:
I don't think Fellowship.cs is in the right place. Wasn't sure where to put it. Any and all feedback is great.

### DIFF
--- a/Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
+++ b/Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
@@ -46,15 +46,10 @@ namespace ACE.Network.GameMessages.Messages
                 // Write data associated with each fellowship member
                 WriteFellow(fellow);
             }
+
             Writer.WriteString16L(fellowship.FellowshipName);
-
-            // guid of fellowship leader
             Writer.Write(fellowship.FellowshipLeaderGuid);
-
-            // todo: Does this fellowship share XP?
             Writer.Write(Convert.ToUInt32(fellowship.ShareXP));
-
-            // todo: Is this an open fellowship?
             Writer.Write(Convert.ToUInt32(fellowship.Open));
 
             // todo: fellows departed?
--- a/Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
+++ b/Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
@@ -14,71 +14,56 @@ namespace ACE.Network.GameMessages.Messages
         public GameMessageFellowshipFullUpdate(Session session)
             : base(GameMessageOpcode.GameEvent, GameMessageGroup.Group09)
         {
+            // This is a naive, bare-bones implementation of 0x02BE, FullFellowshipUpdate.
+            // 0x02BE is fairly complicated, so the following code is at least valuable as an example of a valid server response.
+
+            // todo: The current implementation has race conditions,
+            // and there are questions that must be answered before it can be fixed.
+            // We need to figure out who "owns" the fellowship data.
+            // Does everyone get a turn to read from and modify the fellowship data, and if so, how is this managed?
+
+            // Currently, creating and leaving a fellowship is supported.
+            // Any other fellowship function is not yet supported.
+
             Fellowship fellowship = session.Player.Fellowship;
-            ActionChain chain = new ActionChain();
-            chain.AddAction(session.Player, () =>
+
+            Writer.Write(session.Player.Guid.Full);
+            Writer.Write(session.GameEventSequence++);
+            Writer.Write((uint)GameEvent.GameEventType.FellowshipFullUpdate);
+
+            // the current number of fellowship members
+            Writer.Write((UInt16)fellowship.FellowshipMembers.Count);
+
+            // todo: figure out what these two bytes are for
+            Writer.Write((byte)0x10);
+            Writer.Write((byte)0x00);
+
+            // --- FellowInfo ---
+
+            ActionChain fellowChain = new ActionChain();
+            foreach (Player fellow in fellowship.FellowshipMembers)
             {
-                Writer.Write(session.Player.Guid.Full);
-                Writer.Write(session.GameEventSequence++);
-                Writer.Write((uint)GameEvent.GameEventType.FellowshipFullUpdate);
+                // Write data associated with each fellowship member
+                WriteFellow(fellow);
+            }
+            Writer.WriteString16L(fellowship.FellowshipName);
 
-                // the current number of fellowship members
-                Writer.Write((UInt16)fellowship.FellowshipMembers.Count);
+            // guid of fellowship leader
+            Writer.Write(fellowship.FellowshipLeaderGuid);
 
-                // ????
-                Writer.Write((byte)0x10);
-                Writer.Write((byte)0x00);
+            // todo: Does this fellowship share XP?
+            Writer.Write(Convert.ToUInt32(fellowship.ShareXP));
 
-                // --- FellowInfo ---
-                
-                foreach (Player fellow in fellowship.FellowshipMembers)
-                {
-                    if (fellow.Guid.Full != session.Player.Guid.Full)
-                    {
-                        chain.AddAction(fellow, () => { WriteFellow(fellow); });
-                    } else
-                    {
-                        WriteFellow(fellow);
-                    }
-                }
+            // todo: Is this an open fellowship?
+            Writer.Write(Convert.ToUInt32(fellowship.Open));
 
-                Writer.WriteString16L(fellowship.FellowshipName);
+            // todo: fellows departed?
+            Writer.Write(0u);
+            Writer.Write(0u);
 
-                // guid of fellowship leader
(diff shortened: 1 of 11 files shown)


Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,90 @@
...
+                Writer.WriteString16L(session.Player.Name);
+            }
+
+            Writer.WriteString16L(fellowship.FellowshipName);
+            
+            // guid of fellowship leader
+            Writer.Write(fellowship.FellowshipLeaderGuid);
+
+            // todo: xp share?
+            Writer.Write((byte)0x01);
```
ddevec: These are okay for now, but they should eventually be written out as integers:

// TODO: is xp share supported?
Writer.Write((uint)0x1);

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,90 @@
...
+            Writer.Write((UInt16)fellowship.FellowshipMembers.Count);
+
+            // ????
+            Writer.Write((byte)0x10);
+            Writer.Write((byte)0x00);
+
+            // --- FellowInfo ---
+            foreach (Player fellow in fellowship.FellowshipMembers)
+            {
+                Writer.Write(fellow.Guid.Full);
```
ddevec: This part is actually tricky :-/.

We have a multi-threaded server, and you can't just go reading the Player's values without any form of concurrency control.

The way we manage concurrency control between in-game entities is with an event-based system we call "Actions" the most common way we use "Actions" is with an "ActionChain" (or a series of actions which happen one after the other).

In summary:
Actions run on a WorldObject's turn (a Player is a WorldObject).  This guarantees that any accesses to that object's data wont race with any others.

The proper code skeleton to do this would look something like this:

ActionChain chain = new ActionChain();
foreach (Player fellow in fellowhsip.FellowshipMembers)
{
    chain.AddAction(player, () =>
    {
        ...Write Player properties here...
    }
}

chain.AddAction(session.Player, ()
{
    ... Write fellowship name and leader guid here ...
}
chain.EnqueueChain();

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                Writer.Write((byte)0x10);
+                Writer.Write((byte)0x00);
+
+                // --- FellowInfo ---
+                
+                foreach (Player fellow in fellowship.FellowshipMembers)
+                {
+                    if (fellow.Guid.Full != session.Player.Guid.Full)
+                    {
+                        chain.AddAction(fellow, () => { WriteFellow(fellow); });
```
rtmruczek: Correct me if I have any of this wrong.. `session.Player` is already locked from being added to the chain above, so we can't have any other actions try to read from the player. If we try to add `session.Player` as an Actor again, we won't be able to get any data from it in the action that is passed.

However, we still need to add the other `Player`s to the chain before we try to read from them, so the rest of the `Player`s are added to the chain.

ddevec: The ActionChain is a method of delaying action (not locking), it stops two threads from accessing the data at the same time by delaying work until its sure only one element will access that data.

This means it will delay the work until the lock is released, and you wont have to worry about deadlocks.

We didn't use a locking scheme for synchronizing between different interacting game entities due to lock-cycles and deadlocking dangers.

ddevec: Here your modifying "chain" from within an action running on the chain.  This probably doesn't result in the behavior your after.  Have you tested this works?

ddevec: I just realized, I misunderstood your comment last night, and probably didn't provide the information you needed.

ActionChains are designed to sequence "Actions" which should run one after the other.  In this instance, your adding to the chain while its running.  This is messy business and probably doesn't do what you want.

Instead consider something like this:

```cs
chain.AddAction(session.Player, () =>
{
    ... Write up until the fellowship info foreach ...
    ActionChain internalChain = new ActionChain();
    foreach (Player fellow in fellowhsip.FellowshipMembers)
    {
        internalChain.AddAction(fellow, () => WriteFellow(fellow) );
    }
    internalChain.AddAction(session.Player, () =>
    {
        ... Write fellow name through end ...
    }
    internalChain.EnqueueChain();
}
chain.EnqueueChain();
```

This will cause internalChain to be created when the action in chain is actually run.

There's a larger question here of when the "fellowship" work should be done (the fellowship itself is a structure which can be concurrently accessed by other players -- at what point should its work be scheduled so we guarantee it doesn't race with other accesses/modifications to the fellowhship).

ddevec:
After our discussions, I think we've decided to punt on concurrency within fellow requests for now (with a FIXME: todo later).

If you add a "TODO: properly handle concurrency for fellowship requests" around all of the chain parts that modify the fellowship, I'll be okay with accepting this as is.

rtmruczek:
Cool, that's pretty much the road I started going down for now. I have some comments that explain the problem and why/how the code needs to be fixed. I'll push it up tonight.

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+
+                // todo: open fellow?
+                Writer.Write((byte)0x01);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+
```
OptimShi: This is "even_xp_split". Presumably (uint)1 for all members > 50 / or within 5 levels of leaders. Will need to find logs to clarify, but stands to reason based on property name.

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                // todo: fellows departed?
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+
```
OptimShi: This is "open_fellow". (uint)1 or (uint)0

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                Writer.Write(0u);
+
```
OptimShi: This is locked. (uint)1 or (uint)0. (You'll notice most of the items that are technically bools, are still typically still 4 byte uints.)

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x20);
+                Writer.Write((byte)0x00);
+
```
OptimShi: "fellows_departed". PackableHashList<uint, uint>

Source/ACE/Network/GameMessages/Messages/GameMessageFellowshipFullUpdate.cs
```diff
@@ -0,0 +1,110 @@
...
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x00);
+                Writer.Write((byte)0x20);
+                Writer.Write((byte)0x00);
```
OptimShi: And this is some sort of fellowship lock/flag. Not used by the client, but still send it to keep aclogview happy as it is what the retail servers did. I'd hazard a guess that we'll never have this be anything other than what you've got here (though you could more efficiently use "Write.Write((uint)0x00200000);")

Mogwai-TheFurry:
It would make sense to me that the server would own Fellowships as a top-level-entity, just as it does Sessions.  ActionChains should be used to queue alterations to them.  The biggest reason is due to how XP is shared - by distance.  Distance and movement calculations have to be done in Action Chains, and earning XP is no different.  XP earned on the same landblock is fully shared by others on the same landblock.

Side effect: this also means earning XP while in a fellowship becomes an Action Chain.

ddevec:
Right now our system is designed so that position will not change during the "action phase".  This means that if fellowships handle their actions (and xp distributions, etc) at the same time as Player's act we wont have to use an ActionChain to do so.   We may consider doing it anyway, for consistency, but its unneeded.

ddevec:
I was about to merge, and realized this commit doesn't have a changelog entry.  Could you add one to changelog.md?

PR #505, opened 2017-09-13
Weenie editor & basic recipes

Mogwai-TheFurry:


Source/ACE/Entity/Player.cs
```diff
@@ -2456,6 +2473,30 @@ public void HandleActionDropItem(ObjectGuid itemGuid)
 
             dropChain.EnqueueChain();
         }
+        
+        public void HandleActionUseOnTarget(ObjectGuid sourceObjectId, ObjectGuid targetObjectId)
+        {
+            CurrentLandblock.EnqueueAction(new ActionEventDelegate(() =>
```
ddevec: This should be an action chain on the player, not a direct action on the landblock (its actually not safe to access landblock here, because it could be actively being modified while read :-/).  Instead try:

ActionChain chain = new ActionChain(this, () => 
{ 
    ... 
});
chain.EnqueueChain();

Source/ACE/Managers/RecipeManager.cs
```diff
@@ -0,0 +1,135 @@
...
+                    case RecipeType.Tinkering:
+                        break;
+                    case RecipeType.Unlocking:
+                        break;
+                    case RecipeType.None:
+                        return;
+                }
+            });
+
+            craftChain.AddAction(player, () => player.HandleActionMotion(new UniversalMotion(MotionStance.Standing, new MotionItem(MotionCommand.Ready))));
```
ddevec: These are performance, not correctness -- fix if you wish...

You may consider coalescing this action with the prior one in the chain.

Your also calling player.HandleAction<thing> from within the Player's action.  HandleAction functions by convention create their own actions, so your doubling down on action creation here (this will effectively spend the player's next action phase creating the action, then running the action in the phase after that.  Ideally you would run the action in the next phase).

Source/ACE/Entity/Player.cs
Mogwai-TheFurry: will do.

Source/ACE/Managers/RecipeManager.cs
ddevec: Yes, its fine to call then (although I would still coalesce the call into the prior action on the chain).  I screwed up naming it -- It should have been something like "HandleMotion".  I should fix that in a future PR :-/.

ogmage78:
Cool stuff!    Let's get this in

PR #506, opened 2017-09-13
Stubs and enums for crafting dialogs

ogmage78:
Added one new game action, and two new event message handlers - as well as the associated enums.   This work is around crafting dialogs.


ogmage78:
@mogwai added this to his pr.    Closing.

PR #509, opened 2017-09-18
items from failed crafting

Mogwai-TheFurry:


ogmage78:
Tested - this works great.   Should be merged asap.   tested crafting (motes to slivers) works as expected.   Regression tested, saving character, adding other items via ci - dropping and picking back up.   Equipping items.   No persist but that is waiting on my PR.

PR #510, opened 2017-09-18
Starter gear

Mogwai-TheFurry:


ogmage78:
I merged on my test - there was really no conflict - just the readme and player just had a comment header to merge.   This also looks good.   As a TODO - the shirt, pants, head gear and boots need to be equipped.   I can add that in after my PR as all of the tools will be in place.   Regression tested - all passed - Ready for merge.

PR #511, opened 2017-09-19
Don't bend the knee - Wielded items persisted  and a few old bugs squashed.   READY TO MERGE

ogmage78:
LATE UPDATE - all known bugs squashed via a workaround until we can enhance the ORM to deal with our inventory issue.
**Update 9/27/2017 PM  - I have tried everything - I must be missing something simple.   I it looks like it should work - but it still crashes on second save after moves. I really need some help here - I have looked at this so much I am going to just trash on it - maybe a fresh set of eyes?  
Steps to reproduce.   For simplicity's sake, I have a slimmed down set of test items, 1 wand and 1 pack. Make sure you start out with no items as it will make debugging much easier:

1.  login clean - ie neckid
2.   " at sign weapons"
3.   " at sign save-now"
4.    You can use the sql attached below to see what is going on in the db.   You will need to rename, it will not take sql attachments
5.  now move the wand to equipped and @save-now
6.  check db the db is not updated correctly at this point.   If you hit save again it works however, it is fubar from there.**  

[containerwieldedview.txt](https://github.com/ACEmulator/ACE/files/1338833/containerwieldedview.txt)

UPDATE 9/27/2017 - Fixed all of Morwai's feedback.   Bug progress - it looks like the duplicate key issue on save is that on a container to container move, the item still shows in both inventory lists.   Not sure how that is happening.   Still debugging.

UPDATED: 9/26/2017
* Implemented persisting of wielded items.   
* Changed inventory to manage a list of world objects instead of ace objects.   This was needed for sequence issues.
* fixed combat animation that had been broken for a while due to prior refactoring.
* fixed moving inventory from container to container with persistances which was broken due to prior refactoring.
* implemented the response message to send object description.
* cleaned up so code, aligned enums.
* TODO.   I have a dirty flag bug that causes issues when you do container to container moves and save.   
* I put in a hack to make this work, see around line 767 in Shard database. - it has changed a lot and is probably holding others up.   
* This seems to work in all cases.   I have spent hours on the dirty flag bug and this was the only way I could make it work.   Once someone @Mogwai-TheFurry  :) has time to look at this, we can take out the hack.   

BTW - since no one gets my humor, this is not about the NFL it is about it has taken me a month to get this close and I refused to "bend the knee" that it was over my head to fix.  ;)
maxresdefault 1


OptimShi:
I believe this is it. It worked for me, at least.

ShardDatabase.cs, line 935...
Should be
 transaction.AddPreparedDeleteStatement<ShardPreparedStatement, AceObjectPropertiesInstanceId>(ShardPreparedStatement.DeleteAceObjectPropertyIid, prop);

DeleteAceObjectPropertyIid, singular, not Properties, plural. (Might as well fix line 905 while you're at it, same issue I would think with Did

ogmage78:
This is just about complete.   Only known bug at this point is taking a wielded item and dropping it to the ground.   I will fix that last bug, do some code clean up and this should be ready to go.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -592,23 +602,31 @@ public bool SaveObject(AceObject aceObject)
...
 
             // Do we have any inventory to save - if not, we are done here?
-            if (aceObject.Inventory.Count <= 0)
-                return transaction.Commit().Result;
-
-            foreach (AceObject invItem in aceObject.Inventory.Values)
+            if (aceObject.Inventory.Count > 0)
             {
-                SaveObjectInternal(transaction, invItem);
+                foreach (AceObject invItem in aceObject.Inventory.Values)
```
Mogwai-TheFurry: Sent you a note about this via Discord.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -747,7 +763,15 @@ private bool SaveAceObjectPropertiesInt(DatabaseTransaction transaction, uint ac
                 }
                 else if (prop.PropertyValue != null)
                 {
-                    transaction.AddPreparedInsertStatement<ShardPreparedStatement, AceObjectPropertiesInt>(ShardPreparedStatement.InsertAceObjectPropertiesInt, prop);
+                    // FIXME: There is an issue with the dirty flags.   This is a hack to fix it until someone with more knowledge of this ORM can take a look.
```
Mogwai-TheFurry: sent you a note about this via Discord

Source/ACE/Entity/Player.cs
```diff
@@ -1390,16 +1406,57 @@ public ActionChain GetSaveChain()
         }
 
         /// <summary>
+        /// This method is used to clear the wielded items list ( the list of ace objects used to save wielded items ) and loads it with a snapshot
```
Mogwai-TheFurry: I would still prefer all the Snapshop methods accept a boolean indicating whether or not to clear the dirty flags.  It's a bit of a trap call right now that could cause data loss down the road if called incorrectly for whatever reason.

Source/ACE/Entity/Player.cs
```diff
@@ -2211,114 +2417,109 @@ public void UpdateAppearance(Container container, uint? itemId)
             }
         }
 
-        public void HandleActionWieldItem(Container container, uint itemId, uint location)
+        /// <summary>
+        /// This method sets properties needed for items that will be child items.
+        /// Items here are only items equipped in the hands.  This deals with the orientation
+        /// and positioning for visual apperence of the child items held by the parent. Og II
+        /// </summary>
+        /// <param name="container"></param>
```
Mogwai-TheFurry: please remove the param tags if they're just empty

Source/ACE/Entity/EquippedItem.cs
```diff
@@ -9,16 +9,19 @@ namespace ACE.Entity
     /// <summary>
     /// This Class is used to add children
     /// </summary>
-    public class EquippedItem
+    public class HeldItem
```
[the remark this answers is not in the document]
ogmage78: Fixed

Source/ACE.Database/ShardDatabase.cs
ogmage78: reworked - I think the problem might be in a different area.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -592,23 +602,31 @@ public bool SaveObject(AceObject aceObject)
...
-                    SaveObjectInternal(transaction, contInvItem);
+                    foreach (AceObject contInvItem in invItem.Inventory.Values)
+                    {
+                        SaveObjectInternal(transaction, contInvItem);
+                    }
                 }
             }
 
+            // Do we have any wielded items to save - if not, we are done here?
+            if (aceObject.WieldedItems.Count <= 0)
```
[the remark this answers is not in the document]
ogmage78: fixed.

Source/ACE/Entity/Player.cs
ogmage78: Done, but not using this yet - I put this in but put a default parameter of false to mimic current behavior.   When you have time, can you define when we would want to use that vs when it is handled behind the scenes.

Source/ACE/Entity/Player.cs
ogmage78: opps - missed those all filled in now.

Mogwai-TheFurry:
Not perfect, as we've discussed, but a giant step in the right direction.  I'll take it.

Source/ACE/Entity/Player.cs
```diff
@@ -1896,6 +1975,19 @@ public void DoFinishBarber(ClientMessage message)
                 new GameMessageObjDescEvent(this));
         }
 
+        /// <summary>
+        ///  Sends object description if the client requests it
+        /// </summary>
+        /// <param name="item"></param>
+        public void HandleActionForceObjDescSend(ObjectGuid item)
+        {
+            WorldObject wo = GetInventoryItem(item);
```
ddevec: HandleAction without an ActionChain.  This is a problem.

Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -779,7 +779,8 @@ public static void HandleListPlayers(Session session, params string[] parameters
             "Creates 1 of each weapon class in your inventory.")]
         public static void HandleWeapons(Session session, params string[] parameters)
         {
-            HashSet<uint> weaponsTest = new HashSet<uint>() { 93, 127, 130, 136, 136, 136, 148, 300, 307, 311, 326, 338, 348, 350, 7765, 12748, 12463, 31812 };
+            // HashSet<uint> weaponsTest = new HashSet<uint>() { 93, 127, 130, 136, 136, 136, 148, 300, 307, 311, 326, 338, 348, 350, 7765, 12748, 12463, 31812 };
+            HashSet<uint> weaponsTest = new HashSet<uint>() { 120, 134, 300, 307, 36561 };
```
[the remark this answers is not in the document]
ogmage78: I did a quick set of testing enums to get rid of the magic numbers.   I was going to add them all but 32k enums makes the build time go to minutes instead of seconds.   We had discussed (Mogwai) of not wanting that large of a list of enums.   That said, if you know another way to accomplish that, I will be glad to do it.

PR #514, opened 2017-09-30
Enum all the things with attributes - Restrict send to the client based on enum attributes.

ogmage78:
# Ready for Review and Merge

**UPDATE 10/02/2017 - moved to static method per request and used generics to get rid of duplicate code.**

* Added enum attributes for all of our property enums per Mogwai using the data from OptimShi found here
* http://ac.yotesfan.com/ace_object/not_used_enums.php
* Filter out all property types sent to client based on them not being [ServerOnly] attributes.
* NOTE: the big gotcha with this is if we have any of the [ServerOnly] attributes set incorrectly, that data will not get sent to the client
* and if you don't remember that we are filtering by that attribute you will be going what the hell is wrong.   Just an FYI
* There is probably a better way to do this, if so, please let me know.
image


Source/ACE/Network/GameEvent/Events/GameEventPlayerDescription.cs
```diff
@@ -50,9 +51,12 @@ private void WriteEventBody()
 
             var aceObj = Session.Player.GetAceObject() as AceCharacter;
 
-            // < 9000 to filter out our custom properties
-            var propertiesInt = aceObj.IntProperties.Where(x => x.PropertyId < 9000).ToList();
-            if (propertiesInt.Count != 0)
+            var enumValuesInt = typeof(PropertyInt).GetFields().Select(x =>
```
ddevec: These sets are actually static values that are recalculated every execution, that's sub-optimal both computationally and stylistically, I would encourage you to make them a static class property somewhere.

ogmage78: Just trying to make sure I understand - the call that gets what is a server only enum is not going to change so it can be static - the comparison to the items to be written - that can change -so basically make the first part static correct?

ddevec: yup.

Mogwai-TheFurry: ditto this.

PR #516, opened 2017-10-01
Turn right Round - Initial work on content

ogmage78:
* Initial work on content interactions.
* Cleaned up the using statements on several files.
* Aligned the enums to = sign
* Fixed bug in universal motion using the TurnToObject
* Moved a method from player to creature.   We need to access that method for the ancestor class

Normally, I would not submit this as a PR yet, but my PR's tend to get big and touch a lot of stuff.   I just put in the very start of this, and cleaned up a few files - and aligned some enums and squashed a bug.   It is functional and moves us forward, so I thought better to review in small functional chunks.   

You have to know 80's MTV to get the picture.... 
image


Source/ACE/Entity/Creature.cs
```diff
@@ -587,5 +586,66 @@ public void DoWorldBroadcast(string message, ChatMessageType messageType)
...
+            CurrentLandblock.ChainOnObject(chain, playerId, wo =>
+            {
+                Player player = wo as Player;
+                if (player == null)
+                {
+                    return;
+                }
+
+                if (!player.IsWithinUseRadiusOf(this))
+                    player.DoMoveTo(this);
```
ddevec: Does this actually work?  If you "Use" something while your away, you should move to it, then "Use" it.  From this code it looks like your just doing the motion and not the use.

ogmage78: Yes it works.   I just put in the move to and turn.   I need to read the expected behavior from the weenie that does not exist yet.   Just reall prof of concept but it does work and is a building block for what I need to do next

ogmage78: yea after you reach your destination it calls back and then does the rest of the actions.   Sorry misunderstood the question.

Forbiddenz: It works, I use t in vendor and my guy will walk to the vendor and then then the I use fires and it opens

PR #517, opened 2017-10-02
Movement Updates

OptimShi:
* Changed MotionCommand and MotionState enums to full uint instead of short and adjusted the movement serializing functions to work with these changes.
* Fixed bug when leaving CombatMode with no ammunition equipped (e.g. melee or magic mode)
* Added "acecommands" tip to console start up


Source/ACE.Entity/Enum/MotionCommand.cs
```diff
@@ -6,415 +6,415 @@
...
-        PunchSlowHigh,
-        PunchSlowMed,
-        PunchSlowLow,
-        OffhandPunchFastHigh,
-        OffhandPunchFastMed,
-        OffhandPunchFastLow,
-        OffhandPunchSlowHigh,
-        OffhandPunchSlowMed,
-        OffhandPunchSlowLow
+        Invalid = 0x0,
```
ddevec: Could you use Og's enum value alignment plugin, so we don't get another PR where we have an entire file rewritten over alignment?

PR #526, opened 2017-10-06
Vendor improvements

Forbiddenz:
Simplified some of the vendor code.. little faster less action chains same result.
Fixed bug with purchasing multiple items of the same type with no max stack size.
Fixed Filter / Shop Only Items to support latest World DB changes of destinationTypes

### DIFF
--- a/Source/ACE/Entity/Player.cs
+++ b/Source/ACE/Entity/Player.cs
@@ -50,6 +50,12 @@ namespace ACE.Entity
         /// </summary>
         private Dictionary<ObjectGuid, WorldObject> interactiveWorldObjects = new Dictionary<ObjectGuid, WorldObject>();
 
+        private Dictionary<ObjectGuid, WorldObject> vendorsLastKnownDefaultList = new Dictionary<ObjectGuid, WorldObject>();
+
+        private ObjectGuid currentVendorObjectGuid;
+
+        private double currentVendorSellRate = 1.0;
+
         /// <summary>
         /// Used to store money during a transaction in progress.
         /// </summary>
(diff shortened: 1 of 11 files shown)


Source/ACE/Entity/Player.cs
```diff
@@ -1106,39 +1126,112 @@ public void HandleActionApproachVendor(Vendor vendor, List<WorldObject> itemsFor
...
-            });
-            chain.EnqueueChain();
-        }
-
-        public void HandleActionBuyStartTransaction(WorldObject vendor, List<WorldObject> purchaselist, uint cost)
-        {
             new ActionChain(this, () =>
             {
-                // todo: check inventoy space.
+                List<ItemProfile> filteredlist = new List<ItemProfile>();
```
Mogwai-TheFurry: This is just too much code in a Lambda, and it's simply in the wrong class.  Make this a function in the Vendor class - it doesn't belong in player.

Source/ACE/Entity/Player.cs
```diff
@@ -50,6 +50,12 @@ public sealed class Player : Creature, IPlayer
         /// </summary>
         private Dictionary<ObjectGuid, WorldObject> interactiveWorldObjects = new Dictionary<ObjectGuid, WorldObject>();
 
+        private Dictionary<ObjectGuid, WorldObject> vendorsLastKnownDefaultList = new Dictionary<ObjectGuid, WorldObject>();
+
+        private ObjectGuid currentVendorObjectGuid;
+
+        private double currentVendorSellRate = 1.0;
```
Mogwai-TheFurry: Not a fan of this implementation.  This is a Vendor property for use in the Vendor class.  Putting it here is a major code smell for violation of proper encapsulation.

PR #528, opened 2017-10-07
Keys are key

kdmxp256:
An easy location to test is in the Training Academy Central Courtyard.  The first door encountered is locked.  Use @ci 30999, to spawn the Academy Library key.

- Keys will decrement their usages.  Examining the key shows the decreasing number of usages; usage icon overlay works.  Key will be removed from both the Shard DB and the player's inventory on last usage.

- KeyCode  and LockCode are used in the code, but no data is loaded into the DB, yet, resulting in all keys working with all locked door, ie. each key is a AC master key.

TODO:
- Not sure if there should be a sound played when the door re-locks, and if so, what it would be.
- The notification message for attempting to use a key on an open door needs to be corrected.  Enum defined in StatusMessageType1.cs is not correct.
- A notification message appropriate for using a key on an unlocked door that is closed does not appear to exist/defined in StatusMessageType1.cs.


Source/ACE/Entity/Key.cs
```diff
@@ -36,26 +39,49 @@ public void HandleActionUseOnTarget(Player player, WorldObject target)
...
+                                {
+                                    // Remove item from players
+                                    player.DestroyInventoryItem(this);
+                                    // Clean up the shard database.
+                                    DatabaseManager.Shard.DeleteObject(SnapShotOfAceObject(), null);
+                                }
+                                player.Session.Network.EnqueueSend(sendUseDoneEvent);
+                                break;
+                            case Entity.Door.UnlockDoorResults.DoorOpen:
+                                var messageDoorOpen = new GameEventDisplayStatusMessage(player.Session, StatusMessageType1.Enum_0481);
```
Mogwai-TheFurry: 0481?  If you can come up with a better name for that, please do.

kdmxp256: That is just the name of the Enum from StatusMessageType1.cs already present.  I did not create that.  Technically the message associated with that enum isn't even correct, anyways.  The client displayed "The monarch doesn't own either a mansion or villia", instead of "You cannot unlock or lock open doors" as defined in StatusMessageType1.cs

PR #529, opened 2017-10-10
Vendor Improvements

Forbiddenz:
Fixed buy rates, sell rate math.
Added SetCoin debug cmd
Attempted to tighten up loops..
Reduced action chains down.


Source/ACE/Entity/Player.cs
```diff
@@ -1205,76 +1172,49 @@ public void HandleActionSell(List<ItemProfile> items, ObjectGuid vendorId)
...
-                        // check to see if item is in players inventory.
-                        WorldObject wo = GetInventoryItem(item.Guid);
-                        DestroyInventoryItem(wo);
-                        escrowWorldObjects.Add(wo);
+                        DestroyInventoryItem(item);
                     }
                 }
 
-                // Send Items to Vendor for processing..
+                // Send Items to Vendor for Final processing..
```
Mogwai-TheFurry: shouldn't this go inside the "if (valid)" above?

Source/ACE/Entity/Player.cs
```diff
@@ -1164,18 +1134,15 @@ public void HandleActionBuyFinalTransaction(WorldObject vendor, List<WorldObject
...
-                    }
                 }
                 else
                 {
                     // vendor said no go!
-                    escrowCoin = 0;
                     valid = false;
                 }
 
+                // send items back to vendor with result
```
Mogwai-TheFurry: shouldn't this be in an "if (valid)" check?

PR #530, opened 2017-10-10
DDDInterrogation uses 32 bit quantities, not 64 bit.

dworkin:
Though the client doesn't seem to mind.


Mogwai-TheFurry:
Happy to merge this if someone can validate it.

PR #542, opened 2017-10-20
Fix to Value and Burden

kdmxp256:
* Created WO.StackUnitValue and WO.StackUnitBurden to be initialized by Weenie.{Value,EncumbranceVal}, with builtin null to zero initialization
* Moved WorldObject value and encumbrance calculations down to the WorldObject class, based upon WO.StackUnitValue and WO.StackUnitBurden vs WO.StackSize


Source/ACE/Entity/Vendor.cs
```diff
@@ -215,11 +215,15 @@ public void BuyValidateTransaction(ObjectGuid vendorid, List<ItemProfile> items,
             foreach (WorldObject wo in uqlist)
             {
                 goldcost = goldcost + (uint)Math.Ceiling(SellRate * (wo.Value ?? 0) * (wo.StackSize ?? 1) - 0.1);
+                wo.Value = wo.Value;        // Also set the stack's value for unique items, using the builtin WO calculations
```
LtRipley36706: I find this sorta odd that these lines are required.. i would think it should have auto updated the value based on on the getter.. maybe something else is needed though..

kdmxp256: Yes, I tested it without these lines present; however, the value and burden of the items, regardless of stack count, didn't set themselves correctly, until the stack was manipulated.  I am wondering maybe item examine is one of the processes that acts upon the AO, instead of the WO, as you mentioned at one point.

PR #544, opened 2017-10-22
Fix coin value and pickup of stack-able items from landblock

kdmxp256:
* Add the initial set of CoinValue in Container.cs to the value of contained Coins
* Modified item move, drop, and pickup methods to account for Pyreal changes, using StackOverflow's UpdateCurrencyClientCalculations();

### DIFF
--- a/Source/ACE.Database/ShardDatabase.cs
+++ b/Source/ACE.Database/ShardDatabase.cs
@@ -272,7 +272,6 @@ namespace ACE.Database
             return transaction.Commit().Result;
         }
 
-
         public void RemoveAllFriends(uint characterId)
         {
             throw new NotImplementedException();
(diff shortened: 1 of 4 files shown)


kdmxp256:
Tested splitting Pyreals into multiple stacks of varying sizes.  Tested moving between main pack and sub-packs.  Tested putting pyreals into sub-packs and dropping sub-packs.  CoinValue updates correctly.  Dropping Pyreal stacks also update CoinValue correctly.  Pickup of unsplit stacks works.  However, pickup of stacks of anything that have been split don't work, so cannot fully test Pyreal pickup.

kdmxp256:
If a stack-able item is picked up and no same items exist in inventory, HandleActionPutItemInContainer() method is called.  If a same item already exists in inventory, HandleActionStackableMerge() is called.  Found by debugging.  Added the pickup code with a !HasItem check at the beginning of HandleActionStackableMerge().

PR #546, opened 2017-10-24
Fixed Authentication Handler to log and launch for JwtToken issues, added use to some sql update scripts

Thwargle:


Database/Updates/Shard/2017-08-20-00-vw-ace-wielded-object.sql
```diff
@@ -1,3 +1,4 @@
+use ace_shard;
```
Mogwai-TheFurry: We historically and intentionally do not have using statements.  I understand it makes things easier to mess up, but that's the policy.

Source/ACE/Network/Handlers/AuthenticationHandler.cs
```diff
@@ -29,9 +29,18 @@ public static class AuthenticationHandler
         public static void HandleLoginRequest(ClientPacket packet, Session session)
         {
             PacketInboundLoginRequest loginRequest = new PacketInboundLoginRequest(packet);
-            log.Info($"First 2 of token: {Convert.ToByte(loginRequest.JwtToken[0])}, {Convert.ToByte(loginRequest.JwtToken[1])})");
-            Task t = new Task(() => DoLogin(session, loginRequest));
-            t.Start();
+            if (!string.IsNullOrEmpty(loginRequest.JwtToken))
```
Mogwai-TheFurry: 1) as of .net 4, the proper check is string.IsNullOrWhitespace
2) should probably check for len >= 2 as well
3) I'm fine with removing this debug statement.  it's holdover from when i implemented things.

Database/Updates/Shard/2017-08-20-00-vw-ace-wielded-object.sql
LtRipley36706: I see the value in `use` statements, is this something worth revisiting to make easier folks pulling down the repo and using the typical values?

Mogwai-TheFurry:
these changes were mostly rolled into other PRs throughout the night, aside from the "using" issues.  closing this.

PR #552, opened 2017-10-25
Ready for review; AccessLevel from subscrptions, Api Auth update

fantoms:
* first tests are successful.
* Api will take the first subscription's accesslevel, when checking roles; this should probably be set to take the highest level. AccessLevel of Player (0) is used when a subscription is missing.
* Adding Mogwai's objects in and credits to Mogwai for setting up the functions to test auth.
* Prevent crash and block user when password/username mismatched.


fantoms:
To test, I built out a quick console app and pushed too this github repo: https://github.com/fantoms/ACEJsonExplorer

PR #553, opened 2017-10-25
Service Installer - Phase one - Implement for game server

kdmxp256:
* Implemented phase one of enabling ACE to support running as a Windows service
* Once this PR is accepted and merged, will work on converting the API Host and Auth API Host processes

* TODO: convert console processes to some other means, as the console is not accessible when running as a service


Mogwai-TheFurry:
I'm ok with this as long as the regular console app works, but I want secondary approval prior to merging.

fantoms:
I think the console is kind of essential for some features and very useful for debugging. I would like there to be a way make this optional.

PR #557, opened 2017-10-29
OG's last physics checkin

Mag-nus:


ogmage78:
I was going to delete my fork but I realized this is pointing to it.   I was just getting started on this and it is not in a merge in state.   As of my last check in, it does not compile.   It is really just some of the helper methods and none of that actual physics code.   I had not gotten that yet to port.  I don't know this will do anyone much good, but I will not delete my fork until this is closed.   I would have never PR'd something in this state - it is a mess and not reflective of my finished work product.   Thanks

PR #560, opened 2017-11-01
fixed web project auth to actually work.  *shrug*

Verbal-AC:
no idea where this was going, but at least the auth works now.  also updated nuget packages.


fantoms:
Auth login was successful in my tests.

PR #565, opened 2017-11-13
added reading of xp sharing in fellowship creation

Verbal-AC:


Source/ACE/Network/GameAction/Actions/GameActionCreateFellowship.cs
```diff
@@ -17,8 +17,9 @@ public static void Handle(ClientMessage message, Session session)
             // todo: option for sharing XP
```
LtRipley36706: Guess this is no longer a TODO, eh? ;)

PR #566, opened 2017-11-13
Robustify API Hosts

LtRipley36706:
* Removed AnyCPU configuration.
* Removed second Database.Start from ACE main program.
* Changes made to Api.Host and AuthApi.Host to handle automatically fixing issues with urlacl and firewall.


Source/ACE.Api.Host/Program.cs
```diff
@@ -12,15 +15,127 @@ public static void Main(string[] args)
...
+                try
+                {
+                    INetFwPolicy2 firewallPolicy = (INetFwPolicy2)Activator.CreateInstance(Type.GetTypeFromProgID("HNetCfg.FwPolicy2"));
+                    INetFwRule firewallRule = firewallPolicy.Rules.OfType<INetFwRule>().Where(x => x.Name == fwRuleName).FirstOrDefault();
+
+                    if (firewallRule != null)
+                    {
+                        Console.WriteLine(firewallRule.Name + " rule has been found in Firewall Policy");
+                        fwRuleFound = true;
+                        fwCommand = "set";
```
fantoms: `fwCommand = "set";` will this be used if fwRuleFound is true?

LtRipley36706: Yeah, I had anticipated that it would be needed to use "set" if we're modifying the found rule in the firewall policy but hadn't thought of a specific use case for it until...

LtRipley36706:
@fantoms Added port checking to the code so if the port is changed the rule would be adjusted as well should the user click yes to the UAC prompt. 

Thanks for the idea! I figured we'd need the "set" command, just wasn't sure where yet.

ddevec:
I'm pretty hesitant to support this change in ACE.

I'm not sure its the web-server's job to manipulate your firewall at run time.  I would <u>strongly</u> prefer my web server didn't manage my firewall for me.  At minimum this change should be optional.

There are also pretty significant security implications to modifying the firewall, and we need to ensure we do so absolutely properly if we include this code into our server.   The code as written opens this port for all applications, it should open it only for the server itself.  We should probably talk to someone knowledgeable in security to determine best practices here.

Finally, this code is very windows specific, and I believe were attempting to keep ACE as platform agnostic as possible.  I'm unsure if including this in our mainline source really fits with our design philosophy.

PR #568, opened 2017-11-20
EquipTest Updates

OptimShi:
Changed EquipTest debug command to utilize the index of a clothing table item and added an optional shade parameter

### DIFF
--- a/Source/ACE/Command/Handlers/DebugCommands.cs
+++ b/Source/ACE/Command/Handlers/DebugCommands.cs
@@ -18,8 +18,8 @@ namespace ACE.Command.Handlers
 {
     internal enum TestWeenieClassIds : uint
     {
-        Pants        = 120,
-        Tunic        = 134,
+        Pants = 120,
+        Tunic = 134,
         TrainingWand = 12748,
         ColoBackpack = 36561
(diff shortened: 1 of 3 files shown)


Source/ACE/Command/Handlers/DebugCommands.cs
```diff
@@ -18,8 +18,8 @@ namespace ACE.Command.Handlers
 {
     internal enum TestWeenieClassIds : uint
     {
-        Pants        = 120,
-        Tunic        = 134,
+        Pants = 120,
```
ddevec: I don't understand why you shifted these to non indented.  We've been keeping the assignments aligned on enums, I dislike the inconsistency this adds.

OptimShi: *I* didn't change this purposefully, I think Visual Studio must've made this change for me when I did something. Weird. Fixed it, anyways.

fantoms: VS has been changing this automatically for me as well, if you use the CTRL+K and then CTRL+D shortcut, it will show you the same results...

PR #569, opened 2017-11-26
Make Use Wield/Shop Lists

LtRipley36706:
Add AceObjectInventory class and make Vendors/Creatures/WO use it for shop/wield lists.

The WieldList is more hack than viable code likely but it does visually make the world closer to correct, and sets up a foundation for the objects having lists of things they need to generate.


Source/ACE/Entity/Vendor.cs
```diff
@@ -205,14 +191,14 @@ public void BuyValidateTransaction(ObjectGuid vendorid, List<ItemProfile> items,
             // calculate price. (both unique and item profile)
             foreach (WorldObject wo in uqlist)
             {
-                goldcost = goldcost + (uint)Math.Ceiling(SellRate * (wo.Value ?? 0) * (wo.StackSize ?? 1) - 0.1);
+                goldcost = goldcost + (uint)Math.Ceiling((SellPrice ?? 1) * (wo.Value ?? 0) * (wo.StackSize ?? 1) - 0.1);
```
ddevec: Is 1 really the right default if SellPrice is undefined?

Perhaps we should print a warning if SellPrice is undefined, as I don't think there is a good default.

Source/ACE/Entity/Vendor.cs
```diff
@@ -253,7 +239,7 @@ public void SellItemsValidateTransaction(Player player, List<WorldObject> items)
             foreach (WorldObject wo in items)
             {
                 // payout scaled by the vendor's buy rate
-                payout = payout + (uint)Math.Floor(BuyRate * (wo.Value ?? 0) * (wo.StackSize ?? 1) + 0.1);
+                payout = payout + (uint)Math.Floor((BuyPrice ?? 1) * (wo.Value ?? 0) * (wo.StackSize ?? 1) + 0.1);
```
ddevec: Same thing (about default).

Source/ACE/Factories/WorldObjectFactory.cs
```diff
@@ -30,6 +31,18 @@ public static List<WorldObject> CreateWorldObjects(List<AceObject> sourceObjects
                     WorldObject wo = CreateWorldObject(aceO);
                     if (wo != null)
                         results.Add(wo);
+                    // TODO: this is a hack job. Remove this and do it right. 
+                    if (wo.WieldList.Count > 0)
```
ddevec: I know this is in a todo hackjob, but realize that this if statement is unneeded (if the list is empty the for loop does nothing), and it generates the list (an O(n) operation) an extra time, in general this pattern should be avoided.

PR #577, opened 2017-12-06
Account name check at creation

Nalindar:


Rawaho:
This isn't thread safe, accounts can be created on multiple network threads as well as the separate command thread, in the time between checking and creating the account another could have been created.
The database should be checking for constraint violations not the core.

Nalindar:
yup, wasn't meant to be an end-all fix :p kill the pr if you'd like, just learning the process.

Rawaho:
I'm not just singling out this PR, the same thread safety issue arises in the authentication handler as well where accounts can also be created without the REST.

PR #579, opened 2017-12-10
ACE switched over to .net core 2.0

Mag-nus:
ACE.Database\Extensions\DataTableExtensions.cs has been changed. Comments have been added and the new manual method is now the only method used when loading data.

ACE.Database\Database.cs added a debug message before and after the time consuming InitializePreparedStatements()


Mag-nus:
I did not test the secure auth stuff. Can someone please test and confirm that still works?

PR #581, opened 2017-12-19
Generators generated

LtRipley36706:
* Overhauled Generator system to bring it more in line with how it operated on retail servers.
  * You'll need to load some test data included with this update in the Updates/World folder for this code to work.
  * Locations of the test generators are below. The creature generators do not support regeneration, while the item generators will regenerate.
  * /teleloc 0x7f0301ad [12.319899559021 -28.482000350952 0.0049999998882413] 0.33894598484039 0 0 0.94080585241318
  * /telepoi Holtburg

This is by no means the complete system but I think its a good step towards one than what is currently in ACE. Lots more to do no doubt.


Source/ACE/Entity/GenericObject.cs
```diff
@@ -1,10 +1,17 @@
...
+
+namespace ACE.Entity
 {
     public class GenericObject : WorldObject
     {
         public GenericObject(AceObject aceObject)
             : base(aceObject)
         {
+            if (AnimationFrame == null)
+                AnimationFrame = 101;
```
ddevec: What does this mean?  Can you put in a constant, or enum value instead, to make this clear?

Source/ACE/Entity/Landblock.cs
```diff
@@ -843,6 +843,17 @@ private void ItemTransferContainerInternal(ObjectGuid woGuid, Container containe
             wo.Location = null;
             wo.InitializeAceObjectForSave();
             container.AddToInventory(wo, placement);
+
+            // Was Item controlled by a generator?
+
+            if (wo.GeneratorId > 0)
```
ddevec: Should this be a function of the landblock, or a function of the item?

In other words, is it the landblock's responsibility to know that generators exist, and that objects may have generators, or is it the object's responsibility to notify the generator on pickup.  (And the landblock's responsibility to notify the item it was picked up).

IMO landblocks should have objects and be oblivious to what they are, or if they have generators.

LtRipley36706:
Refined the generator system to make it easier to add content/generators and use existing database tables (with a minor alteration). This linking system also uses the guid generation system to avoid collisions.

Source/ACE/Entity/GenericObject.cs
LtRipley36706: Need to apply some researched data to this, will fix shortly.

Source/ACE/Entity/Landblock.cs
LtRipley36706: More thinking to be done here, a TODO perhaps?

LtRipley36706:
Addressed AnimationFrame aka Placement

PR #582, opened 2017-12-28
Tier 1 loot

dgatewood:
WIP Tier 1 loot, new classes for missing loot item, additions to AceObject and WorldObject for variables that needed to be added, command added in commandHandler, etc.  Much more work needs to be done to added all the necessary variables to WorldObject.

This should be discussed. I feel that my coding skills may not be on par with most others. I felt the million if/else statements and rnd numbers was excessive, but didn't know how else to do it since all items had different ranges for their stats, etc.

### DIFF
--- a/Source/ACE/Factories/WorldObjectFactory.cs
+++ b/Source/ACE/Factories/WorldObjectFactory.cs
@@ -3,6 +3,7 @@ using ACE.Entity;
 using ACE.Entity.Enum;
 using ACE.Managers;
 using System.Collections.Generic;
+using System;
 
 namespace ACE.Factories
 {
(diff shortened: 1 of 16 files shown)


kdmxp256:
Corpse creation should probably be moved to its own WO class that is derived from Container, instead of being a method within the WorldObjectFactory class.

Source/ACE/Factories/WorldObjectFactory.cs
```diff
@@ -83,11 +84,30 @@ public static WorldObject CreateWorldObject(AceObject aceO)
...
 
+        public static WorldObject CreateCorpse(Creature cs)
+        {
+            ////Console.WriteLine("Inside CreateCorpse");
+            WorldObject wo = CreateNewWorldObject(1217);
+            ////Console.WriteLine("Added location to corpse");
+            wo.Location = cs.Location;
+            return wo;
+        }
+
```
kdmxp256: Corpse creation should probably be moved to its own WO class that is derived from Container, instead of being a method within the WorldObjectFactory class.

Source/ACE/Factories/WorldObjectFactory.cs
```diff
@@ -83,11 +84,30 @@ public static WorldObject CreateWorldObject(AceObject aceO)
...
+                case WeenieType.Caster:
+                    return new Caster(aceO);
                 default:
                     return new GenericObject(aceO);
             }
         }
 
+        public static WorldObject CreateCorpse(Creature cs)
+        {
+            ////Console.WriteLine("Inside CreateCorpse");
```
kdmxp256: Corpse creation should probably be moved to its own WO class that is derived from Container, instead of being a method within the WorldObjectFactory class.

PR #584, opened 2018-01-09
First round of 16PY fixes

LtRipley36706:
These fixes based on obvious issues as of version 0.0.2 of 16PY.

* Corrected issue with objects spawning at xxxx0000 in landblock.
* Added expected object flags to currently implemented WeenieType classes.
* Adjusted Bool recalls for WorldObjects.
* Added Shade (and TryToBond, which appears unimportant) to Inventory.
* Added Shade and palette to create command for testing.
* Added ClothingBase and Palette lookup for object spawning so they look correct.
* Added proper wielding for clothing on weenies and fixed clothing priority for them.
* Added face randomization for Human based NPCs.

version 0.0.3 of 16PY to follow shortly with shade data included for inventory


LtRipley36706:
No idea why the test is failing.. locally they all check out.

PR #585, opened 2018-01-10
GiveObjectRequest, HearDirectSpeech, and HandleGiveObjectRequest

dgatewood:
The code for the collector is extra and can be taken out if necessary. I just thought it would be cool for the meantime while we are waiting on the emote tables.  People like to see a little flair.


Mag-nus:
I believe I merged your other PR out of sequence.

This PR seems to cover some of what the other PR covered. Is this one good to merge, or, does it need to be reworked?

PR #589, opened 2018-01-18
Update Licence

Cydrith:
GPL V3 to AGPL V3


Mag-nus:
I think these lines should be populated:
+    <one line to give the program's name and a brief idea of what it does.>
+    Copyright (C) <year>  <name of author>

Cydrith:
Can do. Left them that way because they were previously that way in the old licence (they're part of the example area).

PR #598, opened 2018-01-24
Add back Shade and TryToBond

LtRipley36706:
Let's see what this does shall we?


LtRipley36706:
only took 17 commits here plus the others in two PRs to realize i goofed on setting up a table in the database properly with the sql scripts install... terrible.

PR #600, opened 2018-01-24
Typo with that last one

dgatewood:
Reworked GameEventMagicUpdateEnchantment. The way it was currently trying to pack spell data was garbling things up and the packet no longer matched a retail packet. Once I rewrote it, it now matches retail and also shows all spells in use in the active spell tab in the top left corner, with time remaining(client side).  This will be implemented server side soon.

To demonstrate this, I implemented non contract OnUse for gems. You can use Blackmoor's Favor, or the two gems that are available for sale at the gem shop in Holtburg.


dgatewood:
will try to reopen as a fork/branch when I have time to mess with it

PR #602, opened 2018-01-26
UpdateEnchantment/Gem OnUse

dgatewood:
I used TortoiseGit to create branch/master. I tried it by git commands only, but it kept telling me no such git directory... Checked it multiple times. I hope I am getting close with Git in general.

Reworked GameEventMagicUpdateEnchantment. The way it was currently trying to pack spell data was garbling things up and the packet no longer matched a retail packet. Once I rewrote it, it now matches retail and also shows all spells in use in the active spell tab in the top left corner, with time remaining(client side). This will be implemented server side soon.

To demonstrate this, I implemented non contract OnUse for gems. You can use Blackmoor's Favor, or the two gems that are available for sale at the gem shop in Holtburg.


dgatewood:
I will change this later on tonight and fix the appveyor issues. Didn't check it last night before bed

spazmodica:
As this feature is for the Gem to be casting the spell, GameEventMagicUpdateEnchantment need to also have a casterGuid parameter to have the Gem's guid vs the player guid vs another player's guid who is casting the spell.

Message for using Blackmoon Gem is "The gem casts Blackmoorï¿½s Favor on yo".

PR #609, opened 2018-01-31
Update README.md

w4nn3s:
spacing around links


Mag-nus:
I too am curious as to the reasoning

w4nn3s:
I didn't notice the "mastering-markdown" content.
I proposed this change to make these easily double-click selectable etc.

PR #633, opened 2018-02-10
WIP: Switch Database system to Entity Framework Core, DB First

Mag-nus:
This would simplify our DB model.
It should make it easier for new developers to get involved.
It would allow us to use async/await database queries.

Using the database first method, which this shows, we would simply design our tables in MySQL, then scaffold those changes into ACE. The scaffolding will completely rebuild/update the classes you see in ACE.Database.Models.

Scaffolding is done in visual studio, and is only needed if you make a database schema change.
Visual Studio -> View -> Other Windows -> Package Manager Console

Scaffold-DbContext "server=localhost;port=3306;user=root;password=password;database=ace_auth" -Project ACE.Database -OutputDir ".\Models\ace_auth" -f Pomelo.EntityFrameworkCore.MySql

Scaffold-DbContext "server=localhost;port=3306;user=root;password=password;database=ace_world" -Project ACE.Database -OutputDir ".\Models\ace_world" -f Pomelo.EntityFrameworkCore.MySql

Scaffold-DbContext "server=localhost;port=3306;user=root;password=password;database=ace_shard" -Project ACE.Database -OutputDir ".\Models\ace_shard" -f Pomelo.EntityFrameworkCore.MySql

The scaffolding process will automatically create all of the POCOs. In the existing system, they need to be manually created in ACE.Entity with appropriate class and field attributes for mapping. Using Entity Framework needs none of that.

If you're wondering why I chose to use Pomelo instead of a MySql EF package from Oracle, the Oracle packages do not scaffold properly. Pomelo seems to be the go to framework for EF in .net core. I've used it before with success.

Lastly, I believe using Entity Framework will expose contributors to the project to patterns and practices that they're more likely to see out in the field.

I forgot to mention, EF does not support database views. We'll need to construct those objects in a different way. It's unlikely we even need views for performance at our scale.


LtRipley36706:
I fixed the conflicts and switched the branch this PR is targeting. I'll hold off on the merge until you're ready to submit and re-fork to get setup on the new branch for working on for collaboration

Issue #91, opened 2017-02-25
StatusMessageType1 and StatusMessageType2 values need descriptive names


Mag-nus:

Mag-nus:
These files need more descriptive values:
https://github.com/ACEmulator/ACE/blob/c4a3da0b10c110c5b37530dfd82982429605b585/Source/ACE.Entity/Enum/StatusMessageType1.cs
https://github.com/ACEmulator/ACE/blob/c4a3da0b10c110c5b37530dfd82982429605b585/Source/ACE.Entity/Enum/StatusMessageType2.cs

        /// <summary>
        ///  is too busy to accept gifts right now.
        /// </summary>
Enum_001E = 0x001E,

Might become:
IsTooBusyToAcceptGifts = 0x001E,

Mag-nus:
This has been completed by Shawarma-Golem.

PR #637, opened 2018-02-11
StatusMessageType1 Descriptive Names for Issue #91

shawarma-golem:
- Added descriptive names to the first set of Enums for in-game messages to address Issue #91.
- These should allow developers to quickly identify the message represented.
- Many of these were kept to ~40 characters or less, following the nomenclature of the Enums named previously.


Mag-nus:
Looks like there's a few references that need to be fixed now that some of the enum values have been renamed.

Do you need help with that?

PR #641, opened 2018-02-16
Have Landblocks load all terrain, building, and generic obstacle data during start up and store it for use in the physics engine.

gmriggs:
This PR is for loading the landblock objects / obstacles, buildings, and landblock weenies stored in the database

A lot of the code from the last PR in Landblock.cs and Mesh.cs has been moved to a new LandblockMesh.cs class. Mesh.cs is now a base class for both the triangle-based landblock meshes, and the new ModelMesh.cs. Model meshes are polygonal based, and have most of their data stored in a StaticMesh object, which points to a global cache for efficiency.

https://www.youtube.com/watch?v=X4c7VHJ6DHE


kdmxp256:
Missing Changelog

PR #675, opened 2018-03-06
Entity Framework Core 2.0, Database First

LtRipley36706:
Just about ready for merging with master...

As a result of the rebase with current master the following error was introduced:

I noticed an issue with line 134 of Landblock.cs
`LandblockInfo = DatManager.CellDat.ReadFromDat<LandblockInfo>((uint)Id.Landblock << 16 | 0xFFFE);`

This line seems have an issue loading data from the dat file, I'm not 100% sure I know what the fileid its trying to load but seems what is there currently is incorrect.

`@telepoi Arwic` results in the following errors for example
```
2018-03-06 05:26:41,077 INFO : Unable to find object_id 3316318206 in Cell
2018-03-06 05:26:41,097 INFO : Unable to find object_id 3349807102 in Cell
2018-03-06 05:26:41,118 INFO : Unable to find object_id 3349741566 in Cell
2018-03-06 05:26:41,138 INFO : Unable to find object_id 3349872638 in Cell
2018-03-06 05:26:41,205 INFO : Unable to find object_id 3333095422 in Cell
```

These could be harmless and expected errors as maybe unlike CellLandblock files, there can be landblocks with LandblockInfo that don't load because they don't exist.. I manually tested 0xCE95FFFE and it loaded without error. Hopefully someone can confirm what is going on here before merging.


Mag-nus:
I checked the landblocks that throw the warnings. If you go to them in world, you'll find that there are no buildings in that space. I believe that's why no meshes exist for those blocks.

The datloader code is working correctly and loading meshes for arwic.

Mag-nus:
I just checked in a change that will move those log events from info level to debug level which is more appropriate for that file type.

PR #692, opened 2018-03-19
Added NameFilter to Character Creation

CrimsonMage:
Created TabooTable.json (Easily altered by users) to house the names of the filter list.

BannedWords.cs to handle the data of the json for the words

CharacterHandler.cs now includes a NameBanned message that will appear when someone tries to use a name from the list.

--Updated Changelog


LtRipley36706:
As discussed in discord, pending review by @Mag-nus , I'd like to see if we can't get the datreader for that file so as to not include it due to its content in the repo. Absent that, *encrypting* the strings to censor them would make me a lot more comfortable with it. Ideally combining the datreader + table in ace_world db for adding future bad names (Asheron, Martine, etc..) or as a way for server operators to reserve names would be pretty cool

Mag-nus:
I've added the tabootable for you. This should make it a bit easier.

My implementation for taboo table respects the wild card characters. You may want to use this same functionality in the json list as well.

CrimsonMage:
Removing.

PR #764, opened 2018-04-27
Can we get some loot?

dgatewood:
This creates working random loot for all tiers. There are some things that need to be added, or tweaked, but this is very much closer to being finished than not.

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -1,14 +1,23 @@
 using System.Collections.Generic;
-
+using System;
 using ACE.Database;
 using ACE.Entity;
 using ACE.Entity.Enum;
+using ACE.Server.Entity;
 using ACE.Server.WorldObjects;
+using ACE.Server.Managers;
+using ACE.Server.Network.Sequence;
+using ACE.Database.Models.Shard;
+using ACE.Database.Models.World;
+using ACE.Entity.Enum.Properties;
+using ACE.Factories;
 
 namespace ACE.Server.Factories
 {
     public class LootGenerationFactory
     {
+        // This is throw away code to understand the world object creation process.
+
         public static void Spawn(WorldObject inventoryItem, Position position)
         {
             throw new System.NotImplementedException();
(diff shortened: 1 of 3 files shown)


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -35,5 +44,18050 @@ public static List<WorldObject> CreateRandomObjectsOfType(WeenieType type, int c
...
+            }
+            return wo;
+        }
+
+        public static WorldObject CreateTier1Objects()
+        {
+            WorldObject wo;
+            int id = 0;
+            int chance;
+            Random r = new Random();
```
gmriggs: Depending on how quick this new Random() line is called, creating new instances of Random could raise some issues

Once in a previous project I was creating a new instance of Random everytime i generated a new random #, and it turned out to be a big issue. If the code was called fast enough, basically it would return the same random # a bunch of times in a row

Might not be an issue here, but it's something to be aware of for sure!

In the Physics.Common namespace there is a Random class that has a static instance of System.Random, and it is marked with the [ThreadStatic] attribute in an attempt to make it thread safe.

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -35,5 +44,18050 @@ public static List<WorldObject> CreateRandomObjectsOfType(WeenieType type, int c
...
+                            //    {
+                            //        //glyph of coordination
+                            //        id = 37350;
+                            //    }
+                            //    else if (chance == 121)
+                            //    {
+                            //        //glyph of corrosion
+                            //        id = 37342;
+                            //    }
+
```
gmriggs: These arrays will be created each time the method is called, and should ideally be marked as static

A list of ID numbers in the code should also eventually be a mapping to what these items actually are so they can be referenced easier

gmriggs: A lot of data-driven code here... the armorWeenie/armorPieceType/spellArray/cantripArray set in particular seems to be quite extensive.

This all makes perfect sense initially, but eventually the core algorithms should be located in one place, with the data being fed into it

The armorWeenie/armorPieceType/spellArray/cantripArray in particular could maybe be a structure, with some functions that operate on this structure. For example, all of these if statements could be calling a function which populates this structure with the data for that branch

Source/ACE.Server/Factories/LootGenerationFactory.cs
Mag-nus: Locking variables is also costly, and blocks other threads as well.

In this case I'd just instantiate a new random every time like he has it here. If, after profiling, we determine it is an issue, we could then optimize it by having a Random object per landblock and passing that as an argument, but, I don't think that will be necessary.

Mag-nus:
I really didn't dig in with a fine tooth comb, so my minimal feedback is this:

First, awesome work. This will be fun to try out!

LootGenerationFactory.cs will need to be refactored to trim it down from 18k lines. You could consider using partial classes for each tier. As a long term solution though, I believe we'll be able to put a lot of this logic in the database to help drive functions and formulas. That should help reduce hard coded numbers.

Source/ACE.Server/Factories/LootGenerationFactory.cs
LtRipley36706: I think the behavior @gmriggs saw is actually in full effect in our encounter system currently. I've found often enough that it irks me that if I load into a landblock and a rare mob is selected, you'll find that same mob spawned in several generators. Certainly, it feels like something is off there.

PR #769, opened 2018-04-28
Implement the Attuned functionality

kdmxp256:
- Add new GameEvent message for InventoryServerSaveFailed
- Add attuned checking to HandleActionDropItem, with InventoryServerSaveFailed message response

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -480,11 +480,38 @@ namespace ACE.Server.WorldObjects
             }).EnqueueChain();
         }
 
+        private bool IsAttuned(ObjectGuid itemGuid, out bool isAttuned)
+        {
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -480,11 +480,38 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
             }).EnqueueChain();
         }
 
+        private bool IsAttuned(ObjectGuid itemGuid, out bool isAttuned)
+        {
+            WorldObject item = GetInventoryItem(itemGuid);
+            if (item == null)
+            {
+                isAttuned = false;
+                return false;
```
gmriggs: Should it be the responsibility for the main return value of IsAttuned to be if the item is inventory?

It's cool that IsAttuned checks for this, but if the calling function needs to differentiate between an item not being in inventory / unattuned items, maybe there should be a separate check in the calling function?

PR #773, opened 2018-04-29
Can drop things on the ground once again!

dgatewood:
Nullable value was being cast as an int, but was not checked to see if the value was null, only if the object was null. This fixes that so that items can be removed from inventory once again.


kdmxp256:
Good catch on the null exception for the missing Attuned property on some items; however, the first null check on GetInventoryItem() should remain, as that method may return a null, also, which would need to be handled correctly.

PR #787, opened 2018-05-10
Bolt, Streak and Arc updates

Slushnas:
Various bits of code have been added to make the existing spells line up with retail data and lay the groundwork for other spells. There is still more research to do but this is a step closer to retail based on the data I have gathered so far.

I am not sure if the code I added to WorldObjectFactory.cs is the best way to get the spell ID into the constructor (seems like one of the CreateWorldObject functions could be removed/integrated).

I had to add some code to be able to override some PhysicsDescription fields. I did a couple of searches through the pcaps and the fields that I set to false by default don't seem to have been used so I don't believe this will cause any problems.

### DIFF
--- a/Source/ACE.Server/Factories/WorldObjectFactory.cs
+++ b/Source/ACE.Server/Factories/WorldObjectFactory.cs
@@ -21,7 +21,7 @@ namespace ACE.Server.Factories
         /// <summary>
         /// A new biota be created taking all of its values from weenie.
         /// </summary>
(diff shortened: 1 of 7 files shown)


kdmxp256:
Probably, would be a good idea to change the SpellID class variable, in SpellProjectile, to private, as now it shouldn't need to be publicly accessible outside the class.

Slushnas:
I've reverted the world object factory changes and moved some code into a setup function.
This also includes a fix for the wielded item placement showing up incorrectly on login.

PR #790, opened 2018-05-12
@nudge @nudge, got some @crack

fartwhif:
Use /crack to unlock/pick/open doors and chests. (don't do drugs)
Use /nudge to correct player position cell ID after teleporting into black space. (experimental, debug)


gmriggs:
Tested everything, looks good

PR #791, opened 2018-05-14
lockpicking doors

fartwhif:


gmriggs:
Looks good for the success case

For the non success case, no text message is sent, and client gets perma hourglass

Still should be good to merge...

PR #792, opened 2018-05-15
Adding collector EmoteTable stuff..

dgatewood:
Added HandleReceive inside of WorldObject so the emote table can be checked to see if the item given should set off any emoteActions.
 And Minor fix for casters in Loot Factory, casters had no current mana field, this adds that.

### DIFF
--- a/Source/ACE.Server/Managers/EmoteManager.cs
+++ b/Source/ACE.Server/Managers/EmoteManager.cs
@@ -873,6 +873,7 @@ namespace ACE.Server.Managers
 
         public void ExecuteEmote(BiotaPropertiesEmote emote, BiotaPropertiesEmoteAction emoteAction, ActionChain actionChain, WorldObject sourceObject = null, WorldObject targetObject = null)
         {
+            var player = targetObject as Player;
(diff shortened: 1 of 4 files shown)


Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -943,11 +945,54 @@ public void ExecuteEmote(BiotaPropertiesEmote emote, BiotaPropertiesEmoteAction
...
                         player.Session.Network.EnqueueSend(new GameMessageHearDirectSpeech(sourceObject, emoteAction.Message, player, ChatMessageType.Tell));
                     });
                     break;
 
+                case EmoteType.TurnToTarget:
+                    var creature = sourceObject is Creature ? (Creature)sourceObject : null;
+                    actionChain.AddDelaySeconds(creature.Rotate(player));
+                    actionChain.AddAction(sourceObject, () =>
+                    {
+                        creature.Rotate(player);
```
gmriggs: Is this double-rotate intended?

With 2 calls to rotate, the server will be sending the TurnTo MovementEvent to the client, waiting until its complete, and then sending another TurnTo packet

PR #793, opened 2018-05-15
EmoteTable Stuff

dgatewood:
Added HandleReceive in WorldObjects
Checking within PlayerInventory if a HandleReceive is true.
Added new cases inside of ExecuteEmote.
Thiss allows all collectors and rng hand-ins, such as GamesMasters, to work.


gmriggs:
Tested, works great!

PR #798, opened 2018-05-20
@addallspells dev command

mcreedjr:
Be gentle... I am a novice developer at best, and this is my first pull request ever, let alone first pull request on the ACE project.

I humbly submit a developer command to be used to learn all known spells on the current character.

I acknowledge this results in spells being learned that will never be castable by the player, and I also note that it asks the client to display way too many 'bubble' messages upon learning each spell, but the client seems to handle it more or less gracefully.

I had thought about learning all spells, except for the last spell with the AddKnownSpell() Method, and then just calling LearnSpellWithNetworking() on the last spell to update the client, but that seemed more hacky than having the client ingest lots of updateSpellEvent GameEvents.

Feedback is welcomed and encouraged.

### DIFF
--- a/Source/ACE.Server/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE.Server/Command/Handlers/AdminCommands.cs
@@ -855,6 +855,21 @@ namespace ACE.Server.Command.Handlers
             }
         }
 
+        // addallspells
+        [CommandHandler("addallspells", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0, "Adds all known spells to your own spellbook.")]
+        public static void HandleAddAllSpells(Session session, params string[] parameters)
+        {
+            var spells = DatManager.PortalDat.SpellTable.Spells;
+
+            foreach(var spell in spells)
+            {
+                uint spellId = spell.Key;
+
+                if (Enum.IsDefined(typeof(Network.Enum.Spell), spellId))
+                    session.Player.LearnSpellWithNetworking((uint)spellId);
+            }
+        }
+
         // adminhouse
         [CommandHandler("adminhouse", AccessLevel.Admin, CommandHandlerFlag.RequiresWorld, 0)]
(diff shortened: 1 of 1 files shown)


gmriggs:
hey mcreedjr,

Thanks for the awesome PR!

This is definitely a very useful admin command, and should help a lot with testing :)

I've considered about adding a similar command, as well as something to help with managing spell components... ie., something to add components for all known spells, or perhaps a way to disable the spell component requirement when casting spells

This PR surely seems great to me, and I have set it to approved for merging

Some additional notes:

- Should the command add spells for magic schools the player is not trained in? I guess a player could always train additional magic skills later, so this should still be fine

- When running this command, I started to receive WARN : [Unverified] Packet # received again messages in the log. I had to disconnect the character by closing the client, and restart to fix the messages.

I assume this is an issue with the networking code, and not anything with the new code... perhaps something to do with sending such a large # of packets?

Just some notes, everything in the PR still looks great to me :)

Thanks for the great contribution!

mcreedjr:
@gmriggs Thank you for the prompt and thorough response. I can't thank you enough for the contributions you've made to the project, and I am humbled that you would offer such an in depth analysis of my arguably simple and hacky pull request.

My thoughts on your inquiries are below:

- Should the command add spells for magic schools the player is not trained in? I guess a player could always train additional magic skills later, so this should still be fine

Yes, that was my intent for the exact reason you mentioned. I did not want admins/devs to have to run the command multiple times as new schools were trained. If you feel this is overzealous, I can add logic to only learn spells for trained schools.

- When running this command, I started to receive WARN : [Unverified] Packet # received again messages in the log. I had to disconnect the character by closing the client, and restart to fix the messages.

I, too, received this messages and assumed the same. To be candid, I did not see that the client functionality was impacted after running the command, save for the plethora of WARN messages. It was not my intent to have to have users re-log the client to continue playing after using the command.

Again, thanks for the prompt and in depth response!

Issue #783, opened 2018-05-09
Player specific tables need to be moved from biota into character_


Mag-nus:
There are a handful of tables that only populate for players, ie: shortcut bar, friends, spell bar, etc..

These tables shouldn't be in the biota. Instead they should be moved to the character_ tables.

In addition, there are a few propertyints that were shoved into the PropetyInt list that only represent character info. These properties should also be moved to the character table.

Issue #782, opened 2018-05-09
Shortcuts can cause db exceptions with foreign keys


Mag-nus:
### Bug Report or Feature Request (mark with an `x`)
```
- [X ] bug report -> please search issues before submitting
- [ ] feature request
```

2018-05-09 05:57:07,618 ERROR: SaveBiota failed with exception: Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> MySql.Data.MySqlClient.MySqlException: Cannot add or update a child row: a foreign key constraint fails (ace_shard.biota_properties_shortcut_bar, CONSTRAINT wcid_shortcut_objectId FOREIGN KEY (shortcut_Object_Id) REFERENCES biota (id) ON DELETE CASCADE ON UPDATE NO ACTION) ---> MySql.Data.MySqlClient.MySqlException: Cannot add or update a child row: a foreign key constraint fails (ace_shard.biota_properties_shortcut_bar, CONSTRAINT wcid_shortcut_objectId FOREIGN KEY (shortcut_Object_Id) REFERENCES biota (id) ON DELETE CASCADE ON UPDATE NO ACTION)

I think for the shortcut bar, it will make sense to remove the key constraint.

It's not critical if an item is deleted but it's shortcut is not. if it's not critical, the key constraint is not required.

Will have to test the scenario where a user adds a short cut, deletes the item, and relogs. Does the client still get sent the old item id? What does the client show? What happens to the shortcut list when a user drops a new item onto the bar?

I have a theory that we won't even have to manage removal of shortcut items when an item is removed from inventory (or destroyed). I think the client handles this all safely.

If that is the case (which I hope), we wouldn't have to check if an item exists as a shortcut at the time of inventory removal, which keeps this whole process very clean.

fartwhif:
To Reproduce at baff0fa

1. Go to Yaraq.
2. Buy an empty flask from the shopkeep.
3. Use the empty flask on the pool outside. *you get an additional item water, (not sure if this is supposed to consume the flask, unrelated?)*  - the guid for the new item is being duplicated? (just a hunch)
4. Put the empty flask you just used in the shortcut bar.
5. do /save-now, you get an exception (Exception 1)
6. do /save-now again, you get a different exception (Exception 2)

**Exception 1**
[...]

PR #799, opened 2018-05-20
Database Rebase and Table Rename

LtRipley36706:


Mag-nus:
Can you summarize the changes being made here?

Also, if we're making these changes, can we move the player specific tables over to character_ ?

Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -329,14 +329,14 @@ public void RemovePosition(PositionType positionType)
         {
             Positions.Remove(positionType);
 
-            if (Biota.TryRemovePosition(positionType, out var entity, biotaPropertiesPositionLock) && ExistsInDatabase && entity.Id != 0)
+            if (Biota.TryRemovePosition(positionType, out var entity, biotaPropertiesPositionLock) && ExistsInDatabase && entity.ObjectId != 0)
                 DatabaseManager.Shard.RemoveEntity(entity, null);
         }
```
Mag-nus: If we remove the auto generated id and now rely on the biota's id as the primary key for the record, this code will break.

entity.Objectid will always be non-zero as it's generated by ACE to put the object into the world.

entity.Id would be non-zero until it's saved to the db. At that point the db gives it a unique record id.

Mag-nus:
See the comment I made on WorldObject_Properties.cs

LtRipley36706:
This PR resolves #782 and resolves #783

Database/Base/WorldBase.sql
```diff
@@ -102,6 +153,456 @@ CREATE TABLE `points_of_interest` (
...
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+--
+-- Table structure for table `spell`
+--
+
+DROP TABLE IF EXISTS `spell`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `spell` (
```
Mag-nus: In this table, was the duplicate data that exists in the dat removed from the db?

There's no need to duplicate the data, and, in fact, we must not. The client will be pulling from the dat for those values and the server will need to as well.

Source/ACE.Database/Models/Shard/Character.cs
```diff
@@ -12,6 +12,11 @@ public partial class Character
         public bool IsDeleted { get; set; }
         public ulong DeleteTime { get; set; }
         public double LastLoginTimestamp { get; set; }
+        public int TotalLogins { get; set; }
+        public int CharacterOptions1 { get; set; }
+        public int CharacterOptions2 { get; set; }
+        public uint HairTexture { get; set; }
+        public uint DefaultHairTexture { get; set; }
```
Mag-nus: This looks good.

Database/Base/WorldBase.sql
LtRipley36706: I made no changes here. before removing the data, I would need know exactly what columns are duplicated and it would also make sense to note what differences, if any, there are to the exported data vs DAT data. I think the coding itself uses the combined data, starting with the DAT then moving over to the DB to fill in the rest but I'm not as well versed in this part as @kdmxp256 or maybe @gmriggs

PR #802, opened 2018-05-23
added SurfaceTexture parser

fartwhif:
added Texture file and Texture file metadata getter
if a few things are mastered it might come in handy

### DIFF
--- a/Source/ACE.DatLoader/FileTypes/SurfaceTexture.cs
+++ b/Source/ACE.DatLoader/FileTypes/SurfaceTexture.cs
@@ -1,0 +1,95 @@
+using ACE.Common;
+using ACE.DatLoader.Entity;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace ACE.DatLoader.FileTypes
+{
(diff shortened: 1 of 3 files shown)


Source/ACE.DatLoader/FileTypes/SurfaceTexture.cs
```diff
@@ -0,0 +1,95 @@
...
+            TextureCount = reader.ReadInt32();
+            if (TextureCount > 0)
+            {
+                TextureFileIds.Add(reader.ReadUInt32());
+                if (TextureCount > 1)
+                    TextureFileIds.Add(reader.ReadUInt32());
+            }
+
+            if (ConfigManager.Config == null)
+                ConfigManager.Initialize(Path.GetFullPath(@"..\..\..\..\..\ACE.Server\bin\x64\Debug\netcoreapp2.0\Config.json")); // weak
```
gmriggs: should this be in here?

fartwhif: Unfortunately there doesn't seem to be a kosher way to initialize things.   ACE.DatLoader.Tests, the parent caller in this case, has file paths even harder coded.

PR #814, opened 2018-06-03
implement server side request for retransmission

fartwhif:
fixed bug in Player_Vitals.cs causing a crash upon player disconnect

### DIFF
--- a/Source/ACE.Server/Network/NetworkSession.cs
+++ b/Source/ACE.Server/Network/NetworkSession.cs
@@ -245,7 +245,19 @@ namespace ACE.Server.Network
             uint issacXor = packet.Header.HasFlag(PacketHeaderFlags.EncryptedChecksum) ? ConnectionData.IssacClient.GetOffset() : 0;
             if (!packet.VerifyChecksum(issacXor))
             {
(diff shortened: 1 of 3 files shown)


Source/ACE.Server/Network/NetworkSession.cs
```diff
@@ -245,7 +245,19 @@ private void HandlePacket(ClientPacket packet)
...
             {
-                log.WarnFormat("[{0}] Packet {1} has invalid checksum", session.LoggingIdentifier, packet.Header.Sequence);
+                if (issacXor != 0)
+                {
+                    issacXor = ConnectionData.IssacClient.GetOffset();
+                    log.WarnFormat("[{0}] Packet {1} has invalid checksum, trying the next offset", session.LoggingIdentifier, packet.Header.Sequence);
+
+                    // not sure why there was a missed offset increment
+                    // results in a valid checksum
+                    // however, the client is now in a sort of "can't move" state.  The monsters react so the communication is still working, right?  fartwhif
```
fartwhif: It has come to my attention that a fragment is being lost, so client->server traffic dies because there's no mechanism for requesting that the missing fragment is retransmitted.  Also, the whole network layer needs to have support for multiple queues (Packet.Header.table) as to not confuse the client leading to the missing fragment.

Issue #753, opened 2018-04-20
We need a way to create/restore server gameplay/configuration options/values in the db


Mag-nus:
As we make more progress on the code, we're constantly coming across things that we're developing to replicate retail, but we know a server operator would prefer to tweak.

A couple examples:
- Player logout doesn't have to take 5 seconds. It could be configured to be 1 second (Integer)
- Bonus XP/Luminance (Float)
- Message Of The Day (String)
- Casting Impen/Bane on a player can act like an aura instead of affecting only equipped items. (Bool)

There are many many more things that players might want to configure. It would be great if they could just change a database value on a running server to have the change take effect.

We should have a DatabaseConfigurationManager (or similar) class that exposes all of these options.

Perhaps this class periodically (every minute, 5 minutes, ...?) checks the database for updated values.

We could even include an admin command to force resync of the variables.

We could even include admin commands to modify the variables from the command line, and from in-game.

Scribble1:
Drop items on death (Bool) 
Trophy Drop Rate Increases (Float or Int?)
Quest Timer Multipliers (Float or Int?)
Kill XP Multiplier (Float)
Quest XP Multiplier (Float)

deadlyelectron:
The basic impl can be found at #820. Now migration for stuff that should be properties but aren't need to be added, and once we have a base set of properties we can create a dump of the tables to place in the shard creation dump.

LtRipley36706:
#820 resolves this issue

PR #820, opened 2018-06-10
Add a configurable property storing database to resolve issue #753

deadlyelectron:
This PR adds a fourth (optional) database that contains 4 tables - `bool_stat`, `int_stat`, `float_stat`, and `string_stat`, one for each type of property that can be implemented in the current model. The original issue ( #753 ) mentions a few options that could initially be configured. I have implemented properties for the vitae penalty on death, the minimum vitae value, a flat XP modifier applied to both `GainXP` and `EarnXP` and a flat luminance modifier applied to `GainLuminance`.

The fourth database is completely optional, with the server warning an owner that they should use the SQL provided to create a new database but not forcing the server owner to deal with it. It pulls default values for properties from a `Dictionary` for each property type that is stored in the `DefaultPropertyManager` class.

The server caches values from the database and updates them on a 5 minute timer, where it persists the variables modified from the console or in-game into the database before pulling variables from the database. 

This PR also adds admin commands and console commands `fetch<Type>` and `modify<Type>` to work with the properties while the server is running from the game, as well as `resyncvariables` which will pause the timer and resync the database and cached variables. 

Since I'm not completely well acquainted with the codebase yet, I put in the properties I thought were the easiest for me personally to implement, but the process for implementing new properties is:

1. Locate and extract the property into a `private` variable (preferably `static` as well)
2. Create a unique name for the property
3. In the `DefaultPropertyManager` class, add a tuple into the appropriate `ToDict` call for the datatype the property is that has the default value
4. Replace the `get` method on the property with one calling `PropertyManager.Get<Type>(propertyName)`

The approach I used in developing this patch was to avoid user intervention temporarily, and as such there are tradeoffs involved.

1. The database is completely optional, which means that default values must be placed within the program to prevent a fallback value that doesn't make sense from being hit (instead of blowing the whole server up with a NPE). This can be solved by having the database be non-optional, or making the database non-optional after more properties are added to it.
2. There is now a fourth database ACE would need to run completely after this PR. There are a few potential fixes for this, including using a flatfile for the configuration values (removes the ability to update the backend to cause changes), rolling the 4 tables already implemented into one of the other ACE databases (I wasn't confident in my knowledge of how the process for updates works in order to do this but I'd be willing to learn if necessary) or just simply not implementing this feature right now.


gmriggs:
Hey deadlyelectron,

Thanks for the PR!

There is a lot to review here, but I will be checking everything out over the course of the next day or so :)

Thanks again for all the help, it is much appreciated :)

gmriggs:
Tested and looks good to me from a functionality and coding standpoint

The only issue I ran into was the Config.json.example file had to be manually updated to include the new Config database

Other than that, looks good to me!

@LtRipley36706 or @Mag-nus could maybe offer some better insights as to if the database implementation matches up with everything as expected

Thanks again, great work :)

Mag-nus:
Thank you so much for the effort! This is something we really need.

I haven't had a chance to review your files changed yet, but, what are your thoughts on changing int_stat to long_stat and float_stat to double_stat.

Also, I haven't confirmed the exact naming convention we use in ACE, but, how about also matching what we use in ACE, something like properties_bool, propeties_long, properties_double, properties_string?

deadlyelectron:
I can update int to long and float to double and rename the tables. 

@gmriggs The reason the Config.json.example was not updated with the new database is that the new database is actually optional - the server will run fine as is without it. However, I can easily change it to make the new database not optional and add the new database into the example config if that's what you want.

Source/ACE.Server/WorldObjects/Player_Xp.cs
```diff
@@ -4,19 +4,26 @@
...
+using ACE.Server.Managers;
 using ACE.Server.Network.GameMessages.Messages;
 using ACE.Server.WorldObjects.Entity;
 
 namespace ACE.Server.WorldObjects
 {
     partial class Player
     {
+
+        private static double XpModifier { get { return PropertyManager.GetDouble("xp_modifier"); } }
```
Mag-nus: I wouldn't use properties here.

XpModifier really isn't a property of Player.

Instead, I would either use a local variable when you want to use a property, or, just reference the property directly.

So, it would be something like this:
xpModifier = PropertyManager.GetDouble("xp_modifier");
xpToAdd = incomingXp * xpModifier;

- or -

xpToAdd = incomingXp * PropertyManager.GetDouble("xp_modifier");

Source/ACE.Server/Managers/EnchantmentManager.cs
```diff
@@ -35,12 +35,12 @@ public class EnchantmentManager
         /// <summary>
         /// The amount of vitae reduced on player death
         /// </summary>
-        public static readonly float VitaePenalty = 0.05f;
+        public static double VitaePenalty { get { return PropertyManager.GetDouble("vitae_penalty"); } }
```
Mag-nus: I also wouldn't use a property here.

We used a property here because we had this "magic" value, it needed to go somewhere. However, VitaePenalty really isn't a property of EnchantmentManager, but instead, a property of the shard.

Source/ACE.Database/DatabaseManager.cs
```diff
@@ -15,10 +15,13 @@ public static class DatabaseManager
 
         public static SerializedShardDatabase Shard { get; private set; }
 
+        public static ServerPropertyDatabase ServerConfig { get; private set; } = new ServerPropertyDatabase();
```
Mag-nus: I don't think this is needed anymore now that the config has been rolled into the Shard database.

Source/ACE.Database/Models/Shard/ShardDbContext.cs
```diff
@@ -39,12 +39,19 @@ public partial class ShardDbContext : DbContext
...
         public virtual DbSet<CharacterPropertiesTitleBook> CharacterPropertiesTitleBook { get; set; }
+        public virtual DbSet<ConfigPropertiesBoolean> ConfigPropertiesBoolean { get; set; }
+        public virtual DbSet<ConfigPropertiesDouble> ConfigPropertiesDouble { get; set; }
+        public virtual DbSet<ConfigPropertiesLong> ConfigPropertiesLong { get; set; }
+        public virtual DbSet<ConfigPropertiesString> ConfigPropertiesString { get; set; }
 
         protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
         {
-            var config = Common.ConfigManager.Config.MySql.Shard;
+            if (!optionsBuilder.IsConfigured)
```
Mag-nus: OnConfiguration will need to be manually reverted when you scaffold.

There should be no changes to this function as we add to the database.

Source/ACE.Database/ServerPropertyDatabase.cs
```diff
@@ -0,0 +1,220 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Storage;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace ACE.Database
+{
+    public class ServerPropertyDatabase
```
Mag-nus: I don't think this class is needed since we've rolled the config into the shard.

I have a few ideas that we can further discuss in discord.

PR #824, opened 2018-06-15
Give Request fix

dgatewood:
Selected Wielded Objects as well as Inventory Objects. This fixes crash of gioving wielded objects to the town crier.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -941,7 +941,7 @@ namespace ACE.Server.WorldObjects
         {
             
             WorldObject target = CurrentLandblock.GetObject(targetID) as WorldObject;
-            WorldObject item = GetInventoryItem(itemGuid) as WorldObject ?? GetWieldedItem(itemGuid) as WorldObject; ; ;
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -941,7 +941,7 @@ public void HandleActionGiveObjectRequest(ObjectGuid targetID, ObjectGuid itemGu
         {
             
             WorldObject target = CurrentLandblock.GetObject(targetID) as WorldObject;
-            WorldObject item = GetInventoryItem(itemGuid) as WorldObject;
+            WorldObject item = GetInventoryItem(itemGuid) as WorldObject ?? GetWieldedItem(itemGuid) as WorldObject;
```
gmriggs: Is it certain that GetWieldedItem() covers all of the possible scenarios?

Perhaps item should be checked for null after this line, and return if no item was found?

kdmxp256: I would agree that to be safe a null check should be added.

Warloxx01:
Hi,
Well it did stop the crash but it doesn't update the visible model in the game. They still look to be wearing the items. Also the Burden doesn't change. I tested it with the starter clothing and a Matty Robe.

PR #826, opened 2018-06-18
Unequip and burden correction

dgatewood:
Unwield and RemoveWithNetworking added/moved to the first thing done.


gmriggs:
looks good!

PR #829, opened 2018-06-23
Fix for Undead player characters not visually equipping clothing

OptimShi:


LtRipley36706:
great to see a PR from you again, @OptimShi ! 👍

PR #830, opened 2018-06-24
added sentinel command /buff [name]

fartwhif:
/buff - Buffs you (or a player) with all beneficial spells.
Usage: /buff [name]
This command buffs yourself (or the specified character).

curate the spell enum a bit (the other and self counterparts need to be exactly the same except for 'Self' and 'Other')


fartwhif:
thinking maybe those messages should be aggregated and enqueued in the same go

Issue #691, opened 2018-03-18
ACE doesn't restore the clients UI window positions


Mag-nus:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x ] feature request
```

### Repro steps.
Move some windows around, log out, log back in. You'll notice your window positions aren't restored to your new locations.

### Desired functionality.
Server should save/restore a users desired window positions/sizes.

### Mention any other details that might be useful.
The restore happens in GameEventPlayerDescription.cs
            /*if ((optionFlags & DescriptionOptionFlag.WindowLayout) != 0)
            {
            }*/

I don't recall where the save happens, but you might need to modify that as well.

PR #842, opened 2018-07-03
Save Window Position

OptimShi:
Saves window positions and restores them on login.

Note this requires a small SQL update in \Database\Updates\Shard\


OptimShi:
This resolves issue #691

Source/ACE.Server/Network/GameEvent/Events/GameEventPlayerDescription.cs
```diff
@@ -299,6 +299,9 @@ private void WriteEventBody()
             if (shortcuts.Count > 0)
                 optionFlags |= CharacterOptionDataFlag.Shortcut;
 
+            if (Session.Character.GameplayOptions.Length > 0)
```
kdmxp256: Null check needed on Session.Character.GameplayOptions.Length.  Other than this one item, PR works great.

OptimShi:
Added the null check so newbs shouldn't crash the server by having the audacity of logging in!

PR #850, opened 2018-07-06
Emote/Quests

dgatewood:
Quest Manager added to save/load to database.
Emotes filled out to do some basic quests.
There are some emotes that do need work, such as UpdateQuest, and all movement emotes.
Event  Manager was already in game, so I just wired the emotes that had already been filled for those.  Have not tested an event, but should work just like a quest but on a global scale.


gmriggs:
Test instructions from https://github.com/ACEmulator/ACE/pull/843:

Create a new 22754 (for the specialization dungeon)

keep clicking through the replies and watch for Jevak... when he asks for essence of jevak, give him @ci 777

Tested in-game, looks great from a functionality perspective!

For the requested code changes, I think it would be best if we merged this PR now, and I can help with refactoring the EmoteManager code in the future

PR #852, opened 2018-07-07
Image Export

OptimShi:
Added "image-export" console function to extract textures from client_portal.dat and client_highres.dat. Run the command with no parameters for syntax.


gmriggs:
Tested image-export command. Worked as expected

Very cool, this will definitely be useful!

PR #857, opened 2018-07-08
persistent tracking of item affectance

fartwhif:
item affectance is restored at Player.EnterWorld

### DIFF
--- a/Source/ACE.DatLoader/ACE.DatLoader.csproj
+++ b/Source/ACE.DatLoader/ACE.DatLoader.csproj
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="log4net" Version="2.0.8" />
+    <PackageReference Include="System.Drawing.Common" Version="4.5.0" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.5.0" />
   </ItemGroup>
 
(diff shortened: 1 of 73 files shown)


gmriggs:
Tested, functionality worked as expected

Before patching, I logged in with a character who had a magical item already equipped, and verified that the mana was not being depleted (pre-patch)

Still pre-patch, I dequipped and re-equipped the sword, and verified that the mana was indeed being depleted after re-equipping.

I also noticed that equipping magic items instantly used 1 mana upon wielding, not sure if that is intended / matches retail?

After patching, I logged in with the character while they had the magic sword already equipped, and the mana started ticking properly.

The spells for the wielded item also appeared in the player's enchantments properly

The 'IsAffecting' nomenclanture seems a bit non-standard to me (IsActivated?) but that is just a personal preference

Nice job with this patch, it really makes things feel more consistent!

fartwhif:
Thanks for the testing and critique!  If I remember correctly retail did have items burn 1 point of mana immediately upon equipping it.  I actually  went and changed ACE so match that pattern (mana stone PR).

fartwhif:
A bug, unless there's something wrong with my test conditions, crummy database or something?  One person already said that this bug was not there, but I tested to confirm it is still there for me.  Please help test.

1. /ci focusingstoneimbued
2. /ci 27329 (massive mana charge)
3. charge the focusing stone
4. equip the focusing stone
5. log out to the character selection screen
6. log back in with the same character
7. observe enchantments list, the spells from the focusing stone are still there
8. unequip the focusing stone
9. observe enchantments list, the spells from the focusing stone are still there even though the item that was casting them on the player is now unequipped.

Additional information: You can then re-equip the focusing stone and unequip it again to remove the focusing stone spells from the enchantments list.

PR #860, opened 2018-07-09
Fixing sliding along the edge of cliffs

gmriggs:
- Fixed a bug where StepUpHeight and StepDownHeight were reversed
- Fixed a conversion error between degrees and radians
- Fixed a bug with EdgeSlide/PrecipiceSlide where the player and creatures would get stuck at the edge of cliffs
- Updated CellArray iterators to check newly added cells


kdmxp256:
Should the first commit that looks like it comes from PR 857 be included in this one?

Issue #779, opened 2018-05-05
Splitting peas


Warloxx01:
### Bug Report or Feature Request (mark with an `x`)
```
- [x ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.

When splitting peas in the main pack no issue. when placing the pea say a (copper pea) in a secondary pack with the splitting tool and splitting the pea, you get the message that the pea was spit, and the content is placed in the main pack, but the pea is never deleted allowing you to split endlessly.

PR #867, opened 2018-07-10
Fix for using recipe items located in Packs

OptimShi:
This addresses Issue #779 

To test:

* /ci 8327 (Gold Pea)
* /ci 8283 (Splitting Tool)
* /ci 136 (Pack)
* Place these items in the Pack.
* Use the Splitting Tool on the Gold Pea in Pack.

Gold Pea should be removed and Gold Scarabs placed in main backpack.

### DIFF
--- a/Source/ACE.Server/Managers/RecipeManager.cs
+++ b/Source/ACE.Server/Managers/RecipeManager.cs
@@ -94,7 +94,8 @@ namespace ACE.Server.Managers
 
                     if (destroyTarget)
                     {
-                        if (target.OwnerId == player.Guid.Full)
+                        // TODO - Check if the target is wielded and handle appropriately
+                        if (target.OwnerId == player.Guid.Full  || player.GetInventoryItem(target.Guid) != null)
                         {
                             player.TryRemoveItemFromInventoryWithNetworking(target, (ushort)targetSuccess.DestroyAmount);
                         }
@@ -112,7 +113,8 @@ namespace ACE.Server.Managers
 
                     if (destroySource)
                     {
-                        if (source.OwnerId == player.Guid.Full)
+                        // TODO - Check if the source is wielded and handle appropriately
+                        if (source.OwnerId == player.Guid.Full || player.GetInventoryItem(target.Guid) != null)
                         {
                             player.TryRemoveItemFromInventoryWithNetworking(source, (ushort)sourceSuccess.DestroyAmount);
                         }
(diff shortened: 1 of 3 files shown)


Source/ACE.Server/Managers/RecipeManager.cs
```diff
@@ -112,7 +113,8 @@ public static void UseObjectOnTarget(Player player, WorldObject source, WorldObj
 
                     if (destroySource)
                     {
-                        if (source.OwnerId == player.Guid.Full)
+                        // TODO - Check if the source is wielded and handle appropriately
```
gmriggs: A new function was recently added for this: TryRemoveItemWithNetworking
It checks if the item is currently wielded first, and if so, tries to unwield it before calling TryRemoveFromInventoryWithNetworking

Warloxx01:
Works but doesn't work. Splits fine. but when i try to move the splits items ( in this case the Gold Scarabs) from the Main pack to the secondary pack/bag/pouch were i keep the comps it pulls the gold scarabs from the secondary pack to the main pack and stacks them there. then I can move the Stack( in this case the Gold Scarabs) to the secondary pack.

Source/ACE.Server/Managers/RecipeManager.cs
OptimShi: This function does not remove enchantments applied by the item. Is that intentional, or oversight? I'm going to guess oversight, since I see nothing in the functions that use it that would handle this, either.

gmriggs: Not removing enchantments - oversight, although this should be handled in the UnwieldItemWithNetworking function

Placing in backpack first - this is known, and was a limitation of the current TryRemoveFromInventoryWithNetworking system

OptimShi:
Test Recipe on Wielded Items
/ci 6127 (Stone Tool)
/ci 6128 (Peerless Shivering Atlan Staff)
Equip Peerless Shivering Atlan Staff
Use Stone Tool on Peerless Shivering Atlan Staff
(Note that enchantments applied by the staff will not be removed)

gmriggs:
Tested both scenarios in-game, looks good

gmriggs:
@Warloxx01 - thanks for the info!

When trying to merge stacks, this is a bug with the existing inventory system. @Mag-nus will be taking a look at these inventory bugs and fixing that system up very soon!

PR #868, opened 2018-07-12
EmoteManager refactoring

gmriggs:


gmriggs:
Fixed the double rotation bug:

Player rotates toward target, and target rotates toward player are 2 separate events

Give chain waits for just the player rotation, and receive chain waits for both

PR #885, opened 2018-07-22
More enums added from the client/ACLogView to ACE.Entity.Enum

Mag-nus:
Note that HookType exists in both:
ACE.Entity.Enum
ACE.Server.Physics.Hooks

The one in Physics should probably be renamed to PhysicsHookType so that the more global one in ACE.Entity.Enum can maintain consistant naming with the client.

### DIFF
--- a/Source/ACE.Entity/Enum/AttackType.cs
+++ b/Source/ACE.Entity/Enum/AttackType.cs
@@ -5,24 +5,24 @@ namespace ACE.Entity.Enum
     [Flags]
     public enum AttackType
     {
-        Undef_AttackType                = 0,
-        Punch_AttackType                = (1 << 0),
-        Thrust_AttackType               = (1 << 1),
-        Slash_AttackType                = (1 << 2),
-        Kick_AttackType                 = (1 << 3),
-        OffhandPunch_AttackType         = (1 << 4),
-        DoubleSlash_AttackType          = (1 << 5),
-        TripleSlash_AttackType          = (1 << 6),
-        DoubleThrust_AttackType         = (1 << 7),
-        TripleThrust_AttackType         = (1 << 8),
-        OffhandThrust_AttackType        = (1 << 9),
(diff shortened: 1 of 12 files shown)


Source/ACE.Entity/Enum/AttackType.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+
+namespace ACE.Entity.Enum
+{
+    [Flags]
+    public enum AttackType
+    {
+        Undef_AttackType                = 0,
```
LtRipley36706: can we strip _AttackType off?

Source/ACE.Entity/Enum/HookTypeEnum.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace ACE.Entity.Enum
+{
+    /// <summary>
+    /// The client calls this HookType, but, that name conflicts with HookType that resides in ACE.Server.Physics.Hooks.
+    /// The one in Physics should probably be renamed to PhysicsHookType
```
LtRipley36706: 💯 agree

Source/ACE.Entity/Enum/CombatBodyPart.cs
```diff
@@ -0,0 +1,36 @@
...
+        Tail            = 17,
+        Arm             = 18,
+        Leg             = 19,
+        Claw            = 20,
+        Wings           = 21,
+        Breath          = 22,
+        Tentacle        = 23,
+        UpperTentacle   = 24,
+        LowerTentacle   = 25,
+        Claok           = 26,
```
gmriggs: misspelling?

Source/ACE.Entity/Enum/HookTypeEnum.cs
Mag-nus: This should happen in a separate PR.

PR #889, opened 2018-07-22
This moves the .sql file generation code that Ripley has been using into ACE.Database

Mag-nus:
This will help us import data from any source into ACE EF objects, and then use these SQLWriters to create a unified output.


Mag-nus:
There are still 4 more writers to port over.

Before I complete them, this helps introduce what I'm working on into master.

PR #892, opened 2018-07-24
Three Emote Manager fixes/changes

kdmxp256:
-EmoteManager.Execute() moved to a new partial class of EmoteManager and Obsolete attribute and commenting added, as it may cause confusion as to which emote execute function to use
-Implement EmoteType.Act
-Change the target of the PlayScript in EmoteType.PhysScript from targetObject to the WorldObject that is using the emote


Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -904,10 +238,16 @@ public void ExecuteEmote(BiotaPropertiesEmote emote, BiotaPropertiesEmoteAction
...
-                //Console.WriteLine($"ExecuteEmote({emoteType})");
+            //Console.WriteLine($"ExecuteEmote({emoteType})");
+
+            var text = emoteAction.Message;
 
             switch ((EmoteType)emoteAction.Type)
             {
+                case EmoteType.Act:
+                    // Short for Acting...
+                    sourceObject.CurrentLandblock?.EnqueueBroadcast(sourceObject.Location, new GameMessageSystemChat(text.Replace("%n", sourceObject.Name), ChatMessageType.Broadcast));
```
gmriggs: Should be using common/existing Replace method already defined in EmoteManager

Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -42,672 +42,6 @@ public void Cancel()
         }
 
         // Thanks to the original pioneers in the emulator community for help decoding these structures!
-        public void Execute(Emote emote, WorldObject target)
```
gmriggs: This certainly makes EmoteManager a lot cleaner / easier to deal with, but it also greatly decreases the visibility of the example code.

Suggested to move this example code to a new file, EmoteManagerPrototype.cs perhaps?

PR #908, opened 2018-07-30
Auras, imp, and banes for @buff [player name] and @fellowbuff

fartwhif:
including auras and equipped item impenetrability and banes for sentinel commands /buff [player name] and /fellowbuff
fixing bug causing wielded item buffs not to work for shields in some situations
curating some elements of Ace.Server/Network/Enum/Spell.cs to follow uppercase/lowercase guidelines and self/other differentiation


Source/ACE.Server/Command/Handlers/SentinelCommands.cs
```diff
@@ -220,27 +221,60 @@ public static void BuffPlayers(IEnumerable<Player> players, Player caster, bool
...
+                    var items = (targetPlayer as Player).GetAllWieldedItems();
+                    var itembuffs = buffMessages.Where(k => k.Bane).ToList();
+                    foreach (var itemBuff in itembuffs)
+                    {
+                        foreach (var item in items)
+                        {
+                            if (item.WeenieType == WeenieType.Clothing || item.IsShield)
+                            {
+                                itemBuff.SetLandblockMessage(item.Guid);
+                                var enchantmentStatus = targetPlayer.ItemMagic(item, itemBuff.SpellBase, itemBuff.Spell, caster);
```
kdmxp256: I think it would be better to call the code, CreatePlayerSpell() that is already setup to handle the inner magic routines, rather than changing ItemMagic() from protected to public.  Mostly to make sure something isn't missed.  No need to reinvent the wheel.

Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -575,7 +575,7 @@ protected EnchantmentStatus CreatureMagic(WorldObject target, SpellBase spell, D
         /// <param name="spell"></param>
         /// <param name="spellStatMod"></param>
         /// <param name="itemCaster"></param>
-        protected EnchantmentStatus ItemMagic(WorldObject target, SpellBase spell, Database.Models.World.Spell spellStatMod, WorldObject itemCaster = null)
+        public EnchantmentStatus ItemMagic(WorldObject target, SpellBase spell, Database.Models.World.Spell spellStatMod, WorldObject itemCaster = null)
```
kdmxp256: See comment in SentinelCommands.cs

Source/ACE.Server/Command/Handlers/SentinelCommands.cs
fartwhif: Maybe.  Does it allow for the suppression of chat messages?  Can I add another parameter for such a feature?

Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -825,7 +825,12 @@ private EnchantmentStatus CreateEnchantment(WorldObject target, WorldObject cast
                 else
                 {
                     if (caster is Creature)
-                        message = $"You cast {spell.Name} on {targetName}{suffix}";
+                    {
+                        if (caster.Guid == Guid)
+                            message = $"You cast {spell.Name} on {targetName}{suffix}";
+                        else
+                            message = $"{caster.Name} casts {spell.Name} on {targetName}{suffix}"; // for the sentinel command `/buff [target player name]`
```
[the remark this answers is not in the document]
fartwhif: A command is invoking the target player's ItemMagic procedure, a different player in some cases.

Source/ACE.Server/Command/Handlers/SentinelCommands.cs
fartwhif: I tried this and it causes the target player(s) to go into combat mode and do a casting animation, and this is unwanted in this use case.  The intention for this command is for it to function as transparently as possible.  It will be doing up to 98 of these invocations per player, which is the intention.  If Player.CreatePlayerSpell() could have some options to disable some of the normal behavior it could be used for this.

kdmxp256: My vote would be to create a new public API call in Player_Magic to handle interfacing to the magic classes, rather than exposing them unnecessarily, somewhat similar to what I did for Item cast spells, with CastItemSpell().  This would also keep core magic functionality contained to a small footprint, rather than having it spread out into other locations.

PR #917, opened 2018-08-01
Add support for Volley spells

Slushnas:
This adds initial support for Volley type spells. I just put static creation offsets in for now until the correct formula is know for finding them.

Thanks to @gmriggs for some pointers on some of the physics related errors I encountered after implementation that needed to be fixed/tweaked.


gmriggs:
Code looks good, tested this functionality in-game:

Animation looks good, multi-spell projectile angles and speeds looks good, and I was able to cast the various volley spells and thousand fists to hit multiple targets simultaneously

Great PR, definitely looking forward to using these spells when battling many mobs!

PR #919, opened 2018-08-04
Resolve apparent logic error in learnspellsinbulk method which caused…

mcreedjr:
… certain spells to not be learned as expected


gmriggs:
Hey @mcreedjr,

Could you explain this patch a bit more? It looks like it's just skipping over the fist index in PlayerSpellID, which is SpellID=1 (StrengthOther1)?

mcreedjr:
My purported logic error had a logic error as pointed out by @gmriggs. Closing for now. Sorry for the SPAM.

PR #953, opened 2018-08-14
WIP - Add new spells and some refactoring

Slushnas:
This PR includes some refactoring that should make it cleaner/easier to add the missing spell types. It should also fix issues with projectiles not spawning in the correct cell (previously they were assumed to be in the same cell as the caster).

### DIFF
--- a/Source/ACE.Entity/Position.cs
+++ b/Source/ACE.Entity/Position.cs
@@ -36,8 +36,6 @@ namespace ACE.Entity
             return new Tuple<bool, bool>(blockUpdate, cellUpdate);
         }
 
-        public Vector3 GlobalPos { get => ToGlobal(); }
-
         public Quaternion Rotation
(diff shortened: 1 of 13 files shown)


Slushnas:
This adds support for the missing spell types: Volley, Blast, Ring and Wall. I have not added support for monsters to use the new untargeted types (ring and wall) yet.

Known issues:
1. When firing ring spells at close range the projectiles may not register a hit.
2. There appears to be some issues with ring spell projectiles not being cleaned up as fast as other spell types. If you fire a bunch of successive ring spells you may get stuck behind an invisible wall of projectiles that have not been updated/deleted yet.
3. In a quick wall test the first drudge hit registered properly and was killed but the second drudge hit by the right side of the wall did not. There may be some hit detection issues with these that require further testing/digging.

PR #959, opened 2018-08-18
Network fatal exception recovery

gmriggs:
This patch recovers from fatal network errors, such as 'FATAL: Network Socket has thrown: An existing connection was forcibly closed by the remote host' which seems to be thrown often from ThwargLauncher connections

The ConnectionListener.Listen() function starts receiving data from a socket, calling OnDataReceive. At the end of OnDataReceive, the Listen() function is called again to resume listening for the next request

If an exception is thrown in Listen(), the exception handler needs to call Listen() again to recover gracefully and to start listening for the next request.


Mag-nus:
Do you know the actual exception being thrown that we can recover from?

I would add multiple catches. Above the catch (Exception exception) I would ass the catch for the specific exception you were seeing that is recoverable.

I would then add a message indicating that the listener is being restarted.

PR #960, opened 2018-08-19
Shard Biota Save & Threading Model Rework

Mag-nus:
This is a major change on how Biotas are added/saved to shard.

The AddBiota function is no more. Now, to add or save a biota, the same function is used, SaveBiota.

SaveBiota will pull (in full) the existing biota (if it exists) from the database. It will do a full diff between the new biota to save and the existing biota. It will copy over any differences, add any new rows, and delete from the database any rows that have been removed.

In addition, a ReadWriterLockSlim is passed to the Save/Remove functions in the database. This is necessary to prevent collection modification from the server while Entity Framework is processing the object.

There are still some bugs I need to fix.
BiotaPropertiesEnchantmentRegistry needs to be accessed through this new model for thread safety.
Character also needs to be converted to this new model for thread safety
_Save/Remove InParallel functions need to have ReadWriterLockSlims passed for each biota._ -Completed

The new model offers thread safety, and is much more performant.
The existing model would submit a database request every time a new record was removed from any biota collection. This was required because our biota is detached from the context. On an active server, this would hammer the database.

The new model offers a true X minute save for a biota where only the changed properties are submitted to the database via an sql statement.

This could use playtesting.

### DIFF
--- a/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
+++ b/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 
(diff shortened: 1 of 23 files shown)


Mag-nus:
I'm seeing this exception, but I can't reliably duplicate it:
2018-08-19 12:07:53,129 [30] ERROR (ACE.Database.ShardDatabase) SaveBiota failed with exception: Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> MySql.Data.MySqlClient.MySqlException: Duplicate entry '1342177286-15' for key 'wcid_bool_type_uidx' ---> MySql.Data.MySqlClient.MySqlException: Duplicate entry '1342177286-15' for key 'wcid_bool_type_uidx'

PR #964, opened 2018-08-20
ShardDbContext caching. 1 for every Biota via a ConditionalWeakTable

Mag-nus:
I believe this to be the proper way to use EF with our application.
If we are able to improve db performance, it will be very marginal past this. This is near a best case scenario, if not already best case.

This utilizes Entity Frameworks tracking abilities for the life of the biota.

This does not suffer from context bloat because a single context only manages a single biota.

This does not suffer from data pollution because ACE does not support manipulating a biota via the database while it is currently used by the server.

When a biota is no longer used by ACE, the ConditionalWeakTable will release the reference to its cached DbContext. The garbage collector will dispose that for us.

If we want to explicitly dispose of the DbContext, we can use the FreeBiotaAndDisposeContext functions.

Comparing the current master method vs this method:
GetBiota and RemoveBiota are approximately the same performance.

Update via the current master:
1000 individual save   Duration:  36.3 s. Queue Wait Time:   0 ms. Average Execution Time:  36 ms. Success/Fail: 1000/0.
1000 bulk save         Duration:   5.6 s. Queue Wait Time:   0 ms. Average Execution Time:   6 ms. Success/Fail: 1/0.

Update via this PR:
1000 individual save   Duration:  10.0 s. Queue Wait Time:   0 ms. Average Execution Time:  10 ms. Success/Fail: 1000/0.
1000 bulk save         Duration:   2.7 s. Queue Wait Time:   0 ms. Average Execution Time:   3 ms. Success/Fail: 1/0.

The reason it is so much faster is because when the entity is tracked, only the changes need to be sent to the database. When the entity is disconnected (like in our current master), we must first get the entire biota from the database first to determine the changes.

There is an alternative to not caching the context. We could instead clone the biota and then diff the changes from the clone, however, there are some difficulties with this from an implementation pov, and it's much more prone to errors. In addition, it would not be as performant as this.

### DIFF
--- a/Source/ACE.Database/SerializedShardDatabase.cs
+++ b/Source/ACE.Database/SerializedShardDatabase.cs
@@ -260,6 +260,22 @@ namespace ACE.Database
             }));
         }
 
+        public void FreeBiotaAndDisposeContext(Biota biota)
+        {
+            _queue.Add(new Task(() =>
+            {
+                _wrappedDatabase.FreeBiotaAndDisposeContext(biota);
+            }));
+        }
+
+        public void FreeBiotaAndDisposeContexts(IEnumerable<Biota> biotas)
+        {
+            _queue.Add(new Task(() =>
+            {
+                _wrappedDatabase.FreeBiotaAndDisposeContexts(biotas);
+            }));
+        }
+
 
         public void RemoveEntity(object entity, Action<bool> callback)
         {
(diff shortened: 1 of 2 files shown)


Mag-nus:
There is a bug as a result of this PR that I am investigating:

Microsoft.EntityFrameworkCore.DbUpdateException

PR #967, opened 2018-08-23
Fixing some issues with stuck / falling monsters

gmriggs:


mcreedjr:
@gmriggs I believe I saw this same problem during play testing. I observed it primarily when surrounded by a mob. I observed monsters "jumping" while attacking. Does this patch address that behavior?

gmriggs:
Fixing a bunch of different things in the physics engine and with monster movement, it could possibly resolve this issue

PR #972, opened 2018-08-25
Threading improvement for WorldManager

Mag-nus:
Very simple functions should not be parallelized. It will actually be slower to do so.

In addition, it's unlikely there will be so many network sessions timed out in a single tick that would even require them to be parallelized.


Source/ACE.Server/Managers/WorldManager.cs
```diff
@@ -464,7 +464,10 @@ private static void UpdateWorld()
                 // Removes sessions in the NetworkTimeout state, incuding sessions that have reached a timeout limit.
                 var deadSessions = sessions.FindAll(s => s.State == Network.Enum.SessionState.NetworkTimeout);
                 if (deadSessions.Count > 0)
```
gmriggs: this if statement is redundant

PR #977, opened 2018-08-27
Ace admin plugin changes

acriaf:
Some more changes for the plug-in. Including a new class/file for plug-in commands.

### DIFF
--- a/Source/ACE.Database/WorldDatabase.cs
+++ b/Source/ACE.Database/WorldDatabase.cs
@@ -373,6 +373,23 @@ namespace ACE.Database
             return null;
         }
 
+        /// <summary>
+        /// Will return all Points of Interests from the DB
+        /// </summary>
+        public List<PointsOfInterest> GetAllPointsOfInterests()
+        {
+            using (var context = new WorldDbContext())
+            {
+                var results = context.PointsOfInterest
+                    .AsNoTracking();
+
+                if (results != null) return results.ToList();
+
+            }
+
+            return null;
+        }
+
         private readonly Dictionary<uint, Dictionary<uint, CookBook>> cookbookCache = new Dictionary<uint, Dictionary<uint, CookBook>>();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Source/ACE.Database/WorldDatabase.cs
```diff
@@ -373,6 +373,23 @@ public PointsOfInterest GetCachedPointOfInterest(string name)
             return null;
         }
 
+        /// <summary>
+        /// Will return all Points of Interests from the DB
+        /// </summary>
+        public List<PointsOfInterest> GetAllPointsOfInterests()
```
Mag-nus: GetAllPointsOfInterests should probably add to the POI Cache, since it's pulling from the database anyway.

Source/ACE.Server/Command/Handlers/PlugInCommands.cs
```diff
@@ -0,0 +1,155 @@
...
+                    AsNoTracking();
+
+                if (results != null)
+                    return results.ToList();
+
+            }
+
+            return null;
+        }
+
```
Mag-nus: Whats up with a lot of these weird empty lines?

Mag-nus:
In regards to PlugInCommands.cs,

I can't help but wonder if this info is better exposed via the debugging API that's being worked on.

At some point, most of the server information will be available via an API that uses a web front end. I think that would be a bit easier to use than a plugin for active server information and diagnostics.

Source/ACE.Database/WorldDatabase.cs
acriaf: The Plug-in does Eat the message so it will not go to the Chat Windows. Currently that line is commented out so I can see what is being sent.
POI Cache, I did have that, I Will add it back.

Source/ACE.Server/Command/Handlers/PlugInCommands.cs
acriaf: I will fix the empty lines.

acriaf:
I agree that the active server information and diagnostics will be easier and better suited for the API. 
The Plug-in is being written to assist with in-game specific functions with a few exceptions (e.g. creating a new account).

PR #987, opened 2018-09-01
Character gameplay_Options fix

Mag-nus:
This has two fixes.

1) ACE logoff code runs before the final GameActionSetCharacterOption event from the client. Because of this order of operations, any changed options weren't being saved to the character on logoff.

2) Character was being modified while it was being saved by EF. This caused some change tracking mismatches.

I mentioned before, but there are still a few lingering places that need this type of locking. It's another thing on my todo list.


gmriggs:
Tested, this appears to fix the window position saving issue

gmriggs:
Re-tested, confirmed window positions saved

PR #996, opened 2018-09-03
Implement secure trade

mcreedjr:
Some of this may be less than elegant, but it appears to work just fine!

### DIFF
--- a/Source/ACE.Entity/Enum/EndTradeReason.cs
+++ b/Source/ACE.Entity/Enum/EndTradeReason.cs
@@ -1,0 +1,9 @@
+namespace ACE.Entity.Enum
+{
+    public enum EndTradeReason
+    {
+        Normal = 0x1,
+        EnteredCombat = 0x2,
+        Canceled = 0x51
+    };
+}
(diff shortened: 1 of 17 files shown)


mcreedjr:
Thanks for the review. I can't remember the functionality in retail, either. It's easy enough to remove that forced trade closure upon a successful trade. We'll see if anyone else can chime in. Regardless, I now think it makes more sense to not do that anyways. I'll push a new commit later today.

mcreedjr:
Okay. I tightened up the code that runs when a trade is completed. It looks like I still have some more to do to trap for players already in session, moving away, entering combat mode, etc, but I'll circle back to those shortly. I think this is complete enough to merge.

gmriggs:
Tested, looks great!

gmriggs:
A few questions, just some minor stuff:

- did retail have any messages like 'You give <item> to player', 'Player gives you <item>' once the transaction goes through?

- if 1 player closes the trade window, the window seems to remain open for the other player, although the name of the other player disappears from the window, sort of indicating that they backed out of the trade. Is this the same as retail, or should the window automatically close?

- 'Ignore all trade requests' character option doesn't work, but maybe outside of the scope for current PR, just to get the functionality going

Still looks good for merging imo

PR #998, opened 2018-09-03
ActionQueue/Tick updates

Mag-nus:
The global LandblockActionQueue has been removed from WorldManager. This ran almost every action queue for all World Objects.

Landblocks are no longer IActors

Now, Landblocks Tick, and in turn, Tick the World Objects that they own.

WorldObjects contain their own ActionQueues. Previously, their action queue was a nested action queue. This just removes one of the layers.

The WorldObject.Tick will run it's own actions, and also check and run it's heartbeat.

Containers will propegate Ticks to it's contained items.

Creatures will propegate Ticks to it's equipped items.

Tick and Heartbeat are virtual. A player can add items to it's Heartbeat, then call it's base (creature), which calls its base (container) which calls it's base (world object).

** THIS IS A MAJOR CHANGE AND REQUIRES PLAY TESTING **

TODO in a following PR: Remove DelayManager from WorldManager


gmriggs:
lots of error messages:

2018-09-03 23:27:13,552 WARN : requested object description for an item I do not know about - 8000C020
i get that message almost immediately when loading in now
never saw that message once before

also a ton of these error messages:

System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.Dictionary`2.ValueCollection.Enumerator.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at ACE.Server.Managers.WorldManager.HandlePhysicsLandblock(Landblock landblock, Double timeTick, ConcurrentQueue`1 movedObjects) in C:\dev\gmriggs\ACE\Source\ACE.Server\Managers\WorldManager.cs:line 600
   at ACE.Server.Managers.WorldManager.HandlePhysics(Double timeTick) in C:\dev\gmriggs\ACE\Source\ACE.Server\Managers\WorldManager.cs:line 585

this was in the first few mins of testing

definitely a lot of bugs that need ironed out with this PR

Mag-nus:
You're seeing those WARN messages because I changed that logging line from DEBUG to WARN. I felt that DEBUG was not the appropriate level for such a message.

I believe these messages were being raised prior to this refactor, but not being noticed to to them being DEBUG.

Mag-nus:
Are you able to duplicate the collection modified exception?

gmriggs:
I have confirmed the above errors are now fixed

Playtested, everything else looks good

PR #1008, opened 2018-09-10
Added random Scroll functionality

dgatewood:
This is not all the scrolls that can be found from loot, but it is roughly half of them. I will add the rest very soon, or they could be added from a new contributor looking to get their feet wet.


Source/ACE.Server/Factories/LootHelper.cs
```diff
@@ -54,6 +54,260 @@ public static class LootHelper
                                                                 {50, "Zircon"}
                                                             };
 
+        public static int[][] ScrollSpells =
```
Mag-nus: In the future, maybe we should look into building these dynamically at startup from the .dat. I think that would help allow people to modify their .dat's to update the server. By doing static lists, I think that we may end up with dat mismatches if that was the case.

Also, keep in mind that there are many spells in the dat that are duplicates. They use different id's for the quest versions vs the loot generated versions.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -250,6 +250,18 @@ public void CacheAllWeenies()
             }
         }
 
+        public String GetScrollBySpellID(uint spellID)
```
Mag-nus: We're going to need to add some sort of caching mechanism here.

The way you have it written, it appears that every request will hit the database.

Issue #866, opened 2018-07-10
Picking Up Open Container


OptimShi:
### Bug Report or Feature Request (mark with an `x`)
```
- [ X ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.

* /ci 136 (Pack)
* Drop pack into world.
* Open Pack
* Pick up pack without closing.

Packed will be placed into inventory, but is locked. If placed back on ground, user will be unable to open it. Not tested, but I assume you can also pick up a pack while another player is using it. YOINK!

### Desired functionality.

* If player picking up the pack is the one with it open, the pack will remove/set to false the IsOpen flag when picking it up.
* If another player has the pack open, generate error (Perhaps WeenieError.Frozen - "The item is under someone else's control!")

mcreedjr:
These issues have been resolved with the referenced PR.

PR #1011, opened 2018-09-10
Fix the picking up of open and in use packs

mcreedjr:
This PR should address both concerns raised by @OptimShi as noted in issue #866.

The messages returned when a user attempts to place pack being viewed by another user were what I remembered of retail. I couldn't find anything in PCAPs to support what I remembered. I'm open to suggestions if those messages should be something else. I'd be happy to update with the consensus.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -492,7 +492,7 @@ namespace ACE.Server.WorldObjects
             bool containerOwnedByPlayer = true;
 
             Container container;
-
+            
             if (containerGuid == Guid)
                 container = this; // Destination is main pack
             else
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -544,6 +544,34 @@ public void HandleActionPutItemInContainer(ObjectGuid itemGuid, ObjectGuid conta
...
+
+                            if (containerToPickup.Viewer == Session.Player.Guid.Full)
+                            {
+                                //We're the one that has it open. Close it before picking it up
+                                containerToPickup.Close(Session.Player);
+                            }
+                            else
+                            {
+                                //We're not who has it open. Can't pick up something someone else is viewing!
+                                Session.Network.EnqueueSend(new GameEventWeenieErrorWithString(Session, WeenieErrorWithString.The_IsCurrentlyInUse, itemToPickup.Name));
```
Mag-nus: Can you add a note here that notes something along the lines of what you wrote in the initial PR: "The messages returned when a user attempts to place pack being viewed by another user were what I remembered of retail. I couldn't find anything in PCAPs to support what I remembered. "

mcreedjr: Thanks for the review @Mag-nus. My latest commit(s) address your suggestion. This should be good to merge now.

PR #1012, opened 2018-09-10
Player housing - adding the ability to place and save items on floor/wall/ceiling hooks

gmriggs:


Source/ACE.Database/SerializedShardDatabase.cs
```diff
@@ -237,6 +237,21 @@ public void GetWieldedItems(uint parentId, Action<List<Biota>> callback)
 
         }
 
+        public Biota GetObjectByGuid(uint guid)
```
Mag-nus: Why the addition of GetObjectByGuid. This seems to do the same thing as GetBiota, except it returns a biota that doesn't use our context caching system.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -387,6 +387,15 @@ public List<Biota> GetWieldedItemsInParallel(uint parentId)
             return wieldedItems.ToList();
         }
 
+        public Biota GetObjectByGuid(uint guid)
```
Mag-nus: Why the addition of GetObjectByGuid. This seems to do the same thing as GetBiota, except it returns a biota that doesn't use our context caching system.

PR #1014, opened 2018-09-12
Adding proficiency points, and skills increasing through usage

gmriggs:


Source/ACE.Server/WorldObjects/Creature_Magic.cs
```diff
@@ -98,42 +98,33 @@ public void CreateCreatureSpell(uint spellId)
 
         public uint CalculateManaUsage(Creature caster, SpellBase spell, WorldObject target = null)
         {
-            var items = new List<WorldObject>();
+            uint mana_conversion_skill = caster.GetCreatureSkill(Skill.ManaConversion).Current;
+            uint difficulty = Math.Max(25, spell.Power);     // for mana conversion only?
```
kdmxp256: Spells are not clamped to a base difficulty of 25; see http://asheron.wikia.com/wiki/Creature_Enchantment_Mastery_(Spell)

Source/ACE.Server/WorldObjects/Player_Magic.cs
```diff
@@ -506,9 +506,18 @@ public void CreatePlayerSpell(WorldObject target, TargetCategory targetCategory,
             float scale = SpellAttributes(player.Session.Account, spellId, out float castingDelay, out MotionCommand windUpMotion, out MotionCommand spellGesture);
             var formula = SpellTable.GetSpellFormula(spellTable, spellId, player.Session.Account);
 
-            if ((Physics.Common.Random.RollDice(0.0f, 1.0f) > (1.0f - SkillCheck.GetMagicSkillChance((int)magicSkill, (int)spell.Power)))
-                            && (magicSkill >= (int)spell.Power - 50) && (magicSkill > 0))
-                castingPreCheckStatus = CastingPreCheckStatus.Success;
+            var difficulty = spell.Power;
+            var difficultyMod = Math.Max(difficulty, 25);   // fix difficulty for level 1 spells?
```
kdmxp256: Spells are not clamped to a base difficulty of 25; see http://asheron.wikia.com/wiki/Creature_Enchantment_Mastery_(Spell).  The player should be able to bootstrap most magic skills into a usable state, but not subject to skill updates, to be able to perform rudimentary magic.

Source/ACE.Server/WorldObjects/Creature_Magic.cs
gmriggs: see discord conversation

Source/ACE.Server/WorldObjects/Player_Magic.cs
gmriggs: see discord conversation

kdmxp256: An exception needs to be made for when the magic skill are "unusable", then, as those skills are only technically unusable when they are in the zero state.  Most other unusable skills were still not useful even after buffing them above zero, as even the most basic tasks for them were still too difficult. However, buffing "unusable" magic skills allowed at least level one spells to be cast.  When these magic skills are unusable, proficiency is not a factor, so clamping the difficulty to 25 minimum should not be applied then.

gmriggs: see discord conversation

PR #1023, opened 2018-09-16
Adding Sneak Attack skill for player combat

gmriggs:


Source/ACE.Server/WorldObjects/SpellProjectile.cs
```diff
@@ -251,6 +251,9 @@ public override void OnCollideObject(WorldObject _target)
 
             var player = ProjectileSource as Player;
 
+            if (player != null)
+                player.LastHitSpellProjectile = spellStatMod;
```
kdmxp256: Might want to use SpellBase, instead of StatMod.  School originates in the SpellBase, which is pulled from the DAT.  This field is an extra in the StatMod from the Database and so could get removed at any time.

PR #1026, opened 2018-09-23
Update SQL Scripts

LtRipley36706:
Rebased and updated for MySQL 8.0

Should be backwards compatible as well.


Mag-nus:
This looks fine.

My only question is if the DirtyFighting.sql changes will be added to the ACE-World-16PY-Patches repo. They should be at the same time this PR is merged.

PR #1031, opened 2018-09-29
PlayerFactory

Mag-nus:
Player/Character creation has been pulled out of CharacterHandler and added to a new PlayerFactory.

PlayerFactory includes a Create275HeavyWeapons function to help with load testing.

SpendAllXp has been added to Player_Xp.cs. A command handler can be added to allow use of this feature from in-game.

SpendAttributeXp, SpendVitalXp and SpendSkillXp no longer perform the work unless SpendXp was successful.

Player ctor from weenie no longer takes a Session object. Instead, only the AccountId is passed. This allows player creation without needing a session.


Mag-nus:
spendallxp developer command added

fixed SpendSkillXp so that the rankup is only applied if the SpendXp was successful

PR #1036, opened 2018-10-01
Fix DamageType for Boost spells

kdmxp256:
Change DamageType to DamageType2 for LifeMagic Boost spells


Mag-nus:
Tested revitalize and it works.

PR #1038, opened 2018-10-01
Landblock Preloading PoC

kdmxp256:
-Add functionality for preloading landblocks and making them perma-loaded

### DIFF
--- a/Source/ACE.Server/Managers/LandblockManager.cs
+++ b/Source/ACE.Server/Managers/LandblockManager.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Diagnostics;
 
 using log4net;
 
@@ -59,23 +60,27 @@ namespace ACE.Server.Managers
         }
 
         /// <summary>
-        /// This should only be used for debugging/development purposes.
+        /// Loads the specified list of landblocks and optionally their adjacents.
         /// </summary>
-        public static void ForceLoadLandBlock(LandblockId blockid)
+        public static void ForceLoadLandBlock(LandblockId blockid, bool propagate = false)
         {
-            GetLandblock(blockid, false);
+            GetLandblock(blockid, propagate, true);
         }
 
         /// <summary>
         /// gets the landblock specified, creating it if it is not already loaded.  will create all
         /// adjacent landblocks if propagate is true (outdoor world roaming).
         /// </summary>
-        private static Landblock GetLandblock(LandblockId landblockId, bool propagate)
+        private static Landblock GetLandblock(LandblockId landblockId, bool propagate, bool permaload = false)
         {
             lock (landblockMutex)
             {
                 var landblock = landblocks[landblockId.LandblockX, landblockId.LandblockY];
                 var autoLoad = propagate && landblockId.MapScope == MapScope.Outdoors;
+
+                // Set Permaload flag, as required, for an already loaded landblock
+                if (landblock != null)
+                    landblock.Permaload = permaload;
 
                 // standard check/lock/recheck pattern
                 if (landblock == null || autoLoad && !landblock.AdjacenciesLoaded)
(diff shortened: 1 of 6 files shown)


Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -212,5 +220,28 @@ public static void UnloadLandblocks()
...
+            {
+                Stopwatch sw = Stopwatch.StartNew();
+                ForceLoadLandBlock(landBlockId, true);
+                sw.Stop();
+                log.DebugFormat("Landblock {0:X4} preloaded in {1} milliseconds", landBlockId.Landblock, sw.ElapsedMilliseconds);
+            }
+        }
+
+        // TODO: Change the RawLandblockId list used for preloading defined landblocks to some other, more easily-modified format, rather than a compiled uint array
+        private static readonly uint[] RawLandblockId = { 0xa9b4ffff, 0xabb2ffff, 0xaab3ffff, 0x7d64ffff, 0x7e64ffff, 0xe64effff, 0xe74effff, 0xda55ffff, 0xdb54ffff,
```
Mag-nus: Can this be changed so that each id is on it's own line followed by a comment describing the landblock by name?

Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -212,5 +220,28 @@ public static void UnloadLandblocks()
...
+
+            for (uint x = 0; x < RawLandblockId.Length; x++)
+            {
+                var landBlockId = new LandblockId(RawLandblockId[x]);
+                landBlockIdList.Add(landBlockId);
+            }
+
+            foreach (var landBlockId in landBlockIdList)
+            {
+                Stopwatch sw = Stopwatch.StartNew();
```
Mag-nus: You don't really need the stopwatch stuff here unless you're really interested in logging the time.

Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -59,24 +60,28 @@ public static List<Landblock> GetActiveLandblocks()
         }
 
         /// <summary>
-        /// This should only be used for debugging/development purposes.
+        /// Loads the specified list of landblocks and optionally their adjacents.
         /// </summary>
-        public static void ForceLoadLandBlock(LandblockId blockid)
+        public static void ForceLoadLandBlock(LandblockId blockid, bool propagate = false)
```
Mag-nus: The arguments for this function should be the same as GetLandblock:
LandblockId landblockId, bool propagate, bool permaload = false

Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -59,24 +60,28 @@ public static List<Landblock> GetActiveLandblocks()
...
-        private static Landblock GetLandblock(LandblockId landblockId, bool propagate)
+        private static Landblock GetLandblock(LandblockId landblockId, bool propagate, bool permaload = false)
         {
             lock (landblockMutex)
             {
                 var landblock = landblocks[landblockId.LandblockX, landblockId.LandblockY];
                 var autoLoad = propagate && landblockId.MapScope == MapScope.Outdoors;
 
+                // Set Permaload flag, as required, for an already loaded landblock
+                if (landblock != null)
```
Mag-nus: This check shouldn't be here. The one you have right after the new Landblock(landblockId) is the correct location.

Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -212,5 +220,47 @@ public static void UnloadLandblocks()
                 }
             }
         }
+
+        public static void Initialize()
```
Mag-nus: personal preference but I like to see Initialize functions near the top due to their significance. I would probably make it the first function in the class.

Source/ACE.Server/Managers/LandblockManager.cs
kdmxp256: I'm not sure that I understand why the arguments need to be the same.  The function originally statically defined bool propagate as false.  For landblock preloading, it makes sense to have it be true, hence the option argument for propagate with a default of false so as to not break anything already using ForceLoadLandBlock().  When forcing a landblock to load, it doesn't really make sense to allow the same landblock to then unload itself five minutes later, which is why I didn't see any issue with making that a static assignment for the internal function call.

Mag-nus: They should be the same to give the caller the power to chose.

In the current implementation, every landblock loaded via Force is set to permaload. That's not an assumption we can make every time.

I added ForceLoadLandblock, and it's only used in one place, by a developer load test command.

It doesn't hurt to have the signatures the same and let the caller decide if the landblock should propegate and permaload.

PR #1039, opened 2018-10-01
Alleg Fixes plus fix for No Default SQL

acriaf:
Fixes an issue where a new characters will not show as joining an allegiance until a server restart. If the new character is the patron the server will crash, also fixed. During testing of the latest server code found that character creation was broken. Added that fix here by setting up Character.IsPlussed to 'true' in ..ACE\Source\ACE.Server\WorldObjects\Player.cs. I am not sure if it should be 'true' or 'false'.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -49,7 +49,6 @@ namespace ACE.Server.WorldObjects
             Character.Id = guid.Full;
             Character.AccountId = accountId;
             Character.Name = GetProperty(PropertyString.Name);
-            Character.IsPlussed = true;
             CharacterChangesDetected = true;
 
             // Make sure properties this WorldObject requires are not null.
@@ -59,6 +58,8 @@ namespace ACE.Server.WorldObjects
             Attackable = true;
 
             SetProperty(PropertyString.DateOfBirth, $"{DateTime.UtcNow:dd MMMM yyyy}");
+
+            SetEphemeralValues();
         }
 
         /// <summary>
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -49,7 +49,7 @@ namespace ACE.Server.WorldObjects
             Character.Id = guid.Full;
             Character.AccountId = accountId;
             Character.Name = GetProperty(PropertyString.Name);
-            Character.IsPlussed = true;
+            Character.IsPlussed = false;
             CharacterChangesDetected = true;
 
             // Make sure properties this WorldObject requires are not null.
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -49,7 +49,7 @@ namespace ACE.Server.WorldObjects
             Character.Id = guid.Full;
             Character.AccountId = accountId;
             Character.Name = GetProperty(PropertyString.Name);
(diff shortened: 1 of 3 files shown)


gmriggs:
Hey Riaf,

Thanks for this patch! This has been a known issue for quite awhile

The root cause is indeed the AllPlayers list not being maintained properly. This AllPlayers structure was admittedly quite a large hack, and there have been discussions about refactoring this into a better architecture since it was created.

This will fix a lot of issues with AllPlayers in the meantime, great patch

Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -49,6 +49,7 @@ public Player(Weenie weenie, ObjectGuid guid, uint accountId) : base(weenie, gui
             Character.Id = guid.Full;
             Character.AccountId = accountId;
             Character.Name = GetProperty(PropertyString.Name);
+            Character.IsPlussed = true;
```
Mag-nus: I don't believe setting IsPlussed = true here to be correct.

Can you describe steps to reproduce the + issue that you're seeing?

Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -58,8 +59,6 @@ public Player(Weenie weenie, ObjectGuid guid, uint accountId) : base(weenie, gui
             Attackable = true;
 
             SetProperty(PropertyString.DateOfBirth, $"{DateTime.UtcNow:dd MMMM yyyy}");
-
-            SetEphemeralValues();
```
Mag-nus: Why the removal of this call? SetEphemeralValues is a pretty critical call to the setting up of the object.

Source/ACE.Server/WorldObjects/Player.cs
acriaf: I just did some more testing and IsPlussed should be set to 'false' at this point. 

The issue I saw was with the latest server code, creating a new character, On login the Sever would throw an error. 
"SaveCharacter failed with exception: Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details. ---> MySql.Data.MySqlClient.MySqlException: Field 'is_Plussed' doesn't have a default value ---> MySql.Data.MySqlClient.MySqlException: Field 'is_Plussed' doesn't have a default value"

Source/ACE.Server/WorldObjects/Player.cs
acriaf: I did some further checking and this is what I found.
Existing characters and Character creation both call...
ACE.Server\Managers\WorldManager.cs -> DoPlayerEnterWorld
that calls ACE.Server\WorldObjects\Player.cs -> Player which calls SetEphemeralValues

On new characters with the code in this pull request. 
ACE.Server\Network\Handlers\CharacterHandler.cs
Now calls. ACE.Server\Managers\WorldManager.cs -> AddPlayer
that calls ACE.Server\WorldObjects\Player.cs -> Player which calls SetEphemeralValues

So in the end we are still calling it twice on character creation even with the first call removed.

Mag-nus: SetEphemeralValues needs to be called from every ctor. If you remove it from Player, it should be removed from every child/parent from weenie ctor, which isn't right.

If there's an issue you're trying to solve by removing the call, then the problem is likely that there is code in SetEphemeralValues that shouldn't be in there.

Also, Character.IsPlussed = false; shouldn't be there as well. No need to set it to the default value.

Feel free to reach out to me in discord.

PR #1041, opened 2018-10-02
Refactored generators, improved RNG selection formula

gmriggs:
- This fixes various bugs with generators, including:
  - Generators spawning too many objects
  - Generators RNG selecting the wrong objects
  - Generators missing various RNG formulas

===

- InitCreate and MaxCreate should be ints instead of uints (still needs fixed for SQL output)
- Generators now use RegenInterval heartbeats

===

- Improved MotionRange broadcast performance slightly w/ distance squared
- Improved performance for landblock objects by staggering the heartbeat offsets
- Added default values for Generator properties
- Moved HandlePhysics and ActiveLandblocks to UpdateGameWorld(), targeted to run @ 60hz
- Updated SessionHandler for improved performance in common low usage scenarios. This needs to be re-evalulated to find the threshold where Parallel.ForEach improves performance
- Removed calls to legacy Landblock.UpdateStatus() methods
- Fixed a bug where resisted spells weren't alerting monsters
- Fixed a crash for landblocks with no Info record


Source/ACE.Database/Models/World/WeenieExtensions.cs
```diff
@@ -363,8 +362,8 @@ public static Biota CreateCopyAsBiota(this Weenie weenie, uint id)
                     Probability = value.Probability,
                     WeenieClassId = value.WeenieClassId,
                     Delay = value.Delay,
-                    InitCreate = value.InitCreate,
-                    MaxCreate = value.MaxCreate,
+                    InitCreate = (int)value.InitCreate,     // FIXME in SQL generator: should be int?
+                    MaxCreate = (int)value.MaxCreate,
```
LtRipley36706: Fixed these in the database, cast can be removed now

PR #1046, opened 2018-10-04
Timing

Mag-nus:
gmriggs, if you could please use your timing utils to verify that UpdateGameWorld is averaging 60fps.

Also, I think checking if the session count >= 5 just to save 1ms for this line of code is wasteful: Parallel.ForEach(sessions, s => s.TickInParallel());

On my system, the average overhead for the Parallel.ForEach call was 1ms to 2ms.

TickInParallel takes on average, .1ms to .3ms but can spike to 5ms to 15ms.

I think adding extra optimizations to ACE for player ranges of < 10 are a bit unnecessary. The server will already be extremely fast in such a scenario.


gmriggs:
Looks good, locked to 60 fps

PR #1048, opened 2018-10-06
Adding support for level 8 spells

gmriggs:
- Updated Spell enum to have consistent casing between spell levels
- Fixed a bug in the spell DAT reader for component loss (uint -> float)
- Fixed a 'collection was modified' error during player death while removing enchantments
- Added /dispel developer command
- Added /showstats debug command to show all player attribute and skill levels from server
- Improved /buff command with an optional spell level # parameter
= Fixed a bug where /buff command was applying multiple versions of the same spells
- Fixed a bug where level 8 spells had their minimum power set to 350 instead of 400
- Fixed a bug where level 8 spell effects were too small
- Added TransferFlags enum for life magic transfer spells
- Fixed a bug where no death message was displayed for death blows from DamageType.Health (harm, drain, life projectiles)
- Fixed a bug where item enchantments were having their ticks applied twice
- Fixed a bug where mages would cast spells before rotating completely to target
- Added level 8 item auras to EnchantmentManager.GetModifier()
- Player magic refactoring
	- Fixed a bug where players would windup / fizzle if they didn't have enough mana
	- Updated text messages for life magic transfer and boost spells to match retail more closely
	- Fixed a bug where life projectiles were getting the target stats instead of the source stats
	- Updated life magic transfer spells to use TransferFlags


Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1181,14 +1181,14 @@ public static void Handledirect_emote_select(Session session, params string[] pa
         }
 
         // dispel
-        [CommandHandler("dispel", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 0)]
+        [CommandHandler("dispel", AccessLevel.Developer, CommandHandlerFlag.None, 0, "Removes all enchantments from the player", "dispel")]
```
Mag-nus: I think this should still be RequiresWorld.

You wouldn't be able to do it while in portal space, but, None would allow this from the console I believe. Have to confirm this.

PR #1052, opened 2018-10-08
Adding methods to QuestManager

gmriggs:
- Added quest solve timers to item pickups
- More methods added to QuestManager
- Fixed a bug where recipe items could be combined with themselves
- Added ConvertToMoASkill() to recipe skill checks
- Added Item Tinkering to skill updates


Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -622,6 +604,51 @@ protected EnchantmentStatus ItemMagic(WorldObject target, Spell spell, WorldObje
             return enchantmentStatus;
         }
 
+        /// <summary>
+        /// Spawns a player-summoned portal from item magic or gems
+        /// </summary>
+        protected void SummonPortal(uint portalId, Position destination, double portalLifetime)
```
LtRipley36706: this code is kind of strange to me.. I'm confident it works similarly enough to what retail did but, it goes about it differently than I think it maybe should.. 

summoned portals, be they from players, npc, or items, are copies of `portalgateway` or wcid 1955.. As such I think we should clone/copy 1955, alter/apply its destination from whatever portal destination or id we're using as the link and then put a TimeToRot value so that portals Rot using that heartbeat rot system similar to other item decay.

https://github.com/ACEmulator/ACE-World-16PY/blob/master/Database/3-Core/9%20WeenieDefaults/SQL/Portal/PORTAL/01955%20Gateway.sql

LtRipley36706: there are likely other properties we should copy as well from the portal that the gateway portal is cloning destination from... things like min/max level, restrictions..

gmriggs: I did not write this code... this was existing code pulled out of the item magic block

gmriggs:
- Fixed a bug where /smite all would kill nearby NPCs
- Fixed a bug where teleporting / dying in hotspots would not trigger hotspot.OnCollideEnd
- Fixed a bug where ethereal objects could prevent creatures from spawning
- Added portal summons to monster item magic / emotes

PR #1060, opened 2018-10-12
LaunchMissile check for null (fix exception)

Mag-nus:
Should this even be wrapped in an actionchain?

We shouldn't wrap things in action chains now unless we really need to.


Mag-nus:
Looks like the action chain is required if we have more ammo, but can probably be refactored out for ammo.StackSize == 1.

PR #1065, opened 2018-10-14
Adding more quest emotes, refactoring EmoteManager

gmriggs:
- Started to add @qst command to match admin client.
- Added Home position for all Creatures, and support for the MoveHome emote
- Removed the redundant delay adding code from the individual emotes.
- Removed most of the ActionChain code from the inner ExecuteEmote function.
  These are both now handled consistently by the calling function, ExecuteEmoteSet.
- The EmoteManager refactoring is geared towards supporting an important feature, the concept of queueing and blocking emote sets.
- Added support for NPCs in the busy state while executing an emote chain
- Added support for monster AiOptions. Currently the only known option is the ability to open doors. Monsters such as Olthoi will now open the squishy doors when they touch them
- Fixed a bug where instant healthbar updates for the player's attack target were not correctly reporting the death blow
- Fixed a bug where some ethereal objects such as portals were blocking the movement of some creatures
- Fixed a possible null pointer exception with ActionChain.AddChain() being passed an empty chain
- Fixed a bug where TurnToHeading in the movement packet was not sending the DesiredHeading


Source/ACE.Server/Entity/Generator.cs
```diff
@@ -88,6 +93,16 @@ public DateTime GetSpawnTime()
                 if (_generator.GeneratorInitialDelay == 6000)   // spawn repair golem immediately?
```
LtRipley36706: leaving this in or want to remove?

PR #1066, opened 2018-10-14
Command verb "list" for developer command /telepoi 

fartwhif:
(re)caches and lists all POIs

### DIFF
--- a/Source/ACE.Database/WorldDatabase.cs
+++ b/Source/ACE.Database/WorldDatabase.cs
@@ -13,6 +13,7 @@ using log4net;
 
 using ACE.Database.Models.World;
 using ACE.Entity.Enum;
+using System.Diagnostics;
 
 namespace ACE.Database
 {
@@ -376,6 +377,14 @@ namespace ACE.Database
             return cachedPointsOfInterest.Count(r => r.Value != null);
         }
 
+        /// <summary>
+        /// Returns the PointsOfInterest cache.
+        /// </summary>
+        public ConcurrentDictionary<string, PointsOfInterest> GetPointsOfInterestCache()
+        {
+            return cachedPointsOfInterest;
+        }
+
         public PointsOfInterest GetCachedPointOfInterest(string name)
         {
             var nameToLower = name.ToLower();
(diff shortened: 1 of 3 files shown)


Source/ACE.Database/WorldDatabase.cs
```diff
@@ -395,10 +404,12 @@ public PointsOfInterest GetCachedPointOfInterest(string name)
         }
 
         /// <summary>
-        /// This takes under 1 second to complete.
+        /// Retrieves all points of interest from the database and adds/updates the points of interest cache entries with every point of interest retrieved.
         /// </summary>
-        public void CacheAllPointsOfInterest()
+        /// <returns>time used</returns>
+        public TimeSpan CacheAllPointsOfInterest()
```
Mag-nus: I'd prefer this function not return the duration it took to do the work.

If a consumer of the function is interested in that information, it should calculate itself.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -376,6 +377,14 @@ public int GetPointsOfInterestCacheCount()
             return cachedPointsOfInterest.Count(r => r.Value != null);
         }
 
+        /// <summary>
+        /// Returns the PointsOfInterest cache.
+        /// </summary>
+        public ConcurrentDictionary<string, PointsOfInterest> GetPointsOfInterestCache()
```
Mag-nus: We shouldn't expose the base cache object.

Instead, if you want the elements, you should create a copy. We want to make sure a consumer doesn't have direct access to our base cache objects. They could end up doing something we don't want.

fartwhif: I debated this with myself, and another developer and we agreed that it should probably return the meat of it instead of only a certain property in case someone wants to do something more like show the coordinates in the list as well.  I started out having it return `IEnumerable<string>` and switched to `ConcurrentDictionary<string, PointsOfInterest>` after thinking about it further.  I understand the need to prevent data from being tainted but really IMHO none of the consumers should be doing that, and if they are it's a bug.  The alternative is to write accessors for every specific type of application within that layer which I'm sure you would agree is crap, or like you said copy the objects, which is almost as crappy.  If it were up to me I would not copy it and just remember not to try and do any write operations to it from other layers.  I'll make this change, just remember that I disagree with your philosophy.

Mag-nus: We don't need to deep copy the elements, we just don't want to return access to the base cache dictionary.

All of the various objects returned by the various world database caching mechanisms are the same, not clones.

Mag-nus: I think a simple line like this will suffice:

return new Dictionary<string, PointsOfInterest>(cachedPointsOfInterest);

fartwhif: I didn't know that, should have asked.  Thanks, I'll handle it.

Mag-nus: The individual elements will be shared by anyone using the cache, and yes, vulnerable to mutation.

This is how all of our world db cache works.

If you want to mutate those objects, you should then clone them. This would be the responsibility of the consumer, not the cache (as it is currently designed).

For weenies, we clone them into biotas and then mutate the biotas.

PR #1069, opened 2018-10-15
 Adding dispels and pressure plates

gmriggs:


Source/ACE.Server/WorldObjects/Door.cs
```diff
@@ -162,7 +162,7 @@ public override void ActOnUse(WorldObject worldObject)
             {
                 if (!IsOpen)
                     Open(worldObject.Guid);
-                else
+                else if (!(worldObject is PressurePlate))
```
LtRipley36706: `else if (!(worldObject is PressurePlate) && !(worldObject is Switch))`

PR #1070, opened 2018-10-19
Fixing jump in multiplayer, refactoring movement and animation systems for network and game layers

gmriggs:
- UniversalMotion has been replaced with a new Motion class, which has a much simpler API
- Fixed a multiplayer bug where jumping couldn't be seen by other players
- Added Network structures for all movement and animation related packets in Network.Motion namespace
- Fixed a bug where players performing standing longjumps would repeatedly appear to run ahead and snapback in multiplayer
- Improved parsing of contact/isgrounded/sticky/standinglongjump/ flags in movement packets
- Added network reader alignment
- Fixed a bug where creatures were broadcasting UpdatePosition messages when performing emotes while standing still
- Improved performance of GfxObjCache


Mag-nus:
Play tested with 32 clients actively running around the world. Seems ok.

Performance seems the same.
I can now see others jump.
Players appear to move the same as they did before.

PR #1072, opened 2018-10-21
Improving monster animation sync between server and client

gmriggs:


Mag-nus:
playtested with 32 clients, seems ok.

PR #1081, opened 2018-10-30
More item destruction fixes

Mag-nus:
Retain TimeToRot magic values of 0 and -1. They may have come from the weenie.

Destroy stacks that are consumed.

Destroy consumables when they're consumed.


Mag-nus:
Can you please create an issue for that as well? I am able to duplicate it.

I think it's an issue with the notification system. I tested two of the same items, one through the process you mentioned and one just held in my inventory the whole time. The item properties were the same for all properties that mattered, between the two.

PR #1086, opened 2018-11-01
Fixes for Anniversary Event emotes

LtRipley36706:


Source/ACE.Server/Program.cs
```diff
@@ -110,6 +110,9 @@ public static void Main(string[] args)
             log.Info("Initializing WorldManager...");
             WorldManager.Initialize();
 
+            log.Info("Preloading Global Event Landblocks...");
+            LandblockManager.PreloadGlobalEventLandblocks();
```
gmriggs: There should be a config option for this

If I am testing non-event related stuff, I would prefer not to load Hebian-To

gmriggs: Also, this should probably just be nixed completely, as there already exists a landblock preloading / permaload system

The existing landblock preloading system should be upgraded so the landblocks is pre-loads is determined by a config file, instead of those landblocks being hardcoded

LtRipley36706: It's a pretty important system, but i can just make the default option be load instead of not load...

I disagree about removing completely as there should be two seperate things.. One is Hebian-To and possibly any other landblocks found to have global things on them, and then the custom or server op configs landblock loads for things like common towns or high-traffic blocks.. 

I'm not going to touch upon that for this fix.. just want to get the global event block loaded by default

gmriggs: That's the purpose of the config files, and specifying which landblocks to permaload in there

The default server config file would have Hebian-To and any other landblocks that are needed. If the server op also wanted to permload other landblocks, such as all the towns, then they would add that to the config. ie. there would be a server.config with Hebian-To, a towns.config with all the towns, and the server op would specify which config files to load

Source/ACE.Server/Config.json.example
```diff
@@ -16,7 +16,8 @@
     "DatFilesDirectory": "c:\\ACE\\",
     "ShutdownInterval": "60",
     "WorldDatabasePrecaching": false,
-    "LandblockPreloading": false
+    "LandblockPreloading": false,
+    "GlobalEventLandblockPreloading": true
```
gmriggs: I realize that this is a rather simple change and you just want to get this out as quick as possible, but this is really duplicating the work of the existing LandblockPreloading system. Instead of fixing / improving the existing system so that it is using a config file, it is ignoring and bypassing that system, and adding a duplicate version of that system to load 1 specific landblock.

This is adding more cruft to the code and making things more complicated than it should be, and will just have to be fixed later. I would rather just apply the proper fix for this now, and improve the existing LandblockPreloading system to use config files, as originally described in this comment:

// TODO: Change the RawLandblockId list used for preloading defined landblocks to some other, more easily-modified format, rather than a compiled uint array

LtRipley36706:
Further changes made to implement configurable preload system. Less than thrilled about not having an easy way to default load Hebian-To specifically if config file isn't updated as shown in example.

LtRipley36706:
fixed my default fall back concern in the null check...

Source/ACE.Server/Config.json.example
```diff
@@ -16,7 +16,44 @@
...
     "ShutdownInterval": "60",
     "WorldDatabasePrecaching": false,
-    "LandblockPreloading": false
+    "LandblockPreloading": true,
+    "PreloadedLandblocks": [
+        {
+            "Id": "E74EFFFF",
+            "Description": "Hebian-To (Global Events)",
+            "Permaload": true,
+            "IncludeAdjacents": true,
```
gmriggs: should adjacents default to true here?

PR #1090, opened 2018-11-09
Improved ShadowPart performance

gmriggs:


Mag-nus:
I approve this, but, some comments would be nice to describe why we're simply returning immediately in AddPart and RemovePart.

A dev will most likely stumble on this code eventually and ask the question.

PR #1092, opened 2018-11-12
 Improving DatLoader for third-party apps

gmriggs:


Source/ACE.DatLoader/UnpackableExtensions.cs
```diff
@@ -152,9 +152,9 @@ public static void Unpack(this List<uint> value, BinaryReader reader)
         /// <summary>
         /// A list that uses a Int32 for the length.
         /// </summary>
-        public static void Unpack<T>(this List<T> value, BinaryReader reader) where T : IUnpackable, new()
```
Mag-nus: I believe UnpackSmartArray already covers this case.

Source/ACE.DatLoader/UnpackableExtensions.cs
```diff
@@ -154,7 +154,7 @@ public static void Unpack(this List<uint> value, BinaryReader reader)
         /// </summary>
         public static void Unpack<T>(this List<T> value, BinaryReader reader) where T : IUnpackable, new()
         {
-            var totalObjects = reader.ReadInt32();
+            var totalObjects = (uint)reader.ReadInt32();
```
Mag-nus: You could just change this to ReadUInt32(), then you wouldn't need the cast.

PR #1097, opened 2018-11-18
keepOpen option added to DatDatabase FileStream

Mag-nus:
Default is off, which should act exactly as ACE did before.

The use of this feature is set in Program.cs via the following line:
DatManager.Initialize(ConfigManager.Config.Server.DatFilesDirectory, true);

I did not actually profile this, so, I don't know if reusing the filestream actually makes a difference. This is the reason for the **NEEDS TESTING** tag.

### DIFF
--- a/Source/ACE.DatLoader/CellDatDatabase.cs
+++ b/Source/ACE.DatLoader/CellDatDatabase.cs
@@ -5,7 +5,7 @@ namespace ACE.DatLoader
 {
     public class CellDatDatabase : DatDatabase
     {
-        public CellDatDatabase(string filename) : base(filename)
+        public CellDatDatabase(string filename, bool keepOpen = false) : base(filename, keepOpen)
         {
         }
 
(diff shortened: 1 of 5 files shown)


Mag-nus:
Couple of quick tests:

Dat's are on an intel NVME SSD.

Reading every file (but not unpacking) from both dats **with** keep open:
Elapsed: 12.93s
Elapsed: 12.80s
Elapsed: 12.78s
Elapsed: 12.80s
Elapsed: 13.14s
Elapsed: 12.84s
Elapsed: 12.82s
Elapsed: 12.84s

Reading every file (but not unpacking) from both dats **without** keep open:
Elapsed: 68.01s
Elapsed: 67.70s
Elapsed: 67.26s
Elapsed: 67.57s
Elapsed: 67.35s
Elapsed: 68.21s
Elapsed: 67.60s


Seems like this PR should add some performance to ACE.

PR #1098, opened 2018-11-19
Fix Door/Chest appraisal issue

LtRipley36706:
Fix issue with appraisal of Doors and Chests with respect to lockpick chances, resists and false "bonus"

### DIFF
--- a/Source/ACE.Server/Managers/WorldManager.cs
+++ b/Source/ACE.Server/Managers/WorldManager.cs
@@ -451,7 +451,7 @@ namespace ACE.Server.Managers
             if (character.TotalLogins <= 1 || PropertyManager.GetBool("alwaysshowwelcome").Item)
             {
                 // check the value of the welcome message. Only display it if it is not empty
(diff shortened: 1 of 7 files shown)


Source/ACE.Server/Managers/WorldManager.cs
```diff
@@ -451,7 +451,7 @@ private static void DoPlayerEnterWorld(Session session, Character character, Pla
             if (character.TotalLogins <= 1 || PropertyManager.GetBool("alwaysshowwelcome").Item)
             {
                 // check the value of the welcome message. Only display it if it is not empty
-                string welcomeHeader = ConfigManager.Config.Server.Welcome ?? "Welcome to Asheron's Call!";
+                string welcomeHeader = ConfigManager.Config.Server.Welcome != "" ? ConfigManager.Config.Server.Welcome : "Welcome to Asheron's Call!";
```
Mag-nus: If you're going to change from null to "" check, it should be a String.IsEmpty() check.

PR #1102, opened 2018-11-24
Adding scrolls to loot profiles

dgatewood:
Scroll weenies are cached into lists, then created from those weenies, so the database isnt hit everytime there is a new scroll created.


dgatewood:
Dont merge yet. I need to test some things.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -73,6 +73,113 @@ public uint GetMaxGuidFoundInRange(uint min, uint max)
...
+                    }
+                    else
+                    {
+                        var context = new WorldDbContext();
+                        var result = from weenie in context.Weenie
+                                     join did in context.WeeniePropertiesDID on weenie.ClassId equals did.ObjectId
+                                     where weenie.Type == 34 && did.Type == 28 && did.Value == spellID
+                                     select weenie;
+                        level1Scrolls[spellID] = result.FirstOrDefault();
+                        return result.FirstOrDefault();
```
Mag-nus: Just a few nitpicks.

result should probably be renamed query.

Instead of returning result.FirstOrDefault(), you should return level1Scrolls[spellID] to avoid running the query in duplicate.

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -73,6 +73,113 @@ public uint GetMaxGuidFoundInRange(uint min, uint max)
...
         }
 
+        private static ConcurrentDictionary<uint, Weenie> level1Scrolls = new ConcurrentDictionary<uint, Weenie>();
+        private static ConcurrentDictionary<uint, Weenie> level2Scrolls = new ConcurrentDictionary<uint, Weenie>();
+        private static ConcurrentDictionary<uint, Weenie> level3Scrolls = new ConcurrentDictionary<uint, Weenie>();
+        private static ConcurrentDictionary<uint, Weenie> level4Scrolls = new ConcurrentDictionary<uint, Weenie>();
+        private static ConcurrentDictionary<uint, Weenie> level5Scrolls = new ConcurrentDictionary<uint, Weenie>();
+        private static ConcurrentDictionary<uint, Weenie> level6Scrolls = new ConcurrentDictionary<uint, Weenie>();
+
+        public Weenie GetSpellWeenie(uint tier, uint spellID)
```
Mag-nus: I think this should take a parameter like "worldIsPreloaded". If worldIsPreloaded, you wouldn't actually query the database, but instead would query the cached weenies.

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -2317,6 +2338,42 @@ public static int CreateFood()
             return foodType;
         }
 
+        public static WorldObject CreateRandomScroll(int tier)
+        {
+            uint weenieID;
+            WorldDatabase wDb = new WorldDatabase();
```
Mag-nus: You can reference the world database via DatabaseManager.

We shouldn't be instantiating the databases in any other place, other than test projects.

PR #1111, opened 2018-12-02
Vendors work again, with nearly no change.

dgatewood:
Get's rid of old code. Removes item from player with networking, and saves item to database once it has been transfered to Vendor.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Commerce.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Commerce.cs
@@ -242,20 +242,16 @@ namespace ACE.Server.WorldObjects
                 else
                 {
                     // remove item from inventory.
-                    TryRemoveFromInventory(profile.Guid);
+                    TryRemoveFromInventoryWithNetworking(item);
                 }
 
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Player_Commerce.cs
```diff
@@ -242,20 +242,16 @@ public void HandleActionSellItem(List<ItemProfile> itemprofiles, ObjectGuid vend
...
-
-                // clean up the shard database.
-                throw new NotImplementedException();
-                // todo fix for EF
-                //DatabaseManager.Shard.DeleteObject(item.SnapShotOfAceObject(), null);
-
                 Session.Network.EnqueueSend(new GameMessageDeleteObject(item));
                 purchaselist.Add(item);
+                //update item in database
+                item.SaveBiotaToDatabase();
```
Mag-nus: Can you add the comments here similar to the comments above the other item.SaveBiotaToDatabase in player_inventory

In the future, i'm HOPING to add an improved way to save just the parent property id for these items which should be much smaller hits on the db than saving the whole object like SaveBiotaToDatabase() does currently.

I think my note basically talked about that.

PR #1115, opened 2018-12-03
Profiling based improvements

Mag-nus:
Removed the last reflection based skill lookups. Now all skill lookups go through the portal.dat
This reduces CPU demand by about 2% or more.

Removed the legacy attributes for Skill/PropertyAttribute2nd. These were to allow us to define these enum values with meta data before when we weren't loading the data from the portal.dat. Now that the data comes from the portal.dat, we do not need these. They are being removed to avoid confusion.

Removed the PhysicsObj != null check for the Landblock.GetPhysicsWorldObjects().
This reduces CPU demand by about 6%.


Mag-nus:
Both issues have been addressed.

For the unusable skills, they're still in the untrained state, so they're displayed with that list. They now just have the [Unusable] tag added to the end. They are sorted in the order as they appear in the client.

gmriggs:
Bow: 0
Mace: 0
Sword: 0
UnarmedCombat: 0

These skills are showing up in Player.Skills as 'untrained', instead of 'inactive'

PR #1117, opened 2018-12-04
Adding summoning from @mcreedjr

gmriggs:


gmriggs:
@LtRipley36706
The database objects have been moved to 16py repo as suggested

PR #1122, opened 2018-12-08
fixing: some clients unable to connect via WAN

fartwhif:
This problem appeared as: connection vial full and update vial empty until client disconnects.

primary change:
fixed a connection issue for some client firewalls guarding clients connecting over WAN by changing port strategy from two unidirectional ports to a single bidirectional port with the exception of the connect response packet and CICMDCommand packets

other changes:
changed to using a random cookie for the three-way handshake
using random initialization vectors for the CRC stream ciphers instead of a constant and known IV


Source/ACE.Server/Managers/WorldManager.cs
```diff
@@ -78,9 +81,55 @@ public static void Initialize()
...
+                {
+                    PacketInboundConnectResponse connectResponse = new PacketInboundConnectResponse(packet);
+
+                    // This should be set on the second packet to the server from the client.
+                    // This completes the three-way handshake.
+                    // the client should be sending it back now
+                    // This could be offloaded into a queue and processed on another thread
+                    // TO-DO: generate ban entries here, situation rate
+                    // TO-DO: fix race condition between bob and eve getting and setting session state
+                    var session =
```
Mag-nus: when iterating over the sessionMap, we must enter a:
sessionLock.EnterReadLock();

Alternatively you can create a private Find(args...) function below to find the specific session you're looking for, following the pattern the other Find() functions use.

PR #1136, opened 2018-12-13
Various improvements resulting from excessive load testing

Mag-nus:
Thread safety added to ObjectMaint.ServerObjects
- This is used by every WorldObject ctor. WorldObjects are created on multiple threads

More Session and Player.LogOff cleanup
- This fixes a few rare scenarios

SocketManager changed from list to array.
- This is pretty minor. It might offer a very minor performance improvement as it's used by every packet transmission
- It's mostly just form. There's 2 sockets, they're fixed, and the array is never altered.

Vector3 improvements
- Vector3 is a struct, but in some cases was being treated like a class
- Removed Vector3 extensions that are covered by native .NET core functions

InterpolationManager
- Change PositionQueue from Queue to a LinkedList
- InterpolationManager isn't even used but if need a collection where you want the first, indexed, and last item, and require insertions/deletions, LinkedList is the proper tool.

Enum .HasFlag optimizations
- Changed 2 places that use HasFlag to a & != 0 pattern.
- This ONLY helps performance in debug mode (when the optimize flag is not set)
- We should continue to use .HasFlag() except in cases where the code is close to the core and impacts performance heavily.

### DIFF
--- a/Source/ACE.Server/Physics/PhysicsObj.cs
+++ b/Source/ACE.Server/Physics/PhysicsObj.cs
@@ -8,7 +8,6 @@ using ACE.Server.Physics.Animation;
 using ACE.Server.Physics.Collision;
 using ACE.Server.Physics.Combat;
 using ACE.Server.Physics.Common;
-using ACE.Server.Physics.Extensions;
 using ACE.Server.Physics.Hooks;
 using ACE.Server.WorldObjects;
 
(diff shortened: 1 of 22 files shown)


Source/ACE.Server/Physics/PhysicsObj.cs
```diff
@@ -1481,10 +1480,10 @@ public void UpdateObjectInternal(double quantum)
                     if (State.HasFlag(PhysicsState.AlignPath))
                     {
                         var diff = newPos.Frame.Origin - Position.Frame.Origin;
-                        newPos.Frame.set_vector_heading(diff.Normalize());
```
gmriggs: As already mentioned, change the implementation if needed, not the API

.Normalize() was added for a cleaner API

The implementation is also the same as Vector3.Normalize, the only difference is one uses inlining

If the compiler isn't already inlining these 1-line functions, then the attribute should be added for many of them

PR #1155, opened 2018-12-19
various fixes

fartwhif:
command parser and invoke resiliency
fix /grantxp command so it doesn't overflow


Source/ACE.Server/Network/NetworkSession.cs
```diff
@@ -689,7 +689,7 @@ private void SendPacketRaw(ServerPacket packet)
                 var sb = new StringBuilder();
                 sb.AppendLine(ex.ToString());
                 sb.AppendLine(String.Format("[{5}] Sending Packet (Len: {0}) [{1}:{2}=>{3}:{4}]", payload.Length, listenerEndpoint.Address, listenerEndpoint.Port, session.EndPoint.Address, session.EndPoint.Port, session.Network.ClientId));
-                log.Error(sb.ToString());
+                packetLog.Error(sb.ToString());
```
Mag-nus: This should remain using the standard logger

The packetLog is a very specific use case, and mainly just for logging packets.

On most servers, the packetlog will be disabled.

fartwhif: ahh ok, my fault, thought it was a typo or something

PR #1162, opened 2018-12-21
Moved ACE.Server.Physics.Common.Random to ACE.Common.Random

fartwhif:
Changed Random wrapper to instantiate a new Random when needed and removed the heavyweight wrapping


fartwhif:
We should see a point or two of improved performance from the removal of the thread discriminator.  .NET Core Random is different than .NET Framework in that it should no longer be needed.

fartwhif:
`OldRandom.RollDice(int min, int max): 00:00:10.0000056, iterations: 134002807
OldRandom.RollDice(float min, float max): 00:00:10.0000075, iterations: 137180787
OldRandom.RollDice(uint min, uint max): 00:00:10.0000001, iterations: 136066119
ThreadSafeRandom.Next(int min, int max): 00:00:10.0000001, iterations: 225691575
ThreadSafeRandom.Next(float min, float max): 00:00:10.0000003, iterations: 231988872
ThreadSafeRandom.Next(uint min, uint max): 00:00:10.0000002, iterations: 228039807`

Source/ACE.Common/Random.cs
```diff
@@ -0,0 +1,42 @@
+namespace ACE.Common
+{
+    // important class, ensure unit tests pass for this
+    public static class ThreadSafeRandom
```
Mag-nus: This file should be named ThreadSafeRandom.cs

PR #1163, opened 2018-12-21
Adding sticky melee distance to charge attack

gmriggs:


Source/ACE.Server/WorldObjects/Player_Melee.cs
```diff
@@ -287,5 +287,13 @@ public bool IsMeleeDistance(WorldObject target)
 
             return cylDist <= 0.6f;
         }
+
+        public bool IsStickyDistance(WorldObject target)
+        {
+            var cylDist = (float)Physics.Common.Position.CylinderDistance(PhysicsObj.GetRadius(), PhysicsObj.GetHeight(), PhysicsObj.Position,
+                target.PhysicsObj.GetRadius(), target.PhysicsObj.GetHeight(), target.PhysicsObj.Position);
+
+            return cylDist <= 4.0f;
```
fartwhif: 10 feels about right:
`            return cylDist <= 10.0f;`

gmriggs: We need some additional video and pcap demonstration of this

Melee players 'teleporting' 10 meters seems a bit extreme..

fartwhif: not sure, leave at 4 until we can see what the record shows

PR #1184, opened 2018-12-26
Improving recipe manager, adding tinkering and imbues to crafting system

gmriggs:


dgatewood:
Looks good. I think we went about this the same way. I would say this is ready for merge.

Issue #1189, opened 2018-12-28
Food not Casting Spells


Warloxx01:
### Bug Report or Feature Request (mark with an `x`)
```
- [X ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
 have been playing around with the food(18) system and found that when eating food(18) that restores HP/Mana/Stam  work fine but foods(18) that cast spells do not. they want to add HP, Mana or Stam and don't cast the spell but they are consumed. Now Food(18) that is created as a Gem(38) cast the spell but are NOT consumed. I have created the Candy Corn treat that casts the spell Sweet Speed (Had to create that spell myself, as it was not implemented yet) and it casts fine listed as a Gem (38) but not as a food (18). I looked in the food.cs and found code that deals with food and buffs, but being im no real coder I can't tell if it is complete or not.  Attached is 03860 Sweet Speed.sql and 32207 Candy Corn.sql for testing.
[AC Stuff.zip](https://github.com/ACEmulator/ACE/files/2714091/AC.Stuff.zip)

### Desired functionality.
When the Candy Corn is Eaten it is consumed and the spell is cast.

What would like to see implemented?
What is the use case?
-->

PR #1190, opened 2018-12-28
Updating consumables

gmriggs:
This resolves https://github.com/ACEmulator/ACE/issues/1189


Warloxx01:
Seems to work ( animation and is Consumed ) but it now reads in the Message window "Candy Corn applies Sweet Speed on you." and also now reads out in Green instead of Blue. In Player.cs I found this line...
buffMessage = new GameMessageSystemChat($"{consumableName} applies {spell.Name} on you.", ChatMessageType.Craft);

around Line 886. Show that be changed to casts?

gmriggs:
Yep, good catch

Updated PR

PR #1197, opened 2018-12-30
Major Inventory Refactor

Mag-nus:
This fixes many of the inventory related issues.
It also simplifies the inventory code a bit.

It almost certainly introduces some new issues, so testing/feedback will be required.

Some none related inventory fixes:
- Fix PlayerEnterWorld exception when player has null Location
- Remove IComparable from WorldObject
- Fix item decaying explosions
- Don't decay IsStuck objects


Mag-nus:
Fix: Missile ammo doesn't appear correctly in combat stance.

Warloxx01:
I have applied the 44 Files to my Server (latest build) and so far it appears to be good, My inventory is no longer spread all over hells half acre after a log out/ Login. However I tried to make a new character to check the Attuned and Bonded fix, I found I was UNable to give Jonathan My academy Exit Token, I was able to change that setting (114) Item (29335) in the DataBase from 1 to 0 and he excepted it. on a Good note I am NOT able to drop my Imp staff any longer. still testing more.

kdmxp256:
The if statement at line 1438, within function GiveObjecttoNPC(), in Player_Inventory.cs is preventing giving Attuned items to NPCs.  I believe Attuned was only directed towards player.  Case and point are the Town Criers, which are AC's version of garbage collectors.  Mostly, I believe NPCs will reject items, whether will a blanket statement about not knowing what to do with item or with custom flavor text, if they aren't specifically set up via emotes to accept said items.

Source/ACE.Server/WorldObjects/Gem.cs
```diff
@@ -77,10 +73,10 @@ public override void UseItem(Player player)
                     castMessage = "The gem casts Asheron's Lesser Benediction on you";
                 */
 
-                player.CreateItemSpell(Guid, (uint)SpellDID);
+                player.CreateItemSpell(this, (uint)SpellDID);
 
                 // always consume gem?
```
kdmxp256: There are instances of Gems that are perpetual, such as Blackmoor's Favor: https://pastebin.com/GaBckVtC.  According to the PCAP data, bool property 63 was used to denote unlimited usage.

Mag-nus: I'm not sure what you're getting at. Does a change need to be made?

kdmxp256: If the gem has PropertyBool.UnlimitedUse, as with the case of the Blackmoor's Favor gem that is given to all players since TOD, the gem needs to not be consumed on use, so it can be used over and over again.

kdmxp256: http://asheron.wikia.com/wiki/Blackmoor%27s_Favor

Source/ACE.Server/WorldObjects/Gem.cs
```diff
@@ -77,10 +73,10 @@ public override void UseItem(Player player)
...
-                player.TryRemoveFromInventoryWithNetworking(this);
+                player.TryConsumeFromInventoryWithNetworking(this);
 
                 player.SendUseDoneEvent();
                 return;
```
kdmxp256: Should have included a little more in the code reference.  For gems with unlimited usage, they should be excluded from player.TryConsumeFromInventoryWithNetworking(this).

Mag-nus: gotcha!, should be fixed now *i think*

Warloxx01:
I Started Back to work today but I did find that after you create a new char and equip his/her wand (after talking to Jonathan and receiving the oils and coat) then double click the oil and apply it to the wand (still equiped) that it will place the magic academy wand in your pack but doesnt remove the euiped wand. Once you try to remove it by drag and drop or double clicking the Academy wand to replace it nothing happens, it also can not be dropped on a Town Crier. Once you log out of the Game (fully not just to menu screen) and log back in the wand is gone and the academy wand can be equiped. "Last I checked" that was a day or two ago.

Warloxx01:
Sorry I also forgot to add that (NOT every time) but now and then I will log back in and find myself in a different location then I logged out. Like I will go buy comps at the mage shop, run up to the LS and log out (all done in Holtburg) and when I log back in i will be back at the mage shop W/O the comps I bought before I logged out.

Mag-nus:
I'll check out the first issue you reported.

The second is likely unrelated. It sounds like there could be a db issue, the server isn't detecting the client logout properly and thus saving it, or the server isn't shutting down properly

Mag-nus:
Ok, addressed the final issue.

PR #1203, opened 2018-12-31
This fixes the crash on duplicate layers

Mag-nus:
It's not without its drawback though.

Items and player casted spells will prov have weird override/duration issues.


LtRipley36706:
closing this PR as #1206 supersedes it.

PR #1204, opened 2018-12-31
Character migrate/backup/import and Web API

fartwhif:
**Authentication**:<br>
The Web API uses [basic access authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) and requires a verified username and password to access some of the data resources.

If a resource is marked Anonymous then authorization isn't needed.  If you attempt to access a resource that isn't marked Anonymous without supplying valid credentials then you will get a 401 Unauthorized response.

Authenticated requests assume the accesslevel of the account authenticated as a series of claims.

If a resource is marked Requires: [Admin], that means unless your account includes the Admin claim then you won't be able to access the resource and you will get a 403 Forbidden response.

If a resource is marked RequiresAny: [Admin,Advocate,Developer,Envoy,Player,Sentinel] then unless your authenticated account claims at least one of the required claims you won't be able to access the resource and you will get a 403 Forbidden response.

**Web API**<br>
[documentation](https://documenter.getpostman.com/view/6422801/RznLHGgq)<br>
The Web API server listens at the base ACE port + 2, TCP, TLS, HTTPS, JSON, REST, depends on CryptoManager within ACE.Server project that Generates and saves a self-signed certificate for general purposes, when it can't find one in `C:\Users\<user>\AppData\Roaming\ACEmulator_<WorldName>\Certificates` on Windows or `/home/<user>/.config/ACEmulator_<WorldName>/Certificates` on Linux.  Self-Hosting using NancyFX/OWIN/Kestrel in its own project that depends on and wraps around ACE.Server.  WebSockets is a future possibility.  To run ACE with web API capability start the batch file: `\Source\ACE.Server.WebApi\bin\x64\Debug\netcoreapp2.1\start_server+webapi.bat`

**Warning**<br>
If the world name in the server configuration is changed then the current transfers and certificate are lost unless the files are moved to the new location.  To mitigate this move or copy the contents of `C:\Users\<user>\AppData\Roaming\ACEmulator_<OldWorldName>\` to `C:\Users\<user>\AppData\Roaming\ACEmulator_<NewWorldName>\` replacing `<user>` with the applicable username, and `<OldWorldName>` and `<NewWorldName>` with applicable names.  For linux the path will usually be: `/home/<user>/.config/ACEmulator_<WorldName>/`.  If your world name contains special characters then the folder names will have them stripped out, allowed world name to folder name characters are: `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`.

**Migrations**:<br>

1. Via Web API, a request is made to migrate a character to another server.
2. Character snapshot is packaged and character is  placed in a "frozen" state.
3. Requester makes an API call to get the list of characters, of which is the migrating character in frozen state.
* At any time before the next step is initiated, or if the target server rejects the source server, an API request can be made to the source server to unfreeze the character and cancel the migration.
4. Requester makes a request to the target server (the character's new home).
5. Target server makes sure source server is trusted before downloading and rejects the request if it doesn't.
6. Target server downloads packaged snapshot from source server.
7. Source server permanently deletes character upon package download completion.
8. Target server performs cryptographic verification of the snapshot, and rejects it if it's been forged.
9. Target server adds the character to the requester's account.

Note:  Allowing migrations from itself enables cross account transfers and character renaming.

**Backups**:<br>
Via Web API, a request is made to backup a character.
Character snapshot is packaged and returned in the response to the requester.

**Imports**:<br>
**with imports enabled the exploitation flood gates are essentially open**<br>
requester makes a request to the target server API including in the request the snapshot package encoded in base64 and the new character name.
target server adds the character to the requester's account.

**Recommended settings for flawless contract groups:**<br>
allow backup: yes<br>
allow migrate: yes<br>
allow import: no<br>

**Limitations**:<br>
- ACE specific data constructs such as character options are not exported/imported.  This is to keep it as compatible as possible with LSD.

**Disclaimer**<br>
The migration feature requires the assumption that all server operators involved are non defectors.  A defector is someone in the trust group who does malicious or negligent action of any kind.  Providing a system that works for flawless trust groups is easy, but as soon as an attempt is made to eliminate the human element and commission the possibility of defectors the entire system breaks.  Manual approval process and/or central "antivirus" server may be something that someone might want to come up with in the future, but that's NOT what this is meant for.  **_As a server operator, DO NOT blame anyone but yourself upon defection(s) when using this system.  NOBODY BUT YOURSELF IS LIABLE_**


Source/ACE.Server/Managers/PlayerManager.cs
```diff
@@ -82,11 +82,20 @@ public static void SaveOfflinePlayersWithChanges()
...
         public static void AddOfflinePlayer(Player player)
+        {
+            AddOfflinePlayer(player.Biota);
+        }
+
+        /// <summary>
+        /// This would be used when a new player is created after the server has started.
+        /// When a new Player is created, they're created in an offline state, and then set to online shortly after as the login sequence continues.
+        /// </summary>
+        public static void AddOfflinePlayer(Biota bioPlayer)
```
Mag-nus: bioPlayer -> playerBiota

Source/ACE.Server/Managers/TransferManager.cs
```diff
@@ -0,0 +1,788 @@
+using System;
```
Mag-nus: I *THINK* I'd prefer this file in the following namespace instead:

Source/ACE.Server/TransferServer/

Source/ACE.Server/ACE.Server.csproj
```diff
@@ -16,6 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="HttpMachine" Version="0.9.0" />
```
Mag-nus: Is there a way to do this with a package provided by microsoft?

Source/ACE.Common/ThreadSafeRandom.cs
```diff
@@ -38,5 +40,18 @@ public static uint Next(uint min, uint max)
                 return (uint)random.Next((int)min, (int)(max + 1));
             }
         }
+        public const string CookieChars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
Mag-nus: This prob doesn't need to go into ThreadSafeRandom and can stay local to the TransferServer namespace

Source/ACE.Server/ACE.Server.csproj
fartwhif: There's no .NET library to use for this outright.  I guess I'll roll my own parser for it.

Here's where I got this information:  https://stackoverflow.com/questions/975308/c-sharp-http-request-parser

Source/ACE.Server/Managers/TransferManager.cs
fartwhif: I don't.  The server is just a server, but the transfer manager is an ACE feature, not a server feature.  I know you don't like having an HTTP server running in ACE but that doesn't mean features dependent on the HTTP server are the HTTP server.  This is especially the case if we tie TransferManager into the AC server as well, or instead of, you wouldn't want TransferManager to be moved into `ACE.Networking`.

Source/ACE.Server/Managers/PlayerManager.cs
fartwhif: will do

Source/ACE.sln
```diff
@@ -43,48 +43,78 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ACE.Server", "ACE.Server\AC
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ACE.Adapter", "ACE.Adapter\ACE.Adapter.csproj", "{3C925D75-B149-4898-8806-054D3E63A4BE}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ACE.Server.WebApi", "ACE.Server.WebApi\ACE.Server.WebApi.csproj", "{F79B3B88-7473-45D6-81FC-FE048254B628}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
```
Mag-nus: Any CPU needs to be removed.

Must be Release|x64 and Debug|x64 only

fartwhif: done

esoterick:
Couple of concerns... 

Generally speaking API's are meant to be utilized by machines not users.  Therefor Basic Authentication isn't usually the best form of Auth for them.  Not to mention Basic Auth provides no Authorization components.  So if you plan on adding multiple administrative components directly to the server I would recommend also adding a proper RBAC and some sort of API token.

The following concerns me
6. Target server downloads packaged snapshot from source server.
7. Source server permanently deletes character upon package download completion.
8. Target server performs cryptographic verification of the snapshot, and rejects it if it's been forged.
9. Target server adds the character to the requester's account.

If there is any issue in steps 7 and 8 the user is out of luck.  The origin server deleted the origin char and the destination server failed to extract the snapshot of the user's migration.  I feel like before anything is deleted both sites are checked for consistency.  Once verified drop origin, and toggle/enabled destination.

It also seems like there are numerous requests made to both servers for the migration.  It seems like there shouldn't need to be.  If trust is established between the two servers, the client should only have to make one a single request origin or destination it doesn't matter.  The servers can then communicate across their own common api and handle the workflow themselfs.

It would also be helpful prior to merging in to the master branch to have this PR squashed to one commit.

Source/ACE.Server/Managers/TransferManager.cs
```diff
@@ -0,0 +1,1152 @@
...
+                try
+                {
+                    unverifiedThumbprints = JsonConvert.DeserializeObject<TransferManagerTransferConfigResponseModel>(unverifiedThumbprintsSerialized, serializationSettings);
+                }
+                catch
+                {
+                    return new ImportAndMigrateResult() { FailReason = ImportAndMigrateFailiureReason.ProtocolError };
+                }
+
+                if ((metadata.PackageType == PackageType.Migrate && !ConfigManager.Config.Transfer.AllowMigrationFrom.Any(k => k == unverifiedThumbprints.MyThumbprint)) || (metadata.PackageType == PackageType.Backup && !ConfigManager.Config.Transfer.AllowImportFrom.Any(k => k == unverifiedThumbprints.MyThumbprint)))
```
fartwhif: this needs to move outside the "migration only" branch, and run the test for imports too

Source/ACE.Server/Entity/Actions/ActionQueue.cs
```diff
@@ -30,5 +30,7 @@ public void EnqueueAction(IAction action)
         {
             Queue.Enqueue(action);
         }
+
+        public int Count { get { return Queue.Count; } }
```
Mag-nus: This is suspect. Getting the queue count is a very costly operation.

Alternatively, you can use IsEmpty.

Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -772,5 +772,8 @@ public bool IsHouseDungeon
                 return isHouseDungeon.Value;
             }
         }
+
+        public int ActionQueueCount { get { return actionQueue.Count; } }
```
Mag-nus: Another suspect line

Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -772,5 +772,8 @@ public bool IsHouseDungeon
...
+        public int ObjectCount { get { return worldObjects.Count; } }
```
Mag-nus: Suspect as well.

fartwhif:
replaced with https://github.com/ACEmulator/ACE/pull/1390

PR #1205, opened 2018-12-31
Update LootGenerationFactory.cs

dgatewood:
Updates max damages per tier to be variable value. Previously, all weapons defaulted to their max value. This changes it so they are based on a bell curve for values within a range.

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -17688,7 +17688,7 @@ namespace ACE.Server.Factories
                     tieredDamage = twohandedWeaponDamageTable[baseWeapon - 1, wieldPlace];
                     break;
             }
-            double chance = ThreadSafeRandom.Next(0.000f, 1.000f);
+            double chance = ThreadSafeRandom.Next(0.0f, 1.0f);
             if (tieredDamage < 10)
             {
 
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -17688,7 +17688,7 @@ namespace ACE.Server.Factories
                     tieredDamage = twohandedWeaponDamageTable[baseWeapon - 1, wieldPlace];
                     break;
             }
-            double chance = ThreadSafeRandom.Next(0, 1);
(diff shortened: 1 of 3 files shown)


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -17687,7 +17688,63 @@ public static int GetMaxDamage(int weaponType, int tier, int baseWeapon)
                     tieredDamage = twohandedWeaponDamageTable[baseWeapon - 1, wieldPlace];
                     break;
             }
-            return tieredDamage;
+            double chance = ThreadSafeRandom.Next(0, 1);
```
gmriggs: should be .Next(0.0f, 1.0f) here

Source/ACE.Server/WorldObjects/Creature_Death.cs
```diff
@@ -266,6 +269,18 @@ private void GenerateTreasure(Corpse corpse)
                     }
                 }
             }
+
+            //If creature didnt have a CreateList, then this creates the loot.
+            if(!hasCreateList && Level > 3)
```
LtRipley36706: Should we not be doing this check regardless of createlist being populated?

Based on what I read of the earlier discussion it would suggest both were evaluated

someeodtech: Both were being evaluated when I removed !hasCreateList

Source/ACE.Server/WorldObjects/Chest.cs
```diff
@@ -55,7 +55,9 @@ private void SetEphemeralValues()
             //Eventually these case statements would be linked to indivual treasure generators. Each one should be a different profile, but currently it will be the complete appropriate tier profile.
             for (int i = 0; i < GeneratorProfiles.Count; i++)
```
LtRipley36706: in the generator profiles.. if the where create hasflag RegenLocationType.Treasure
it should do a look up in the death treasure table for that value of the wcid
if its just Contain, it needs no look up and that's the actual wcid

right now the code below is assuming anything below 500 is a death treasure which could be wrong

dgatewood:
just created another PR. some things no longer needed to be touched from here, so just started clean.

Issue #1194, opened 2018-12-29
EF key lookup bugs


gmriggs:
### Bug Report or Feature Request (mark with an `x`)
```
- [* ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.

create a new char, /addallspells, log out

### The log given by the failure.

2018-12-29 14:26:18,371 ERROR: SaveBiota failed with exception: System.InvalidOperationException: The instance of entity type 'BiotaPropertiesSpellBook' cannot be tracked because another instance with the same key value for {'Id'} is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached. Consider using 'DbContextOptionsBuilder.EnableSensitiveDataLogging' to see the conflicting key values.
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.ThrowIdentityConflict(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry, Boolean updateDuplicate)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.StartTracking(InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntry.SetEntityState(EntityState oldState, EntityState newState, Boolean acceptChanges)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityGraphAttacher.PaintAction(EntityEntryGraphNode node, Boolean force)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.EntityEntryGraphIterator.TraverseGraph[TState](EntityEntryGraphNode node, TState state, Func`3 handleNode)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.NavigationFixer.NavigationCollectionChanged(InternalEntityEntry entry, INavigation navigation, IEnumerable`1 added, IEnumerable`1 removed)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntryNotifier.NavigationCollectionChanged(InternalEntityEntry entry, INavigation navigation, IEnumerable`1 added, IEnumerable`1 removed)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.ChangeDetector.DetectNavigationChange(InternalEntityEntry entry, INavigation navigation)
[...]

PR #1206, opened 2018-12-31
Improved enchantment layering

gmriggs:
This fixes some differences between the way ACE was handling Enchantment layers that differed from retail

Previously, for the 'surpassed' and 'surpassing' cases, ACE would not write enchantments if they were surpassed by an existing enchantment, or in the 'surpassing' case, it would modify the old enchantment, instead of adding a new layer.

The proper functionality here is to write the new enchantments in both cases, always increasing the layer id.

Also previously, ACE was sort of implicitly managing the most powerful enchantment as always being the top layer enchantment for that category (in GetEnchantments_TopLayer). Since it can no longer be assumed that the top layer id for category is the most powerful enchantment, this function has been modified so that it sorts by descending PowerLevel within the category, instead of descending LayerID

The enchantments db table has also been updated to index all of the fields we are searching on in the enchantment queries. This greatly improves the speed of the EnchantmentManager queries.

This also fixes some bugs with layering between creature-cast and item-cast enchantments, particularly when an item casts a spell that a creature already has. This was the cause of one of the EF duplicate key errors in https://github.com/ACEmulator/ACE/issues/1194, particularly the one reported by @fartwhif for the enchantments table.

The repro steps for this issue are:

/dispel
/buff &lt;self&gt; 3
/ci gemlugianarmor3
&lt;use gem&gt;
&lt;log out&gt;

This should fix some of the core architecture issues that https://github.com/ACEmulator/ACE/pull/1203 patched

### DIFF
--- a/Database/Updates/Shard/2018-12-31-00-Enchantment-Index.sql
+++ b/Database/Updates/Shard/2018-12-31-00-Enchantment-Index.sql
@@ -1,3 +1,0 @@
-ALTER TABLE `biota_properties_enchantment_registry` ADD INDEX `enchantment_Category_idx` (`enchantment_Category`);
-ALTER TABLE `biota_properties_enchantment_registry` ADD INDEX `power_Level_idx` (`power_Level`);
-ALTER TABLE `biota_properties_enchantment_registry` ADD INDEX `layer_Id_idx` (`layer_Id`);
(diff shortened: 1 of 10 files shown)


Database/Updates/Shard/2018-12-31-00-Enchantment-Index.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `biota_properties_enchantment_registry` ADD INDEX `enchantment_Category_idx` (`enchantment_Category`);
```
Mag-nus: Please show me where these indexes are actually used in ACE

PR #1230, opened 2019-01-04
Adding built-in spells to casting items

gmriggs:


mcreedjr:
For what it's worth, here is how I remember things:

- From what I remember, certain wands/staves had different wind-up animations. One that comes to mind is the Focusing Stone. IIRC the wind up animation had the Focusing Stone drawn over the caster's head and a drawn-out, almost slow-motion method before casting. There were no visible spell building 'bubble' in front of the caster that I recall
- It was my understanding that the Spellcraft _was_ equivalent to the skill level used to cast the spell. Wielding (and using) a wand or staff with a spell did not require the user to be trained in a school in order to use the spell embedded in the staff/wand as long as the met the wield requirements. IE if a staff had a level I war spell on it, the wielder would be able to use it even if they weren't trained in War Magic as long as they met the other wield/use requirements of the staff. The spell that was cast using the wand had the 'skill' level associated with the Spellcraft on the item
- And, yes, I agree mana was drawn from the casting implement and not the caster's mana

gmriggs:
Updated PR to use mana from the wand/staff/orb for built-in spells

For the windup/cast animations, to my current understanding of the spell data, this would perhaps come from the spell itself? IE, if a spell had a special cast motion, this would be determined by the talisman in the spell formula, and not necessarily on the casting item.

Looking at some retail pcaps, I found some examples of player's casting Priest's Curse, which I assume to be from the Imp Staff... it seemed like they were doing the regular windup/bolt cast animation there from the ones I found. Some videos here would really help match things up though.

For the item spellcraft, that does indeed make a lot of sense, especially if the player doesn't have a particular magic school trained for the spell. I left a comment for that part, haven't implemented it quite yet, because I wanted to verify the wield requirements for the casting item were being checked first... ie. didn't want level 1 players able to wield some powerful casting item and instantly get level 300 attack skill for that spell =)

gmriggs:
Also guessing the built-in spells wouldn't check or consume any spell components?

mcreedjr:
> Also guessing the built-in spells wouldn't check or consume any spell components?

Definitely correct. Which is also dovetails into my reasoning why the caster doesn't need to have the magic school trained. None of the typical checks are performed when a spell is cast from a staff/wand/item

gmriggs:
Ok, I think a lot of these points should be accounted for now:

- The casting implements with built-in spells did indeed have custom casting animations. These were found in PropertyDID.UseUserAnimation.

- The War Magic: 270 activation requirement was not showing up in the appraisal window. This is another field that required munging in AppraiseInfo, where PropertyDID.ItemSkillLimit is converted into PropertyInt.AppraisalItemSkill for the client display

- The activation checks were added to augment the wield checks. If the player doesn't meet the activation requirements (Arcane Lore 50, War Magic 270) for the Imp Staff, they can still wield it, but they now get the appropriate error when trying to cast the Priest's Curse spell

- The amount of mana consumed from casting Priest's Curse now comes from the ItemManaCost field. As per the staff description, this is reduce with Mana Conversion

- Casting Priest's Curse now uses the Item Spellcraft as the attack skill level

Open questions:

 - Did casting Priest's Curse have any chance of fizzle?

PR #1234, opened 2019-01-05
Adding imbues to damage formulas: Critical Strike, Crippling Blow, Armor Rending, Resistance Rending

gmriggs:


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -887,8 +887,13 @@ private bool DoHandleActionGetAndWieldItem(WorldObject item, Container itemRootO
             return true;
         }
 
+        public static bool UseWieldRequirements = true;
```
Mag-nus: This can't be static, it will break if work is done on multiple threads.

Mag-nus: This can't be static, it will break if work is done on multiple threads.

Mag-nus: Follow up question, where is this set to false?

gmriggs: this is an internal debugging only var

Mag-nus: ahh, ok. A comment on it would be good. I saw it before during the refactor and was confused.

PR #1238, opened 2019-01-06
Store and Use DIDs for Portal magics

LtRipley36706:
Instead of just saving positions, we need to store as DIDs and do proper lookups back to the original weenie for restrictions and other variables.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Monster_Awareness.cs
+++ b/Source/ACE.Server/WorldObjects/Monster_Awareness.cs
@@ -80,6 +80,7 @@ namespace ACE.Server.WorldObjects
         {
             return;
 
+            /*
             var currentTime = Timers.RunningTime;
 
             if (currentTime < NextFindTarget)
@@ -92,6 +93,7 @@ namespace ACE.Server.WorldObjects
                     return;
             }
             FindNextTarget();
+            */
         }
 
         public void SetNextTargetTime()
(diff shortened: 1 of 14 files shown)


Source/ACE.Server/WorldObjects/Monster_Awareness.cs
```diff
@@ -80,6 +80,7 @@ public virtual void HandleFindTarget()
         {
             return;
 
+            /*
```
mcreedjr: If you don't mind my asking, would you mind sharing what error you resolved by commenting out this code block? I believe this is a fairly critical piece of the monster 'think' logic.

LtRipley36706: there's a `return;` line just above the commenting so nothing is getting past that making this code useless. I wanted to clean up the existing warnings that have been lingering for months

mcreedjr: Got it. I'm sorry. I didn't 'review' thoroughly enough to see that! Thank you for taking the time to clean up those lingering warnings at compile time!

LtRipley36706: Not a problem!

gmriggs: This is WIP monster aggro code... it's not being used yet, so it just returns immediately

Commenting it out for now is a good idea

PR #1242, opened 2019-01-06
More profiling tweaks

Mag-nus:


Source/ACE.Server/WorldObjects/WorldObject_Tick.cs
```diff
@@ -56,26 +63,24 @@ public void EnqueueAction(IAction action)
         /// </summary>
         public void QueueFirstHeartbeat(double currentUnixTime)
         {
-            var delay = ThreadSafeRandom.Next(0.0f, DefaultHeartbeatInterval);
+            var delay = ThreadSafeRandom.Next(0.0f, (float)CachedHeartbeatInterval);
```
gmriggs: this should be DefaultHeartbeatInterval

generators get their heartbeats from RegenerationInterval, and this can regularly be 1 mins, 3 mins, 5 mins, 10 mins for generators

the intention of this code was just to spread the heartbeat ticks out a little over a 0-5s range, but with the way objects are loaded in, its possible this line can be removed completely

Mag-nus:
This is ready to be merged.

PR #1276, opened 2019-01-15
Missiles aren't childeren. Return to peace when dequipping ammo

Mag-nus:
Issues that still exist:

Player_Inventory.HandleActionGetAndWieldItem() needs a way to prevent further actions when the player is currently in a busy state (animating from a previous action)

If you are in combat mode, with your arrow visible, and someone logs in next to you, they will not see your arrow. *FIXED*


Mag-nus:
So, missiles are children but only when in combat stance.

Issue #1272, opened 2019-01-14
Can wear multiple pieces of armor that cover the same areas


DrudgeRobber:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```
If you put on, for example, leggings, which only cover the upper and lower legs, you can drag another pair of leggings which also cover the abdomen and wear both at the same time. Seems to work the same for any piece of armor that covers multiple areas. Breastplate - drag coat to arms and wear both the breastplate and coat. Etc. Video included.

[twopantsforme.zip](https://github.com/ACEmulator/ACE/files/2753903/twopantsforme.zip)

### Repro steps.

Get a pair of leggings which cover only the upper and lower legs. Now grab a pair of leggings which cover the abdomen, upper legs and lower legs. Put on the first leggings, switch to slot view, drag the second pair of leggings to the abdomen. Wear both pairs at the same time. Also works with other armor.

### Desired functionality.

Do not allow me to wear two pairs of pants. Or other items.

PR #1278, opened 2019-01-15
Fixing armor coverage in slots inventory

gmriggs:
This fixes https://github.com/ACEmulator/ACE/issues/1272

Also fixes the malformed GameEventInventoryServerSaveFailed packet, as reported by @Mag-nus.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -479,7 +479,7 @@ namespace ACE.Server.WorldObjects
             if (itemRootOwner != this && containerRootOwner == this && !HasEnoughBurdenToAddToInventory(item))
             {
                 Session.Network.EnqueueSend(new GameEventCommunicationTransientString(Session, "You are too encumbered to carry that!"));
-                Session.Network.EnqueueSend(new GameEventInventoryServerSaveFailed(Session, itemGuid, WeenieError.YouAreTooEncumbered));
+                Session.Network.EnqueueSend(new GameEventInventoryServerSaveFailed(Session, itemGuid));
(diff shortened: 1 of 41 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -466,43 +466,43 @@ public void HandleActionPutItemInContainer(uint itemGuid, uint containerGuid, in
...
                 log.WarnFormat("Player 0x{0:X8}:{1} tried to move item 0x{2:X8}:{3}.", Guid.Full, Name, item.Guid.Full, item.Name);
                 Session.Network.EnqueueSend(new GameEventCommunicationTransientString(Session, "You can't move that!")); // Custom error message
-                Session.Network.EnqueueSend(new GameEventInventoryServerSaveFailed(Session));
+                Session.Network.EnqueueSend(new GameEventInventoryServerSaveFailed(Session, itemGuid));
                 return;
             }
 
             if (itemRootOwner != this && containerRootOwner == this && !HasEnoughBurdenToAddToInventory(item))
             {
                 Session.Network.EnqueueSend(new GameEventCommunicationTransientString(Session, "You are too encumbered to carry that!"));
```
Mag-nus: Lines where we also send the string should no longer be needed.

I was adding them because they weren't popping up before. I don't know why. Now we do, our GameEvent...Failed was malformed. With the message now fixed, we can remove the duplicate GameEventCommunicationTransientString messages where applicable.

Mag-nus: For futher clarification, i mean lines like "You are too encumbered to carry that!" and similar.

PR #1288, opened 2019-01-16
Refactor ticks

Mag-nus:
This has several major changes:

Long ago, all ticking and heartbeats were done via ActionQueues. We would enqueue this work using the DelayManager. We migrated away from that system to the system we currently have because for some, the delayed heartbeats via actionqueues was a difficult concept to understand.

The problem we now realize is that the o(n) of the new ticking system adds considerable overhead, as pointed out by gmriggs.

Another thing we did a while back was move an ActionQueue onto each WorldObject. This was to guarantee thread safety for a given world object. No matter where the WorldObject went, the work traveled with it.

Now, there are 3 main actors:
- WorldManager, for a thread safe way to do work
- Landblocks, for a way we can (in the future) do work in parallel across multiple LandblockGroups
- Players, a thread safe way to enqueue work onto a player that travels with the player

Now, when we say "travels with the player", what we mean is, the work stays attached to the player and is executed on whatever LandblockGroup thread the player currently resides.

When work is enqueued onto a WorldObject, the WorldObject relays that work to its local Landblock.

This is a very safe way of doing things and still allows us the flexibility of LandblockGroups because we can make the following assumptions:
- Players can transport across the world (and threads)
- Only player transported objects can transport across the world (and threads)
- Objects on a landblock can communicate with other objects in range (and even across landblock borders) because they will be part of the same LandblockGroup, and thus the same thread.

Reminder, LandblockGroups are a concept that we haven't developed yet, but are putting the framework in place for.

Additional changes in this PR add some complexity in favor of performance to Landblock. Now we cache additions/removals to allow us to iterate over the main object list without having to clone it first.


kdmxp256:
Appears to work fine for me, including the next PR that builds on this one.

PR #1295, opened 2019-01-16
World DB Changes

LtRipley36706:
- Remove Duplicated Spell Changes


Source/ACE.Server/Entity/SpellProperties.cs
```diff
@@ -100,8 +100,7 @@ partial class Spell
         /// The amount of time the spell lasts
         /// usually for EnchantmentSpell / FellowshipEnchantmentSpells
         /// </summary>
-        //public double Duration { get => _spellBase.Duration; }
-        public double Duration { get => _spell.Duration ?? 0; } // special handling for some spells? (void)
+        public double Duration { get => _spellBase.Duration; }// special handling for some spells? (void)
```
gmriggs: this data will need to be verified and playtested for void dots

i think the durations for all of the void DoTs was 0 in the dat, and in the server db we set the duration to the correct #

Mag-nus:
if you're making world db changes, maybe look at WeeniePropertiesEmoteAction.Display and converting that to a bool

PR #1303, opened 2019-01-17
Fixing Oswald's Dirk Quest

gmriggs:
Notes during playthrough:

http://asheron.wikia.com/wiki/Oswald%27s_Dirk_Quest

- On the very first step of the quest, when picking up the Empty Mug from the table, I noticed that I had to jump up onto the table to actually get within pickup range. I updated the Player_Use distance check function to use the same cylinder distance check from physics, but even this seemed to produce the exact same result.

- In the 2nd trial with the Rolling Balls of Death, the balls weren't spawning, and I checked Gaerlan's Citadel - they were also broken again there too. They were broken from the recent change to EmoteType.CastSpellInstant to call WorldObject_Magic.TryCastSpell, which was only handling targetted spells. Updated these code paths to also handle untargeted spells.

Note, when comparing to retail video, the balls do not seem to be moving at diagonal angles - I remember having to manually set them to roll straight forward in Gaerlan's Citadel (using the angle o the invisible NPC that spawns them), so maybe there is some logic missing here to get the correct angle. I will be looking more into this.

- In the 3rd trial room, the target wasn't responding to criticals. EmoteCategory.ReceiveCritical was added, along with all of the appropriate hooks.

Also in the 3rd room, there are a bunch of invisible mobs that should be launching arrows at the player, according to this video:

https://youtu.be/C1OzPYKHqOg?t=524

https://github.com/ACEmulator/ACE-World-16PY/blob/2176c4566725c33ce691a76cf69954fa891cf0e3/Database/3-Core/9%20WeenieDefaults/SQL/Creature/Human/24139%20Invisible%20Assailant.sql

I am not seeing bow/arrows or any create_lists for these Invisible Assailants, so this is another data issue that will need to be investigated more...

- In the reward chest room, Oswald's Chest seemed to spawn 2 daggers. Also, the chest to the right, after lockpicking it, appeared to be empty (although the 2 chests behind the door had items)...

### DIFF
--- a/Source/ACE.Server/Managers/EmoteManager.cs
+++ b/Source/ACE.Server/Managers/EmoteManager.cs
@@ -57,7 +57,7 @@ namespace ACE.Server.Managers
             var emoteType = (EmoteType)emote.Type;
 
             //if (Debug)
-            //Console.WriteLine($"{WorldObject.Name}.ExecuteEmote({emoteType})");
+                //Console.WriteLine($"{WorldObject.Name}.ExecuteEmote({emoteType})");
 
             var text = emote.Message;
 
@@ -89,8 +89,8 @@ namespace ACE.Server.Managers
                 case EmoteType.AddContract:
 
                     //if (player != null)
-                    //Contracts werent in emote table
-                    //player.AddContract(emote.Stat);
+                        //Contracts werent in emote table
+                        //player.AddContract(emote.Stat);
                     break;
 
                 case EmoteType.AdminSpam:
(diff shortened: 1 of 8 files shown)


gmriggs:
Updates:

- The Invisible Assailants are indeed spawning, with their correct ammo (in property did - treasure type)

Some of the key flags here - PropertyBool.Visibility=true is actually more like an 'admin visibility-only' flag, which prevents them from even being broadcast / sent to the client

PropertyBool.Attackable is set to false for these mobs, which the current ACE code was interpreting as not being a monster that aggros. This logic has been adjusted so that Attackable + TargetingTactic is taken into account.

The Invisible Assailants now appear to be attacking the player correctly (although I did notice the arrow hit ground / hit player) sounds might be missing

- For the Rolling Balls of Death, they were indeed rolling at the correct diagonal angles to match the retail video. The issue here is that they were colliding with each other, and just stopping in their tracks. I noticed in the retail video they are clipping through each other, so I set the Ethereal flag for them (as per some other war spells) and it appears to work

PR #1333, opened 2019-01-21
Fix TimeToRot Usage

LtRipley36706:


Source/ACE.Server/WorldObjects/Container.cs
```diff
@@ -360,6 +362,8 @@ public bool TryRemoveFromInventory(ObjectGuid objectGuid, out WorldObject item)
                 item.ContainerId = null;
                 item.PlacementPosition = null;
 
+                item.TimeToRot = DatabaseManager.World.GetCachedWeenie(item.WeenieClassId).GetProperty(ACE.Entity.Enum.Properties.PropertyFloat.TimeToRot) ?? DefaultTimeToRot.TotalSeconds;
```
Mag-nus: This should really happen when the item is placed on the landblock.

There are a lot of intra-inventory actions that uses TryRemoveFromInventory. We dont' want to query the db cache for all of those. It's cached yes, but it's still a concurrent dictionary lookup.

We also don't want to do excessive work in the inventory code that only applies to landblock interactions when the amount of intra-inventory vs landblock movements might be 10,000:1 (just guessing)

The proper way to retain the original (default) TimeToRot would be to cache the property in the dictionaries using a property > 9000.

Mag-nus:
Can you separate the non TTR changes from this PR so it's easier to review for the problem it's attempting to address?

PR #1381, opened 2019-01-26
Change Corpse Rights ID storage properties

LtRipley36706:
Switch Corpse Rights tracking from AllowedActivator + OwnerId to KillerId +VictimId to make better and more clear design usage

### DIFF
--- a/Source/ACE.Server/WorldObjects/Creature_Death.cs
+++ b/Source/ACE.Server/WorldObjects/Creature_Death.cs
@@ -77,7 +77,7 @@ namespace ACE.Server.WorldObjects
             UpdateVital(Health, 0);
 
             if (topDamager != null)
-                Killer = topDamager.Guid.Full;
+                KillerId = topDamager.Guid.Full;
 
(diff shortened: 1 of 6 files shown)


Source/ACE.Server/WorldObjects/Creature_Death.cs
```diff
@@ -200,19 +200,17 @@ protected void CreateCorpse(WorldObject killer)
 
             //corpse.Location.PositionZ = corpse.Location.PositionZ - .5f; // Adding BaseDescriptionFlags |= ObjectDescriptionFlag.Corpse to Corpse objects made them immune to gravity.. this seems to fix floating corpse...
 
-            corpse.OwnerId = Guid.Full;
+            corpse.VictimId = Guid.Full;
             corpse.Name = $"Corpse of {Name}";
 
-            LandblockManager.AddObject(corpse);
```
LtRipley36706: To be clear here, this is a duplicate AddObject line.. there is another at the end of the func and should be the one left

PR #1383, opened 2019-01-27
WIP: Improve DelayManager

Mag-nus:
Reference time is now Time.GetUnixTime(), not Timers.PortalYearTicks

2 additional delay types added:
DelayUntil(double unixTimeInSeconds)
AddDelayForOneTick()

Vendor.ActOnUse has been updated to use the new methods.


Source/ACE.Server/WorldObjects/Vendor.cs
```diff
@@ -72,12 +73,12 @@ public override void ActOnUse(WorldObject wo)
 
             var rotateTime = Rotate(player);    // vendor rotates towards player
 
-            // TODO: remove this when DelayManager is not forward propagating current tick time
+            var delayUntilTime = Time.GetFutureUnixTime(rotateTime);
 
             var actionChain = new ActionChain();
-            actionChain.AddDelaySeconds(0.001f);  // force to run after rotate.EnqueueBroadcastAction
```
gmriggs: This doesn't fix the core problems with ActionChain

var firstChain = new ActionChian();
firstChain.AddAction(this, SomeAction);
firstChain.EnqueueChain();

var secondChain = new ActionChain();
secondChain.AddAction(this, SomeOtherAction);
secondChain.EnqueueChain();

Why is SomeOtherAction running before SomeAction in that case? The actions should be processed in the order they are added

AddDelayForOneTick is not something that should be needed anywhere

Please see https://github.com/gmriggs/ACE/blob/60fa9c0d9fafba775f127183be18dc556f174d70/Source/ACE.Server/Entity/Actions/ActionChain.cs for how ActionChains should work

Source/ACE.Server/WorldObjects/Vendor.cs
```diff
@@ -72,12 +73,12 @@ public override void ActOnUse(WorldObject wo)
...
+            actionChain.AddDelayForOneTick();  // force to run after rotate.EnqueueBroadcastAction
             actionChain.AddAction(this, LoadInventory);
-            actionChain.AddDelaySeconds(rotateTime);
+            actionChain.AddDelayUntil(delayUntilTime);
```
gmriggs: AddDelayUntil is also a function that is not needed, unless we just do a mass search/replace for every call to 'AddDelaySeconds' and change everything to 'AddDelayUntil'

The game server does not need to build action chains with time based on the amount of time the previous actions in the chain took to execute. In fact, this could be considered an unsafe design. If code wants to be designed this way (and I can't think of 1 place in the code that does), it should be chained for the first function to build the next action

Source/ACE.Server/Entity/Actions/DelayManager.cs
```diff
@@ -3,6 +3,8 @@
 
 using log4net;
 
+using ACE.Common;
+
 namespace ACE.Server.Entity.Actions
 {
     public class DelayManager : IActor
```
gmriggs: Having a global 'DelayManager' is also an entirely unneeded construct

There is also a core bug here that has not been addressed, where DelayManager is always 1 tick behind. This causes subtle timing bugs all throughout the code.

gmriggs: For multithreading, having a global delay manager will also be unneeded and hindersome there

The delays should be managed by the action queues

Source/ACE.Server/WorldObjects/Vendor.cs
Mag-nus: That is incorrect.

Action chains enqueued on the same IActor are run in the order in which they are enqueued.

Source/ACE.Server/Entity/Actions/DelayManager.cs
Mag-nus: Global DelayManager is a requirement for thread safety.

Even with the thread safety requirement, it's still a convenient and easy thing to have. The global delay manager isn't doing any heavy lifting. It simply manages delayed actions. Once the delay is expired, the NextAct action is enqueued onto the IActor, and the action is run on the IActors thread. Currently Landblocks are single threaded, but, in the future when Landblocks are split up into groups, the actual action will be run on the appropriate landblock group thread.

I'll investigate the tick behind.

Mag-nus: I disagree with reasons described in the comment above.

PR #1404, opened 2019-02-03
Configuration file change from JSON to JavaScript

fartwhif:
For the ability to embed comments, 
Backwards compatible, 
ACE program startup sequence and ACE.Server project build event renames Config.json to Config.js when it can.

Changed startup batch files to inform user of mistake and pause/exit when they start run_server.bat from the wrong location, and not alongside the ACE.Server.dll


Source/ACE.Common/ConfigManager.cs
```diff
@@ -19,13 +19,29 @@ public static void Initialize(MasterConfiguration configuration)
...
-        public static void Initialize(string path = @"Config.json")
+        public static void Initialize(string path = @"config.js")
         {
+            string fpOld = Path.Combine(Environment.CurrentDirectory, path);
+            string fpNew = Path.Combine(Environment.CurrentDirectory, Path.GetFileNameWithoutExtension(path) + ".js");
+            string fpChoice = null;
             try
             {
-                Config = JsonConvert.DeserializeObject<MasterConfiguration>(new JsMinifier().Minify(File.ReadAllText(path)));
+                if (!File.Exists(fpNew) && File.Exists(fpOld))
```
gmriggs: This should be unneeded. The code for this was posted yesterday:

if not exist "$(ProjectDir)Config.js" if exist "$(ProjectDir)Config.json" move "$(ProjectDir)Config.json" "$(ProjectDir)Config.js"

fartwhif: there are situations where this is needed

gmriggs: It still seems like this shouldn't be needed in the code to me, but since it can catch some possible edge cases at runtime, and shouldn't affect anything besides those special cases, I guess it should be fine.

Just trying to keep the core code as clean as possible, and hopefully after everyone has the new config filename, this code can eventually be removed

fartwhif: That's reasonable.  Let's plan on removing the runtime patcher in say, 6 months.

PR #1416, opened 2019-02-05
Change LootGenerationFactory to support TOD

kdmxp256:
- Update LootGenerationFactory to support the updated TOD weapon weenies


kdmxp256:
In the case of melee weapons, UiEffects property should not be defaulted to 0, as there are separate weenies for each loot weapon type for each element that set the default.  Only set UiEffects to 1, if the item contains spells, thus overriding the weenie default.  Missile weapons and casters have the element externally applied, so a default should be set to zero, in these two cases.

PR #1436, opened 2019-02-11
Load Landblock resources async again

Mag-nus:
We must wait for Landblock.CreateWorldObjects() to complete before we can teleport in. For that we have a new bool CreateWorldObjectsCompleted

Players now check this bool in OnTeleportComplete() and delay materliazation until the landblock resources have finished loading. This will keep players in a pink bubble state until it's safe to spawn in.

### DIFF
--- a/Source/ACE.Server/Entity/Landblock.cs
+++ b/Source/ACE.Server/Entity/Landblock.cs
@@ -52,6 +52,12 @@ namespace ACE.Server.Entity
         /// </summary>
         public bool Permaload = false;
 
+        /// <summary>
+        /// This must be true before a player enters a landblock.
+        /// This prevents a player from possibly pasing through a door that hasn't spawned in yet, and other scenarios.
+        /// </summary>
+        public bool CreateWorldObjectsCompleted { get; private set; }
+
         private DateTime lastActiveTime;
 
         private readonly Dictionary<ObjectGuid, WorldObject> worldObjects = new Dictionary<ObjectGuid, WorldObject>();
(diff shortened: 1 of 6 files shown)


Source/ACE.Server/Managers/LandblockManager.cs
```diff
@@ -153,7 +153,7 @@ public static Landblock GetLandblock(LandblockId landblockId, bool loadAdjacents
             {
                 var adjacents = GetAdjacentIDs(landblock);
                 foreach (var adjacent in adjacents)
-                    GetLandblock(adjacent, false, false, false);
+                    GetLandblock(adjacent, false);
```
gmriggs: This re-introduces the invisible player bug

The landblock loading needs a callback for when CreateWorldObjects() as completed

Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -155,28 +154,33 @@ private void CreateWorldObjects()
             var shardObjects = DatabaseManager.Shard.GetStaticObjectsByLandblock(Id.Landblock);
             var factoryObjects = WorldObjectFactory.CreateNewWorldObjects(objects, shardObjects);
 
-            // for mansion linking
-            var houses = new List<House>();
-
-            foreach (var fo in factoryObjects)
+            actionQueue.EnqueueAction(new ActionEventDelegate(() =>
```
gmriggs: This breaks RootHouse, and anything that depends on the landblock synchronous loading functionality

With this line, CreateWorldObjects() returns immediately even when sync=true, effectively breaking anything that depends on the the objects to be loaded in before returning

Mag-nus: Is this resolved by the order patch?

PR #1454, opened 2019-02-19
Fix bugs in showstats and RemoveItemSpell

kdmxp256:
- Add three null checks in EnchantmentManager.cs
- Add null check to showstats, in DeveloperCommands.cs
- Correct the enchantments sent for removal from items, in Creature_Magic.cs, to those of on items


Source/ACE.Server/Managers/EnchantmentManager.cs
```diff
@@ -434,6 +438,9 @@ public virtual void Dispel(BiotaPropertiesEnchantmentRegistry entry)
         /// </summary>
         public virtual void Dispel(List<BiotaPropertiesEnchantmentRegistry> entries)
         {
+            if (entries?.Any() != true)
```
Mag-nus: Probably better off doing a .Count here since the argument is a List<T>

PR #1462, opened 2019-02-23
ensuring monsters can damage target

gmriggs:


Source/ACE.Server/WorldObjects/Creature_Combat.cs
```diff
@@ -916,5 +916,22 @@ public virtual AttackType GetAttackType(WorldObject weapon, CombatManeuver comba
         {
             return combatManeuver != null ? combatManeuver.AttackType : AttackType.Undef;
         }
+
+        public virtual bool CanDamage(Creature target)
+        {
+            if (target is Player targetPlayer)
```
Mag-nus: You don't need the var targetPlayer here.

gmriggs: It's preparing for further code to be added to this function

PR #1468, opened 2019-02-26
Update account authentication to use BCrypt.

zegegerslittlesis:
This updates/simplifies account authentication to use the BCrypt hashing algorithm in place of SHA512 and the need for a separate password salt.

Not only is BCrypt more secure than SHA512 since it uses a "Work Factor" to discourage password brute force attacks but the implementation is much cleaner overall since the salt is randomly generated and added as part of the password hash itself, effectively protecting against rainbow table attacks.

The default Work Factor is set to 10 which is considered the current industry standard. A server owner can easily increase this value to whatever level of security they desire which in turn will make password verification slower and much harder to crack if ever brute forced by a hacker.

Furthermore, BCrypt is used in PHP's passwordHash() function so this change will open up many more account management possibilities through PHP registration scripts, especially CMS's and database bridging.


Mag-nus:
What happens if we start with WorkFactory X, but at a later date switch to WorkFactor Y?

zegegerslittlesis:
> What happens if we start with WorkFactory X, but at a later date switch to WorkFactor Y?

The higher the number, the more iterations it takes to verify a password is all. The original password hash will still be the same even though new password hashes from that point forward will be marked with a higher Work Factor number.

Lets say we start at the default of 10. All new password hashes will start with a $2y$10.

Later you switch to a Work Factor of 12. All new password hashes will now look like $2y$12. The older passwords with a factor of 10 will still work without an issue.

Mag-nus:
Ok great.

One thing we'll need to look into is how this overhead affects/interrupts the server. What thread(s) is/are slowed down when the bcrypt work is done? Should we start out with a lower work factor and improve it once we solve some of the other vectors (like plain text passwords sent over the wire).

My next question is, does this handle migrations at all? If a server operator with accounts already established applies this patch, will those accounts still work?

PS. Tell your big bro we miss him in discord.

kaos00723:
BCrypt is a well known and used hashing algorithm used worldwide. It causes absolutely no overhead with the server.

The factor of 10 is literally the default which is used industry wide and verifies a password almost immediately. Raising the factor simply causes the password verify itself to take slightly longer to keep brute force attackers at bay.

Honestly, there will pretty much never be a reason to change the workfactor from 10.

ACE is also still in development so there are no "established" server accounts which makes this a good time to add now. People with existing accounts will either have to create a new account or create their own password reset form. Again, this is hardly an issue at the moment since nobody has a permanent account on an ACE server.

I've tested this extensively and it works great.

zegegerslittlesis:
BCrypt does not cause any form of overhead with any other part of the server. It's literally one of the most common hashing algorithm's used around the world. There are no threads that are slowed down.

The default WorkFactor setting of 10 is the current industry standard and honestly there won't ever be a reason to change it from that for an AC server. It verifies passwords very quickly at this setting while still offering more security than SHA512. Increasing the WorkFactor slows down the amount of time it takes to verify a password which is not overhead. That's a security feature of BCrypt to keep ever faster computers from being able to brute force passwords easier with each new generation. Again, there won't be a reason to switch it from10 for an AC emulator but feel free to test it with different WorkFactor values yourself.

Also, I don't believe there are any established server operators for ACE yet so this would be a great to time get this implemented before there are. If there are though, accounts prior to this change will not work without a password reset form which can be created in PHP easily enough.

I've done a good amount of testing with this and it works great. This is also an overall cleaner and simpler implementation than the mess that's there now as you can see by looking at the code. It doesn't need any overhead tests because it doesn't cause any. It's just a simple SHA512 replacement for password hashing and verifying those hashes, nothing more.

Mag-nus:
We have no issues moving to BCrypt. It makes sense.

We do have some established test servers though, so we'll need to come up with a migration solution. Is that something you're interested in adding?

PR #1471, opened 2019-02-26
Add created IP address and time data for accounts.

zegegerslittlesis:
This adds the clients IP address and creation time to the Auth DB account table.


Mag-nus:
This could be useful to help validate users that have forgot their password.

How do you feel about storing the IP address in a VARBINARY(16)? IPV4 is 32 bits and IPV6 is 128 bits.

For a human readable address, you'd have to add the following functions: INET6_ATON() and INET6_NTOA()

Mag-nus:
A few more notes that were raised in chat.

If we're going to add the initial connection IP, we should probably also add:
LastLoggedInIp
LastLoggedInTime
TotalTimesLoggedIn

Mag-nus:
This is superseded by a #1985

Issue #1459, opened 2019-02-21
Configurable Chat Message On Login


cyberkiller6:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Desired functionality.

The configurable 'Welcome' variable in config.js currently allows you to set a custom message in the welcome pop-up message that is displayed the first time a new character logs into the world. 

I propose a new configurable variable in config.js that would allow the server operator to set a message to displayed in the chat box for existing users when they login to the world. This would replace or append the current hard coded message of "Welcome to Asheron's Call powered by ACEmulator For more information on commands supported by this server, type @acehelp"

The use case for this would allow a server operator to set varying messages such as server version, scheduled downtime notices/news and any other pertinent information they would want existing characters to see in the chat box upon first logging into the world.

Another option would be to use the current 'Welcome' variable, but to change it's functionality from displaying the custom message in the welcome pop-up message to the chat box so that existing users would also see it.

Thanks for your consideration

gmriggs:
Just for clarification:

These settings are not stored in Config.js, the runtime config default properties are stored in PropertyManager.cs

This functionality already exists in motd_string:

/modifystring motd_string hello

gmriggs:
functionality extended in https://github.com/ACEmulator/ACE/pull/1485

PR #1485, opened 2019-03-02
Cleaning up welcome message / server motd

gmriggs:
This addresses https://github.com/ACEmulator/ACE/issues/1459

This normalizes some of the config settings for welcome message / server motd

Note that 'Welcome' stored in Config.js is now deprecated, as Config.js should only be used to store config settings that are required to start the server

The declaration and default values for runtime config properties are shown in PropertyManager.cs. If the server operator would like to change the welcome_msg or server_motd, the following commands should be used:

/modifystring server_motd "Welcome to the custom server!\nHave fun"

### DIFF
--- a/Source/ACE.Server/Managers/PropertyManager.cs
+++ b/Source/ACE.Server/Managers/PropertyManager.cs
@@ -439,7 +439,6 @@ namespace ACE.Server.Managers
 
         public static readonly ReadOnlyDictionary<string, bool> DefaultBooleanProperties =
             DictOf(
-                ("always_show_welcome", false),
                 ("house_purchase_requirements", true),
                 ("house_rent_enabled", true),
                 ("pk_server", false)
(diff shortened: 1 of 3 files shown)


Source/ACE.Server/Managers/PropertyManager.cs
```diff
@@ -462,7 +462,8 @@ public static void LoadDefaultProperties()
 
         public static readonly ReadOnlyDictionary<string, string> DefaultStringProperties =
             DictOf(
-                ("motd_string", "Welcome to Asheron's Call" + "\n" + "  powered by ACEmulator  " + " \n" + "\n" + "For more information on commands supported by this server, type @acehelp" + "\n")
+                ("welcome_msg", "Welcome to Asheron's Call!\nTo begin your training, speak to the Society Greeter. Walk up to the Society Greeter using the 'W' key, then double-click on her to initiate a conversation."),
+                ("server_motd", "Welcome to Asheron's Call\n  powered by ACEmulator  \n\nFor more information on commands supported by this server, type @acehelp\n")
```
LtRipley36706: i think server_motd should be separate from initial banner text
```
Welcome to Asheron's Call
  powered by ACEmulator  

For more information on commands supported by this server, type @acehelp
```
followed by server_motd if there is one

gmriggs:
Since having a 2nd server motd type that appears in a popup window wasn't in retail, that would be an additional feature that would be outside of the scope of this PR.

Reducing the feature set to have a hardcoded server motd would also be an additional feature from what has already been in ACE for the better part of year already. What is already there supports both scenarios: if you want to keep the default motd prefix, the option is there for those who wish to do that. And for those who want to customize their motd, the option is also there for that. That is the main purpose of the config options: to provide the features for however people want to configure their server.

PR #1493, opened 2019-03-05
Reuse available Dynamic GUIDs

Mag-nus:
On ACE startup, the shard database is queried for available id's in the dynamic guid range.

This helps further reduce (max + 1) id allocation.

It's a bit premature, but, based on current architecture, I don't know if we'll ever need to do a id defragment.

A requirement of this method is that ace_world may only contain guids in the static range.


Mag-nus:
This PR needs some testing.

I also need to add the SQL alter statements to ace_world to restrict guids to the static ranges.

gmriggs:
Playtested on mariadb - everything still seemed to be speedy, no noticeable errors reported

PR #1499, opened 2019-03-07
Dont persist missiles to shard

Mag-nus:
Also adds logging to detect other missiles that are getting around the WeenieType check.


Mag-nus:
Note: This only applies to missiles that exist on the landblock, not missiles that exist in containers.

PR #1510, opened 2019-03-10
Rework Name and IsPlussed

LtRipley36706:


Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -823,10 +823,44 @@ public void SetPhysicsPropertyState(PropertyBool property, PhysicsState state, b
...
         // ========================================
         // used in SerializeCreateObject()
-        public virtual string Name
+        public string Name
         {
-            get => GetProperty(PropertyString.Name);
-            set => SetProperty(PropertyString.Name, value);
+            get
+            {
+                if (this is Player)
```
Mag-nus: If we do use this pattern, it's still better to be done on an override in the Player class.

However, my goal was to completely avoid any differences in how the Name property is get/set from any other property. The goal was that no matter what, in all cases, it's a 1:1 mapping of what's stored in the dictionaries.

The plan was to inject the + in places where the name is actually displayed and the + is required. I never got that far though so I don't even know if that's easy or difficult.

PR #1511, opened 2019-03-10
network stability

fartwhif:
- fixed bug causing disconnects due to NAK requests being ignored.
- fixed bug causing session to enter an unspecified state after connect request packet sent to the client is corrupted in transit
- fixed bug during handshake causing defunct session to linger
- fixed bug whenever a bad handshake occurs causing crash
- added asynchronous verification of encrypted CRCs
- added checksum caching to ClientPacket
- added handling of trusted packet with ClientSentNetErrorDisconnect flag
- added parsing of optional "flow" header data
- added more network logging
- removed "generational ISAAC" debugging tools


Source/ACE.Server/Managers/WorldManager.cs
```diff
@@ -93,6 +93,7 @@ public static void ProcessPacket(ClientPacket packet, IPEndPoint endPoint, IPEnd
             {
                 if (packet.Header.Flags.HasFlag(PacketHeaderFlags.ConnectResponse))
                 {
+                    log.Debug($"{endPoint} {packet}");
```
Mag-nus: should this be using the packetLog?

Source/ACE.Server/Managers/WorldManager.cs
```diff
@@ -138,6 +139,7 @@ public static void ProcessPacket(ClientPacket packet, IPEndPoint endPoint, IPEnd
             }
             else if (packet.Header.HasFlag(PacketHeaderFlags.LoginRequest))
             {
+                log.Debug($"{endPoint} {packet}");
```
Mag-nus: Same here, should this be using the packetLog?

PR #1512, opened 2019-03-10
Cleanup UseWithTarget and fix GameMessageInventoryRemoveObject

Mag-nus:
GameMessageInventoryRemoveObject was completely malformed. Not even sure how it was working before, if it was working at all.

The combination of the mana stone fixes and the malformed packet now fix mana management plugins like Mag-Tools and vtank.

Fixing the malformed packet may also fix other decal related issues.


Source/ACE.Server/WorldObjects/Player_Use.cs
```diff
@@ -18,18 +24,23 @@ public void HandleActionUseWithTarget(uint sourceObjectGuid, uint targetObjectGu
...
-            var invTarget = GetInventoryItem(targetObjectGuid);
-            var invWielded = GetEquippedItem(targetObjectGuid);
+            // source item is always in our possession
+            var sourceItem = FindObject(sourceObjectGuid, SearchLocations.MyInventory | SearchLocations.MyEquippedItems, out _, out _, out var sourceItemIsEquipped);
 
-            if (invSource == null)
+            if (sourceItem == null)
             {
-                // is this caster with a built-in spell?
-                var caster = GetEquippedItem(sourceObjectGuid);
```
gmriggs: This should be caster / GetEquippedItem here, else players could send this packet to cast spells from items they don't have equipped

PR #1516, opened 2019-03-12
Fix healing

kdmxp256:
- Rework and simplify slightly Player.HandleActionUseWithTarget to allow healing to work, again, and add an allowed target check into ManaStone.HandleActionUseOnTarget()

### DIFF
--- a/Source/ACE.Server/WorldObjects/ManaStone.cs
+++ b/Source/ACE.Server/WorldObjects/ManaStone.cs
@@ -58,7 +58,21 @@ namespace ACE.Server.WorldObjects
 
         public void HandleActionUseOnTarget(Player player, WorldObject target)
         {
+            WorldObject invTarget;
+
             var useResult = WeenieError.None;
+
+            if (player != target)
+            {
+                invTarget = player.FindObject(target.Guid.Full, Player.SearchLocations.MyInventory | Player.SearchLocations.MyEquippedItems);
+                if (invTarget == null)
+                {
+                    player.Session.Network.EnqueueSend(new GameEventUseDone(player.Session, WeenieError.ActionCancelled));
+                    return;
+                }
+                else
+                    target = invTarget;
+            }
 
             if (!ItemCurMana.HasValue)
             {
@@ -74,7 +88,7 @@ namespace ACE.Server.WorldObjects
                         if (!player.TryConsumeFromInventoryWithNetworking(target))
                         {
                             log.Error($"Failed to remove {target.Name} from player inventory.");
-                            return;
+                            player.Session.Network.EnqueueSend(new GameEventUseDone(player.Session, WeenieError.ActionCancelled));
                         }
 
                         //The Mana Stone drains 5,253 points of mana from the Wand.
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/ManaStone.cs
```diff
@@ -58,8 +58,23 @@ public void SetUiEffect(Player player, UiEffects effect)
...
+            if (player != target)
+            {
+                invTarget = player.FindObject(target.Guid.Full, Player.SearchLocations.MyInventory | Player.SearchLocations.MyEquippedItems);
+                if (invTarget == null)
+                {
+                    // Haven't looked to see if an error was sent for this case; however, this one fits
+                    player.Session.Network.EnqueueSend(new GameEventUseDone(player.Session, WeenieError.YouDoNotOwnThatItem));
+                    return;
+                }
+                else
```
gmriggs: this else is unneeded

Source/ACE.Server/WorldObjects/Player_Use.cs
```diff
@@ -97,24 +69,26 @@ public void HandleActionUseWithTarget(uint sourceObjectGuid, uint targetObjectGu
...
-            else if (sourceItem.WeenieType == WeenieType.Key)
-            {
-                ((Key)sourceItem).HandleActionUseOnTarget(this, targetItem);
-            }
-            else if (sourceItem.WeenieType == WeenieType.Lockpick)
-            {
-                ((Lockpick)sourceItem).HandleActionUseOnTarget(this, targetItem);
-            }
-            else
+            switch (sourceItem.WeenieType)
```
gmriggs: this entire switch, and all of these cases, are unneeded

HandleActionUseOnTarget() should be a virtual method, with the base in WorldObject calling RecipeManager

Source/ACE.Server/WorldObjects/Player_Use.cs
```diff
@@ -59,36 +59,8 @@ public void HandleActionUseWithTarget(uint sourceObjectGuid, uint targetObjectGu
...
-                else
-                    RecipeManager.UseObjectOnTarget(this, sourceItem, targetItem);
-
-                return;
-            }
-
-
-            // Is the target on the landblock?
-            targetItem = FindObject(targetObjectGuid, SearchLocations.Landblock);
+            var targetItem = FindObject(targetObjectGuid, SearchLocations.MyInventory | SearchLocations.MyEquippedItems | SearchLocations.Landblock);
```
Mag-nus: If you add SearchLocations.Landblock here, the comment will need to be changed as the search will now include things not in the players possession

Source/ACE.Server/WorldObjects/ManaStone.cs
```diff
@@ -58,8 +58,23 @@ public void SetUiEffect(Player player, UiEffects effect)
 
         public void HandleActionUseOnTarget(Player player, WorldObject target)
         {
+            WorldObject invTarget;
```
Mag-nus: Instead of having to do another lookup to see if the item is in inventory, it would be more efficient to add another arg to this function "bool targetIsInInventory".

kdmxp256: The outer FindObject() isn't being picky about whether the target is inventory or not either.  The ManaStone branch is the only one so far that really cares about the item being inventory, so I was placing the specific check where the specific matters.

Source/ACE.Server/WorldObjects/Player_Use.cs
kdmxp256: True, but outside the scope of making sure that Healing works, again.  As you mentioned, this whole code limb, if you will, should probably have the SendUseDone() pulled back from the individual branches.

PR #1519, opened 2019-03-12
Adding Chess

gmriggs:
Thanks to Anahera for this implementation!


Source/ACE.Server/WorldObjects/GamePiece.cs
```diff
@@ -28,6 +36,8 @@ public GamePiece(Biota biota) : base(biota)
 
         private void SetEphemeralValues()
         {
+            ObjScale = 0.5f;
```
LtRipley36706: Technically data shows this to be 0.75 but i'm sure the smaller value looks better

gmriggs: still need to investigate this more... even though the pieces are ethereal, with the default ObjScale of 0.75, they were running into collision issues, not being able to move to the target square, etc. Setting the scale to 0.5 seemed to fix this, so I will investigate this more

gmriggs: ok the movement bugs for the default size have been figured out, and the chess pieces have been reverted to their default size

PR #1525, opened 2019-03-13
Fix the legacy Open/Close virtual functions in WorldObject base

Mag-nus:
I don't understand the emote system, but it seems odd to me that
CloseMe does **targetObject**.Close()
and
OpenMe does **WorldObject**.Open()

Shouldn't they be performing the action on the same object?


gmriggs:
this should be WorldObject.Open/Close

DridsGhost:
Can we choose to make it work for just containers/doors or make it more generic and add overridable Open/Close to base WorldObject that weenie classes override like Activate?

PR #1533, opened 2019-03-13
Additional checks for built-in weapon spells

gmriggs:


gmriggs:
Requesting review from @kdmxp256 and @Mag-nus, especially for the spell verification stuff

HandleActionCastTargetedSpell() and HandleActionCastUntargetedSpell() now call VerifySpell()

These 2 HandleAction functions are only called from the network code, when the player sends these 2 specific packets, so it should only be validating against rogue inputs

A concern I have is for the performance characteristics of the existing Biota.SpellIsKnown() function:

return biota.BiotaPropertiesSpellBook.FirstOrDefault(x => x.Spell == spell) != null;

BiotaPropertiesSpellBook is a standard Collection type, so I'm guessing this will not be doing any indexing, and will be o(n) instead of o(1)? Should these be cached somewhere?

Mag-nus:
You are correct that this will be an o(n) lookup.

I think this is fine to merge for now. In the future we can add one of the following and pass it to all the BiotaExtensions that interface with BiotaPropertiesSpellBook:
HashSet<int> SpellCache
HashSet<int, BiotaPropertiesSpellBook> SpellCache

We implement a very similar pattern for the property dictionaries.

PR #1542, opened 2019-03-15
plugin system

fartwhif:
Allows loading instances of foreign assemblies into the primary execution context.

[Here](https://github.com/fartwhif/ACE-Plugins) are some examples of plugins.


fartwhif:
WIP until completion of the rest of the projects that I'm currently converting into plugins: `ACE.Plugin.Transfer` and `ACE.Plugin.WebAPI`

fartwhif:
currently the biggest thing to solve is how to have a system dependent distribution for plugins, which means when the system loads a plugin dll, it knows how to compile sources and otherwise obtain the microsoft and nuget based dependencies, otherwise those binaries have to be included in with the plugin and that's `crusty`

fartwhif:
if it makes you feel any better I've been designing world class extension systems since 2005.

fartwhif:
Closed it without explanation.

PR #1543, opened 2019-03-15
Odds and Ends

LtRipley36706:
* Skip Burden in CreateObject if object is Creature
* Update PlayerDescriptionEvent for pulling correct WeenieType and HasHealth bool
* Stop sending all children equipped items

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Tracking.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Tracking.cs
@@ -63,7 +63,8 @@ namespace ACE.Server.WorldObjects
             if (worldObject is Creature creature)
             {
                 foreach (var wieldedItem in creature.EquippedObjects.Values)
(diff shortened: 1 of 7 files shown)


Source/ACE.Server/WorldObjects/Player_Tracking.cs
```diff
@@ -63,7 +63,8 @@ public void TrackObject(WorldObject worldObject)
             if (worldObject is Creature creature)
             {
                 foreach (var wieldedItem in creature.EquippedObjects.Values)
-                    TrackEquippedObject(creature, wieldedItem);
+                    if (IsInChildLocation(wieldedItem))
```
gmriggs: I don't think this will work, ammo is not a child object

LtRipley36706: see new commit 6fc802b19f8dedcc9b54ecb0675f3c536ead84e2

PR #1548, opened 2019-03-16
Four changes

kdmxp256:
- Expand the usage of CreateIOU, namely for EmoteType.Give
- Change InqAttributeStat, InqSecondaryAttributeStat, and InqSkillStat to use Current; Raw versions of these three inquiries continue to check against Base, per Morosity
- Move cooldown processing into base OnActivate and CheckUseRequirements
- Support Active equaling 0xffff on chests with spells, which should only cast their spell on opening the chest


Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -2409,12 +2409,13 @@ public string QuestRestriction
         /* pressure plates */
 
         /// <summary>
-        /// Returns TRUE if this object can be activated (default)
+        /// Returns 1 or 65535 if this object can be activated (1 is default)
+        /// 23 Weenies in Cache.bin contain Active 65535
```
gmriggs: should be 0xFFFF

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -99,9 +99,18 @@ public virtual void OnActivate(WorldObject activator)
             // if ActivationTarget is another object,
             // should this be checking the ActivationResponse of the target object?
 
-            // default use action
-            if (ActivationResponse.HasFlag(ActivationResponse.Use))
+            // Chests with 'ActivationResponse - CastSpell' should still ActOnUse
+            if (WeenieType == WeenieType.Chest && Active == 65535)
```
gmriggs: 0xFFFF

Source/ACE.Server/Factories/PlayerFactory.cs
```diff
@@ -379,14 +379,17 @@ private static WorldObject GetClothingObject(uint weenieClassId, uint palette, d
...
-        private static void CreateIOU(Player player, uint missingWeenieId)
+        public static bool CreateIOU(Player player, uint missingWeenieId, bool withNetworking = false)
         {
             var book = (Book)WorldObjectFactory.CreateNewWorldObject("parchment");
 
             book.SetProperties("IOU", "An IOU for a missing database object.", "Sorry about that chief...", "ACEmulator", "prewritten");
             book.AddPage(player.Guid.Full, "ACEmulator", "prewritten", false, $"{missingWeenieId}\n\nSorry but the database does not have a weenie for weenieClassId #{missingWeenieId} so in lieu of that here is an IOU for that item.");
 
-            player.TryAddToInventory(book);
+            if (withNetworking == false)
```
gmriggs: this should just be if (withNetworking), or if (!withNetworking) to check for false

if (withNetworking == false) is not a form commonly used

Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -372,7 +387,7 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
                     if (targetCreature != null)
                     {
                         var attr = targetCreature.Attributes[(PropertyAttribute)emote.Stat];
-                        success = attr != null && attr.Ranks >= emote.Min && attr.Ranks <= emote.Max;
+                        success = attr != null && attr.Current >= emote.Min && attr.Current <= emote.Max;
```
gmriggs: are you sure this should be Current instead of Base?

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -92,12 +93,24 @@ public virtual void OnActivate(WorldObject activator)
...
 
+            if (player != null)
+                player.EnchantmentManager.StartCooldown(this);
+
             // if ActivationTarget is another object,
             // should this be checking the ActivationResponse of the target object?
 
-            // default use action
-            if (ActivationResponse.HasFlag(ActivationResponse.Use))
+            // Chests with 'ActivationResponse - CastSpell' should still ActOnUse
```
gmriggs: this is the type of logic that shouldn't be in base

this is what the overrides are for, subtype-specific logic

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -60,21 +60,22 @@ public virtual void OnActivate(WorldObject activator)
...
+            // when players double click an object,
+            // and the packet comes in as GameAction 0x36 - UseItem
             // from the game perspective, technically this starts as an 'Activate',
             // which can have a list of possible ActivationResponses - 
             // Use (by far the most common), Animate, Talk, Emote, CastSpell, Generate
 
-            // PropertyInt.Active indicates if this object can be activated, default is true
-            if (!Active) return;
+            // PropertyInt.Active indicates if this object can be activated, default is 1
+            if (Active == 0) return;
```
gmriggs: this should be an enum, instead of having magic values (0, 1, 0xFFFF)

Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
kdmxp256: I was going to change it to a descriptive Enum

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
kdmxp256: I was going to change it to a descriptive Enum

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
kdmxp256: See above

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
kdmxp256: Except when the code was as it was before, the chests didn't actually do anything.  Merely allowing them past the Active check, the chest would cast its spell every time the interacted with the chest in any way, as the flow would perform the ActOnUse and then the OnCastSpell would fire.  Should only be casting the spell when opening the chest.

gmriggs: the fixes for these comments have been posted as a patch in the dev channel

kdmxp256: Your suggestions for OnActivate() won't work for Chests, as the spell is only cast when a player opens the chest, not for all interactions.

Source/ACE.Entity/Enum/Active.cs
```diff
@@ -0,0 +1,9 @@
+namespace ACE.Entity.Enum
```
gmriggs: this is unneeded, as there is currently no defined logic based on the difference between 'Active' and 'SecondaryAction'

Source/ACE.Server/WorldObjects/PressurePlate.cs
```diff
@@ -72,5 +77,17 @@ public override void OnActivate(WorldObject activator)
 
             base.OnActivate(activator);
         }
+
+        public override void ActOnUse(WorldObject activator)
+        {
+            // Pressure Plate should be activating a linked object configured via landblock_link
```
gmriggs: if the data is being fixed for this, then PressurePlate.ActOnUse() should just be removed entirely, and just base.ActOnUse() for the error message

Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -2409,12 +2409,39 @@ public string QuestRestriction
         /* pressure plates */
 
         /// <summary>
-        /// Returns TRUE if this object can be activated (default)
+        /// Returns Active.Active or Active.SecondaryAction, if this object can be activated (1 is default)
+        /// 23 Weenies in Cache.bin contain Active 65535 or 0xffff
         /// </summary>
-        public bool Active
+        private Active Active
```
gmriggs: there is no need for this until there is actually logic defined based on the difference between Active=1 and Active=0xFFFF

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -60,21 +60,22 @@ public virtual void OnActivate(WorldObject activator)
...
+            // when players double click an object,
+            // and the packet comes in as GameAction 0x36 - UseItem
             // from the game perspective, technically this starts as an 'Activate',
             // which can have a list of possible ActivationResponses - 
             // Use (by far the most common), Animate, Talk, Emote, CastSpell, Generate
 
-            // PropertyInt.Active indicates if this object can be activated, default is true
-            if (!Active) return;
+            // PropertyInt.Active indicates if this object can be activated, default is 1
+            if (Active == Active.Inert) return;
```
gmriggs: see above

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -92,12 +93,27 @@ public virtual void OnActivate(WorldObject activator)
...
             // if ActivationTarget is another object,
             // should this be checking the ActivationResponse of the target object?
 
-            // default use action
-            if (ActivationResponse.HasFlag(ActivationResponse.Use))
+            // Chests with ActivationResponse.CastSpell ( and Active 65535 ) should
+            // still process ActOnUse as normal but need to stop further processing
+            // after completing, as the CastSpell portion is handled internally
+            // by the Chest() class only during the Chest().Open event
+            if (WeenieType == WeenieType.Chest && Active == Active.SecondaryAction)
```
gmriggs: this is not needed, and is definitely not related to Active=0xFFFF

the patch posted in the dev channel already had this working without any specialized logic required

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -150,12 +166,12 @@ public virtual void OnEmote(WorldObject activator)
                 EmoteManager.OnActivation(creature);
         }
 
-        public virtual void OnCastSpell(WorldObject activator)
+        public virtual void OnCastSpell(WorldObject activator, WorldObject caster = null)
```
gmriggs: caster is an unneeded parameter. the caster is always 'this'

PR #1564, opened 2019-03-22
updating burden from ammo / spell component

gmriggs:


Source/ACE.Server/WorldObjects/Creature_Missile.cs
```diff
@@ -38,7 +38,14 @@ public float ReloadMissileAmmo(ActionChain actionChain = null)
             actionChain.AddAction(this, () =>
             {
                 EnqueueActionBroadcast(p => p.TrackEquippedObject(this, ammo));
-                EnqueueBroadcast(new GameMessageParentEvent(this, ammo, (int)ACE.Entity.Enum.ParentLocation.RightHand, (int)ACE.Entity.Enum.Placement.RightHandCombat));
+
+                var delayChain = new ActionChain();
+                delayChain.AddDelaySeconds(0.001f);     // ensuring this message gets sent after player broadcasts above...
```
Mag-nus: A note on this, ACE pools up messages to send and dispatches them every 5ms.

Any messages that are enqueued via that time could end up being sent in the same payload.

I know we have situations like this that exist elsewhere, it's just odd, because due to UDP packet loss, the packets could end up coming in any order anyway.

gmriggs: @Mag-nus @LtRipley36706 

with this, it really would be better to just send the CO for the arrow when the monster spawns in, and then broadcast the ParentEvent when it is to be made visible

Issue #1524, opened 2019-03-12
Hooked items do not display all relevant information on appraisal


OptimShi:
```
- [ X ] bug report -> please search issues before submitting
```

### Repro steps.
* Inscribe a hookable item.
* Buy a house.
* Hook the item on the house in an appropriate hook.
* Appraise the hooked item.

The appraised hook item is missing information.

### Desired functionality.
* The Use String should be copied from the source item, if any.
* The Long Desc should read "This hook is owned by [HouseOwnerName]. It contains: \n[Item.LongDesc]"
* The hooked item should pass a hookProfile field in the appraisal info event if the item is inscribed (HookProfile.Flags |= HookFlags.Inscribable) Unsure if this should always be sent for a hook or not. Needs more investigation.
* The Inscription + ScribeName should be displayed, if any, in the appropriate area on the examine UI.
* Name, Burden and Value are all properly set.

OptimShi:
Addressed in PR #1569

PR #1569, opened 2019-03-25
Hook appraisals

OptimShi:
* Appraised items on housing hooks now show the details on the hooked item.

Fixes Issue #1524

### DIFF
--- a/Source/ACE.Server/Managers/HouseManager.cs
+++ b/Source/ACE.Server/Managers/HouseManager.cs
@@ -174,6 +174,7 @@ namespace ACE.Server.Managers
 
             house.HouseOwner = null;
             house.MonarchId = null;
+            house.HouseOwnerName = null;
 
             house.ClearPermissions();
 
--- a/Source/ACE.Server/Network/Structure/AppraiseInfo.cs
+++ b/Source/ACE.Server/Network/Structure/AppraiseInfo.cs
@@ -139,7 +139,6 @@ namespace ACE.Server.Network.Structure
                             var houseOwnerPlayer = PlayerManager.FindByGuid((uint)wo.ParentLink.HouseOwner);
                             if(houseOwnerPlayer != null)
                             {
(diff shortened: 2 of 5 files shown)


Source/ACE.Server/Managers/HouseManager.cs
```diff
@@ -174,6 +174,7 @@ public static async void HandleEviction(PlayerHouse playerHouse)
 
             house.HouseOwner = null;
             house.MonarchId = null;
+            house.HouseOwnerName = null;
```
gmriggs: not a huge deal, but is the overhead of storing HouseOwnerName really worth it here? is there anything to be gained, aside from "that's how retail did it"?

the house owner name is already accessible through PlayerManager.FindByGuid(Hook.House.HouseOwner).Name, and storing things statically here just leaves more room for error and possible bugs. for example, what happens if the player changes their name on the server? all references to their name should be updated to reflect that

retail stored a ton of runtime info in properties for many reasons, a lot were related to the technical implementation of the original servers. anytime anything could potentially be cross-landblock or cross-server (which i would imagine would be most things, ie. the possibility of appraising a housing hook from across a landblock boundary) retail had to store a ton of serialized info in runtime properties for inter-landblock and inter-server communication. it made sense in that situation, but for the current architecture, that same design decision might not apply

OptimShi: I figured the cost of storing one text field outweighed the cost needed to load an "OfflineCharacter" when all we needed was just to pull a name.

gmriggs: technically they are both dictionary lookups (looking up a string property vs. looking up an offline character) so it should take the same amount of processing time

the server keeps a dictionary of all the offline characters loaded on the server. its just returning a reference that has been pre-loaded. keeping fast lookup tables of all the offline characters is important for things like building the allegiance chain structures, so this process should be very efficient

again not a huge issue, i'm fine with the pr either way

PR #1576, opened 2019-03-27
Support AdvocateItem RadarBlipColor changes

LtRipley36706:
When Advocates equipped their Aegis (AdvocateItem) RadarBlipColor changed from White (Default) to Pink (Advocate) and the reverse occurred on unequip.

### DIFF
--- a/Source/ACE.Server/Managers/EmoteManager.cs
+++ b/Source/ACE.Server/Managers/EmoteManager.cs
@@ -898,8 +898,8 @@ namespace ACE.Server.Managers
                     break;
 
                 case EmoteType.SetInt64Stat:
-                    player.UpdateProperty(player, (PropertyInt64)emote.Stat, emote.Amount);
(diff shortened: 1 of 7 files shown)


Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -870,7 +870,8 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
                     break;
 
                 case EmoteType.SetBoolStat:
-                    targetObject.SetProperty((PropertyBool)emote.Stat, emote.Amount == 0 ? false : true);
+                    player.UpdateProperty(player, (PropertyBool)emote.Stat, emote.Amount == 0 ? false : true);
```
gmriggs: this can't be assumed that the targetObject is always a player here... targetObject should never be null, but it's possible the targetObject could not be a player

Issue #1294, opened 2019-01-16
Gear Knight model arrangement for player characters, at least, is not correct


ghost:
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
Create a Gear Knight character

### Mention any other details that might be useful.
![image](https://user-images.githubusercontent.com/22699181/51258308-8b86d380-195e-11e9-9279-7371a21055fb.png)

fartwhif:
Ubiisoft called they said SHIP IT

OptimShi:
Basic fix is working on local branch, so players do not literally have their butts for a head, but still doesn't display the proper body style (as it is labelled during character creation; essentially changes which textures are affected by which palette--the model shape is still the same). Will be adding a new property for players and adjusting how these "body style" player types (Gear Knights, Olthoi) are loaded with that property set.

PR #1610, opened 2019-03-31
Gear Knight display fixes

OptimShi:
This corrects the appearance of Gear Knights. (And should also apply to Olthoi characters, once enabled).

There is a work around for older Gear Knight players, as well, so they should look normal (but may not have accurate colors as chosen by player, and they never will due to missing data)

This addresses issue #1294

### DIFF
--- a/Source/ACE.Server/WorldObjects/WorldObject_Networking.cs
+++ b/Source/ACE.Server/WorldObjects/WorldObject_Networking.cs
@@ -1068,7 +1068,10 @@ namespace ACE.Server.WorldObjects
         protected void AddBaseModelData(ACE.Entity.ObjDesc objDesc)
         {
             // Hair/head
-            if (HeadObjectDID.HasValue && !HairStyle.HasValue)
+
+            // if (HeadObjectDID.HasValue && !HairStyle.HasValue)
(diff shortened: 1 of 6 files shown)


Source/ACE.Server/WorldObjects/WorldObject_Networking.cs
```diff
@@ -1068,7 +1068,10 @@ public virtual ACE.Entity.ObjDesc CalculateObjDesc()
         protected void AddBaseModelData(ACE.Entity.ObjDesc objDesc)
         {
             // Hair/head
-            if (HeadObjectDID.HasValue && !HairStyle.HasValue)
+
+            // if (HeadObjectDID.HasValue && !HairStyle.HasValue)
+            // This Heritage check has been added for backwards compatibility. It works around the butthead Gear Knights appearance.
+            if (HeadObjectDID.HasValue && !HairStyle.HasValue && Heritage.HasValue && Heritage != (int)HeritageGroup.Gearknight)
```
OptimShi: I'm not sure if this is okay, or if we should do a check much earlier on just once and essentially kill this property there (set it to null). That was my initial thinking, but I couldn't see WHERE would be a good place for a check like that.

PR #1628, opened 2019-04-02
Use ServerGarbageCollection instead of Workstation GC

Mag-nus:
This prevents all threads from being blocked while GC runs (Workstation mode).

Server allows the game threads to continue to operate while GC performs. This is also what is done in web applications and other server based applications.


Mag-nus:
Might be my imagination, but this seems to greatly improve server performance under load.

PR #1629, opened 2019-04-02
Support for ACE.Plugin.API

fartwhif:


fartwhif:
command manager changes are moved to Plugin and other changes are obsolete

PR #1632, opened 2019-04-02
Add Coelesced Aetheria to loot drops

Theran578:
- Add Coelesced Aetheria to loot drops
- Adjust LootGenerationFactory.GetValue() to take the inherent weenie value of items into account, rather than a seemingly arbitrary random value


Theran578:
Even if quest to activate the slots are not available, both Janier al-Evv that sells Aetheria Desiccant and Cypher that exchanges stacks of Aetheria powder for MMD notes or Mana Forge keys are present in Xarbydun.

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -88,12 +86,34 @@ public static List<WorldObject> CreateRandomLootObjects(TreasureDeath profile)
...
                 for (var i = 0; i < numItems; i++)
                 {
-                    lootWorldObject = CreateRandomLootObjects(profile.Tier, true, lootBias);
-                    if (lootWorldObject != null)
-                        loot.Add(lootWorldObject);
+                    // Coalesced Aetheria don't drop for tiers less than 5
+                    // According to wiki, Weapon Mana Forge chests don't drop Aetheria
+                    // Otherwise, 50% chance to drop a Coalesced Aetheria 
+                    if (aetheriaGenerated == false && profile.Tier > 4 && lootBias != LootBias.Weapons)
+                        aetheriaDropChance = ThreadSafeRandom.Next(1, 2);
```
gmriggs: Is there a particular reference for the 50% drop rate for Aetheria?

This seems like it would be an extremely high drop rate...

Theran578: I haven't found any reference to the amount that it drops.  There are statements to where it won't drop.

The 50% chance is for each roll in the magic loot category to get either an Aetheria or other magic containing item.   Only one is allowed to generate per call into the loot generation, so even on multiple rolls, only a max of one Aetheria will be generated.  While play testing this patch in multiple loot tiers, the Aetheria spawned less than the Summon Essense do.  With any less chance to generate an Aetheria, it felt too sparse.  At any rate, it is just the first stab at getting them into the mix.  If others that may have an even better sense of it say its still too high, it can be decreased.

Theran578: I did make a generalization and inference from a note in the wiki for Steel Chests, in that either Coalesced Aetheria or a weapon drops for its two non-scroll or mana charge slots.

Theran578: I am certainly open to changing it, if the drop rate still feels to high.

gmriggs: i can try playtesting when i get a chance

knee-deep in bug fixes for the live server atm...

PR #1636, opened 2019-04-03
Update SummonPortal for Gateways

LtRipley36706:
Add disable_gateway_ties_to_be_summonable configurable option.

When enabled, players will not be able to summon gateways from ties of other gateways. This is not default behavior


Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -955,7 +955,6 @@ protected bool SummonPortal(uint portalId, Position location, double portalLifet
             gateway.MinLevel = portal.MinLevel;
             gateway.MaxLevel = portal.MaxLevel;
             gateway.PortalRestrictions = portal.PortalRestrictions;
-            gateway.PortalRestrictions |= PortalBitmask.NoSummon;
```
gmriggs: any reference for this?

LtRipley36706: none, and in fact, I believe it to be an incorrect memory.. adjusted PR

Issue #718, opened 2018-04-02
ACE.Server needs to expose an API for diagnostics and other information


Mag-nus:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [ X] feature request
```

### Desired functionality.
The ACE project is growing rapidly and we need the ability to expose diagnostic information from ACE.Server for 3rd party tools to access.

Such data exposed might be:
Number of players connected
Player locations
Landblock states
Chat
Physics Information
Debugging

We used to have an ACE.Diagnostics project that collected information from ACE.Server, but that was removed because the implementation required ACE.Server to have a direct reference to the ACE.Diagnostics project.

ACE.Server should not require a direct reference to the 3rd party tools collecting the information.

Perhaps an ACE.Diagnostics library is created that is used for the API.

We need someone to help us model the base implementation so that we can build on it as the project grows.

fartwhif:
SurfaceTexture added as PR #802

gmriggs:
Possible technologies / libraries mentioned:

System.Reactive / ReactiveUI - http://reactivex.io/ (gmriggs, initial investigation seemed like Dictionary support w/ updates was not an exact fit)

LINQ expressions - https://msdn.microsoft.com/en-us/library/system.linq.expressions.expression(v=vs.110).aspx, https://jacksondunstan.com/articles/3199 (fartwhif, mentioned that he used these to build a similar API)

WCF - https://docs.microsoft.com/en-us/dotnet/framework/wcf/whats-wcf (Magnus, contract-based model)

OData Open Data Protocol - https://en.wikipedia.org/wiki/Open_Data_Protocol, https://stackoverflow.com/questions/15154469/is-there-a-odata-query-to-linq-where-expression-odata-to-linq (parad0x, looks very promising. LINQ support, and starts to add more functionality for REST WebAPI)

amoeba:
I'd find the current count of players useful for [TreeStats](https://treestats.net). Since TreeStats currently gets player counts from my plugin (which gets it from from Decal `CharacterFilter`), the information I have requires players to run the plugin and suffers staleness issues related to the server protocol only sending the player count at login (I'm not sure what ACE's behavior is here). I think I'd find a public HTTP API the most useful for getting this info. Just my 2c.

deadlyelectron:
It seems like there is 2 usages for this
- high perf/throughput data (live physics/landblock/player loc data)
[...]

PR #1642, opened 2019-04-04
/serverperformance command (Server Performance Monitor)

Mag-nus:
Optional parameters are:
start
stop
reset

If no parameters are present, the current performance metrics are spit out.

When disabled, overhead is the cost of some simple function calls and a bool check. When enabled, the additional cost is stopwatch events. In comparisson to the work that ACE does for normal processing, the work done when serverperformance is enabled is nearly 0.

Default is disabled.

Enable it to start up automatically in the config.js. This is recommend for most servers. Enable it at runtime using /serverperformance start


fartwhif:
`ACE.Plugin.Web` and `ACE.Plugin.API` will have [support for websockets](https://github.com/NancyFx/Nancy/wiki/WebSockets-in-Nancy-using-OWIN) as per the "high perf/throughput data" [requirement](https://github.com/ACEmulator/ACE/issues/718#issuecomment-399276239), and this would be an ideal use with it.  This is semi-related to [the need for this generalization of the /serverstatus admin command](https://github.com/ACEmulator/ACE/pull/1629/files) so that outside utilities can also use it.  I haven't yet had the chance to review this, but just in case I'm asking that this also be "plugin-aware" so that the output can be funneled into another process, by use of event, action, etc.  Make its consumption generalized so that it can be used outside of the normal console+in game chat method.

Mag-nus:
fartwhif, It is written in a way that will allow plugin consumption.

ServerPerformanceMonitor is a public static class.
It exposes simple functions to Start/Stop/Reset.
It exposes a public ToString()
It also exposes public ways to get specific monitors.

I don't know how the plugins interact with the system yet, but, if they can access any of the static manager classes easily, then they will be able to control and get data from this resource.

fartwhif:
Nice, that's good to hear.  Is there any way to have some kind of "streaming" mode, so that whenever the state changes that the new attribute, or the whole thing, is then sent to all event subscribers?  `OnChange?.Invoke(this);`

Mag-nus:
That woudldn't work for an object like this. The overhead would be immense.

This manager can be called tends of thousands of times a second. It's written in a very basic way to maximize performance at the expense of a little readability.

fartwhif:
I see, so a wrapper that polls it would be in order.  Was worth asking.  Thanks!

PR #1650, opened 2019-04-05
rares

dgatewood:
sounds not working, but wont have time to work on it this weekend. We can merge this and somebody else can fix sounds. or I can get it on Monday.


Source/ACE.Server/WorldObjects/Creature_Death.cs
```diff
@@ -223,10 +223,8 @@ protected void CreateCorpse(WorldObject killer)
             }
             else
                 corpse.LongDesc = $"Killed by misadventure.";
-
-            bool saveCorpse = false;
```
Mag-nus: This pr is great!

Looks like the force corpse save code was not merged properly though.

PR #1654, opened 2019-04-06
Retry failed shard db queries

Mag-nus:
On my load test server, occasionally I would see SaveBiota SSL exceptions. With this PR, it passes on the second attempt.

This may also work for many other db related exceptions.


mcreedjr:
Just a suggestion... I would think it would be good to track how many second attempts are required and report on them using the same server performance commands you developed earlier. I think it would be useful for a server admin to know how many retries are occurring as it could be indicative of a problem they may want to investigate and resolve.

Mag-nus:
Good idea.

After this PR gets merged, I'll change /databasequeueinfo to /databaseperformance to report more information as well.

PR #1657, opened 2019-04-06
Landblock Dormancy after 1min of no player activity

Mag-nus:
This is a quick pass at adding dormancy to landblocks.

Currently, the dormancy timer is set to 1m of idle. The existing unload timer is still 5m.

When a landblock is dormant
- Monsters don't tick AI
- Physics aren't handled


Mag-nus:
This needs some more feedback and play testing.

This simple PR as it is can reduce the CPU demand by 50% or more depending on the usage scenario.

PR #1662, opened 2019-04-07
Contingent: consolidated the session collections

fartwhif:


Mag-nus:
If this is for testing the CE issue, it should be merged into CE.

It shouldn't go into master. It's a much less performant way to manage sessions.

fartwhif:
Merge is contingent upon whether or not it ends up fixing the "~30 sessions dropped about the same time" issue.  If it doesn't close without merge.

fartwhif:
1687 supersedes this and was merged

fartwhif:
1726 supersedes this

PR #1671, opened 2019-04-08
Rapid fix test

LtRipley36706:


Source/ACE.Server/Physics/Common/ObjectMaint.cs
```diff
@@ -123,7 +123,7 @@ public List<PhysicsObj> AddObjects(List<PhysicsObj> objs)
             return newObjs;
         }
 
-        public static bool InitialClamp = true;
+        public static bool InitialClamp = false;
```
gmriggs: this should be kept at true for the pending pr

PR #1681, opened 2019-04-09
Code cleanup

Mag-nus:
Just a lot of little Resharper inspired optimizations.

I tried to stick to ones that would improve performance.


Source/ACE.Server/Entity/DeathItem.cs
```diff
@@ -65,7 +65,7 @@ public DeathItems(List<WorldObject> inventory)
             Inventory = Inventory.Where(i => i.Category != DeathItemCategory.None).ToList();
 
             // sort by original value and category
-            Inventory = Inventory.OrderByDescending(i => i.AdjustedValue).OrderBy(i => i.Category).ToList();
+            Inventory = Inventory.OrderByDescending(i => i.AdjustedValue).ThenBy(i => i.Category).ToList();
```
gmriggs: this is incorrect, and breaks death items

PR #1687, opened 2019-04-10
Contingent: This (should) fixe the threading issue with sessions and still be performant

Mag-nus:


Mag-nus:
I've load tested this for 24hrs. Seems solid.

PR #1690, opened 2019-04-10
Process inbound GameAction packets in order received

Mag-nus:
This does not change the performance of ACE. It just shifts the processing to a different order in UpdateWorld()

It changes from an ActionQueue per session for these types of messages to a single action queue per world.

It also changes the order in which messages are processed.

Now, they are processed in the order in which they were received. Before, they were processed session by session, starting with the sessions with lower id's first.


Mag-nus:
I'm in the process of load testing this with latest networking updates.

Mag-nus:
This passed validation testing.

PR #1692, opened 2019-04-10
GfxObjCache + WeakReference

gmriggs:


gmriggs:
has been running on CE for past week, no reported errors

PR #1713, opened 2019-04-12
Add physics cache counts to /ServerStatus

Mag-nus:


Mag-nus:
I'm not even sure if this has value. I forget how many total GfxObj's there are to cache.

PR #1716, opened 2019-04-13
hotspot / corpse fixes

gmriggs:


Source/ACE.Server/WorldObjects/WorldObject.cs
```diff
@@ -262,6 +262,8 @@ private void SetEphemeralValues()
                 Placement = ACE.Entity.Enum.Placement.Resting;
 
             //CurrentMotionState = new Motion(MotionStance.Invalid, new MotionItem(MotionCommand.Invalid));
+            if (WeenieType == WeenieType.Corpse)
```
Mag-nus: This is fine and should not affect decaying.

PR #1726, opened 2019-04-13
replace session array and lock with concurrent collection

fartwhif:
consolidated session mechanisms and collections into a single concurrent collection
heavyweight concurrency for session collection
additional try/catch for FlushPackets


Mag-nus:
I believe this PR is no longer needed.

The issue that this PR was trying to address is resolved in Master as a result of another PR, and testing on CE has confirmed stability.

PR #1731, opened 2019-04-14
adding CanAddToInventory / pre-check for Player.Give

gmriggs:


Mag-nus:
There's already a function you can use to see if a container (Player) has free inventory slots

Container.GetFreeInventorySlots(bool includeSidePacks)

The CanAddToInventory function can simply be replaced with calls to 
HasEnoughBurdenToAddToInventory(item) && GetFreeInventorySlots() > 0

PR #1745, opened 2019-04-16
Fix some issues with secure trade.

LtRipley36706:
* Can ID items in trade window properly
* Only initiator does MoveTo/Rotate now

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -458,8 +458,8 @@ namespace ACE.Server.WorldObjects
                     {
                         result = currentTradePartner.GetInventoryItem(objectGuid);
 
-                        //if (result != null)
-                        //    return result;
+                        if (result != null)
+                            return result;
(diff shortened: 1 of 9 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -449,6 +451,28 @@ public WorldObject FindObject(ObjectGuid objectGuid, SearchLocations searchLocat
...
+
+                        if (result != null)
+                            return result;
+                    }
+                }
+            }
+
+            if (searchLocations.HasFlag(SearchLocations.ObjectsKnownByMe))
+            {
+                result = GetKnownObjects().Where(o => o.Guid == objectGuid).First();
```
gmriggs: GetKnownObjects().FirstOrDefault(o => o.Guid == objectGuid)

First() will crash on null, FirstOrDefault() handles null

LtRipley36706: changed

PR #1749, opened 2019-04-17
Reduce rogue physics landblocks loaded by portal desitnation parsing

Mag-nus:
On my test of 1000 landblocks, x: 0x20-0x40, y: 0x20-0x40,

After all landblocks load usign "loadalllandblocks" and unload, there were still ~36 physics landblocks remaining.

This reduces it to ~15.

I've also commented out the landblock unload debug message.


Mag-nus:
Without this (in current master), entering town network ends up loading ~47 physics landblocks without their world landblock counterparts.

PR #1751, opened 2019-04-17
fixing landblock leak

gmriggs:


Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -219,8 +219,10 @@ private void SpawnEncounters()
 
                 if (wo == null) continue;
 
-                var xPos = Math.Clamp(encounter.CellX * 24.0f, 0.5f, 191.5f);
-                var yPos = Math.Clamp(encounter.CellY * 24.0f, 0.5f, 191.5f);
+                var radius = wo.PhysicsObj.GetRadius();
```
Mag-nus: This throws a nullref exception.

Would the PhysicsObj ever be init at this point?

gmriggs: hmm, i thought this at first too, but from testing, all of the encounters spawned correctly, without throwing null exception...

also testing patch live now, and oddly enough, it hasn't thrown a null exception yet..

looking at the logic, it looks like it should be immediately..

gmriggs: oh i see whats happening...

it actually is getting an invisible null exception, that gets swallowed up and not displayed for some reason (possibly because this is all running in a separate Task)

gmriggs: fixed

gmriggs:
we saw a monster with 136 radius try to spawn on CE, which caused a crash for min being > max, so this part needs investigated more

the classic tremendous monogua example: if one tries to spawn on the edge of a landblock, should it really be pushed all the way to the middle?

in that case, if the data says a tremendous monouga should spawn on the edge, it seems like loading the neighboring landblock would have to naturally be done in that case

PR #1766, opened 2019-04-20
Loot fixes/additions

Theran578:
- Add shirts and pants to loot drop
- Change weapon damage to be based upon wield difficulty, instead of tier number
- Lower Essence drop rate to ~17%
- Convert a number of magic numbers to descriptive, enum based values


Theran578:
Will need to redo the changes, as I cannot figure out what GitHub is doing.

PR #1768, opened 2019-04-20
DamageHistory no longer holds references to WorldObjects

Mag-nus:
This still needs more testing, mainly with healing, spells that cause self harm, etc..

This fixes the issue where DamageHistory was holding onto strong references to WorldObjects. This was mostly a problem when they were holding onto references to a Player. That would prevent the Player (and their session) from being cleaned up.

This is further a problem because a monster could hold onto these references indefinitely if they are never attacked again, but kept alive.

Attacking a monster is what causes DamageHistory pruning. It's possible that this could be switched to a check on creature Tick(). I'm not sure what is more performant.


Mag-nus:
I've switched the TryPrune() call to the creatures heartbeat. This should about the same performance, but be better at releasing stale data.

Mag-nus:
This seems to test ok.

Player/Session references are now being fully released.

PR #1779, opened 2019-04-22
updating tinkering message to match retail, adding /cisalvage

gmriggs:


Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -2355,5 +2355,39 @@ public static void HandleShowAllegiances(Session session, params string[] parame
                 Console.WriteLine("---------------");
             }
         }
+
+        [CommandHandler("cisalvage", AccessLevel.Admin, CommandHandlerFlag.None, 1, "Create a salvage bag in your inventory", "<material_type>, optional: <structure> <workmanship> <num_items>")]
```
LtRipley36706: cm command (HandleCM) i believe was for what this is doing

PR #1782, opened 2019-04-22
Shade adjustments

OptimShi:
* Fixed shade/palette calculation and validated against retail pcap logs
* Added Shade2, Shade3, Shade4 support (nothing in the data currently uses these)
* Removed "(based on STRENGTH 100)" from missile weapons.


OptimShi:
Need to double check some things... Thing I found a bug/flaw in logic.

PR #1792, opened 2019-04-24
fixed during great stress stack overflow exception at listener thread

fartwhif:
exception destroys the listener and crashes the server
added ignorance for certain socket errors

### DIFF
--- a/Source/ACE.Server/Network/ConnectionListener.cs
+++ b/Source/ACE.Server/Network/ConnectionListener.cs
@@ -2,7 +2,7 @@ using System;
 using System.Net;
 using System.Net.Sockets;
 using System.Text;
-
+using System.Threading;
 using ACE.Server.Managers;
 
 using log4net;
@@ -15,6 +15,10 @@ namespace ACE.Server.Network
         private static readonly ILog packetLog = LogManager.GetLogger(System.Reflection.Assembly.GetEntryAssembly(), "Packets");
 
         public Socket Socket { get; private set; }
+
+        private ManualResetEvent ReceiveComplete = new ManualResetEvent(false);
+
+        private bool Listening = true;
 
         private IPEndPoint listenerEndpoint;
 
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/Network/Managers/SocketManager.cs
```diff
@@ -35,8 +36,8 @@ public static void Initialize()
             listeners[1] = new ConnectionListener(host, ConfigManager.Config.Server.Network.Port + 1);
             log.Info($"Binding ConnectionListener to {host}:{ConfigManager.Config.Server.Network.Port + 1}");
 
-            listeners[0].Start();
-            listeners[1].Start();
+            new Thread(new ThreadStart(() => { listeners[0].Start(); })) { Name = $"{listeners[0].Socket.LocalEndPoint} Listener" }.Start();
```
Mag-nus: The listeners do not need to be created on new threads. The current thread they're on doesn't even blip on the CPU usage levels.

Please revert SocketManager changes.

Source/ACE.Server/Network/ConnectionListener.cs
```diff
@@ -50,26 +61,41 @@ public void Start()
...
-                log.DebugFormat("Network Socket has thrown: {0} {1}", socketException.ErrorCode, socketException.Message);
-                Listen();
-            }
-            catch (Exception exception)
+            while (Listening)
             {
-                log.FatalFormat("Network Socket has thrown: {0}", exception.Message);
+                try
+                {
+                    ReceiveComplete.Reset();
```
Mag-nus: Here, i'm just curious why we need a ManualResetEvent.

Can you please describe the scenario in which this function is being called concurrently?

I believe what we're seeing here is buffer corruption. We're accessing the same buffer that is being used to collect data from the socket as well. Is this accurate?

If so, some comments should be added because that'a  pretty big find, and something I could see easily accidentally being "optimized" out in the future.

fartwhif: when spamming connections at the server via an external application it was causing a stack overflow

I tried a few methods to solve the problem which failed and this is the first thing that worked which is the first one i tried based on standard practices outlined by Microsoft [here](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.beginreceivefrom?view=netframework-4.8)

Mag-nus:
Here is the documentation for .NET Core 2.2:
https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.beginreceivefrom?view=netcore-2.2

After reading the documentation, It seems that both current master and this PR implement the system correctly. That might suggest that the underlying issue is present in both implementations, just occurs less in this PR's implementation.

Can you please share your test code so I can run some tests as well?

Thanks.

Mag-nus:
I'm closing this.

The problem as described is replicated on a bench over a LAN, but may not be over the internet.

Even so, the proposed solution changes the threading model which I believe to be unnecessary. In the event this issue does happen on a internet hosted server, hopefully we can address it without changing the threading model and instead changing how we use the Socket and it's callback.

PR #1817, opened 2019-04-27
Fixing monster retired skill #s

gmriggs:


Source/ACE.DatLoader/PortalDatDatabase.cs
```diff
@@ -16,6 +16,7 @@ public PortalDatDatabase(string filename, bool keepOpen = false) : base(filename
             RegionDesc = ReadFromDat<RegionDesc>(RegionDesc.FILE_ID);
             SecondaryAttributeTable = ReadFromDat<SecondaryAttributeTable>(SecondaryAttributeTable.FILE_ID);
             SkillTable = ReadFromDat<SkillTable>(SkillTable.FILE_ID);
+            SkillTable.AddRetiredSkills();
```
Mag-nus: To keep PortalDatDatabase.cs more reusable, the AddRetiredSkills() call should probably be moved to DatManager

A 3rd party tool may want to use PortalDatDatabase in the future. If they query SkillTable, it should show actual results without additional ACE helpers.

PR #1832, opened 2019-04-30
Wire up Ivory salvage application

Theran578:
- Add ability to apply Ivory salvage to Attuned items

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -1149,6 +1149,10 @@ namespace ACE.Server.WorldObjects
             if (!PropertyManager.GetBool("use_wield_requirements").Item)
                 return WeenieError.None;
 
+            var allowedWielder = item.GetProperty(PropertyInstanceId.AllowedWielder);
+            if (allowedWielder != null && (allowedWielder != Guid.Full))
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -1149,6 +1149,10 @@ private WeenieError CheckWieldRequirement(WorldObject item)
             if (!PropertyManager.GetBool("use_wield_requirements").Item)
                 return WeenieError.None;
 
+            var allowedWielder = item.GetProperty(PropertyInstanceId.AllowedWielder);
+            if (allowedWielder != null && (allowedWielder != Guid.Full))
+                return WeenieError.YouDoNotOwnThatItem; // Unsure of the exact message
```
LtRipley36706: I know this doesn't fit the pattern but it might be WeenieErrorWithString.Only_MayUseItemMagic

Theran578: That message seems to fit more with the AllowedActivator, rather than the AllowedWielder.  To me, with AllowedActivator present, one could at least equip the item.  They would just be prevented from using the spell or what have you upon it.

LtRipley36706: you're probably right, my only thought was maybe the client interpreted it differently depending what it knows about that object

PR #1833, opened 2019-04-30
Upgrade to .NET Core 2.2.1 (VS 2017 compatible)

Mag-nus:
This also updates all other nuget packages to latest.

The EF database packages are updated to 2.2.0 because that is the latest pomelo (our database driver) currently supports.

This is an effort to help fix the CE issues


Mag-nus:
Note, for developers, the following SDK will need to be installed:
.NET Core 2.2 (x64 SDK v2.2.106) for Visual Studio 2017 SDK

Which can be obtained here:
https://dotnet.microsoft.com/download/visual-studio-sdks

Mag-nus:
CE is running .net core 2.2 so they should not need to update if this PR is merged.

Theran578:
I've already been running .net 2.2 on my server for awhile now, with no adverse effects

PR #1841, opened 2019-05-01
updating trade for some decal plugins

gmriggs:


Source/ACE.Server/WorldObjects/Player_Trade.cs
```diff
@@ -62,6 +62,9 @@ public void HandleActionOpenTradeNegotiations(uint tradePartnerGuid, bool initia
 
                     Session.Network.EnqueueSend(new GameEventRegisterTrade(Session, Guid, tradePartner.Guid));
 
+                    // was this sent by retail?
+                    Session.Network.EnqueueSend(new GameEventResetTrade(Session, Guid));
```
LtRipley36706: pcaps show it wasn't sent at start of trade, plugins may be looking for it, but it wasn't there originally.

PR #1843, opened 2019-05-02
Some CombatPet tweaks

LtRipley36706:


Source/ACE.Server/Managers/EnchantmentManagerWithCaching.cs
```diff
@@ -13,18 +13,18 @@ namespace ACE.Server.Managers
 {
     public class EnchantmentManagerWithCaching : EnchantmentManager
     {
-        private bool? hasEnchantments;
+        //private bool? hasEnchantments;
```
gmriggs: why is this removed?

this will severely hamper the speed of the server. cpu usage on CE will rise dramatically without caching hasEnchantments

if something is broken with hasEnchantments caching, then fix it... don't remove it entirely.

gmriggs: "Revert Caching of HasEnchantments (Players with no spells do not start ticking until relog)"

this is incorrect, just tested this:

/dispel
verified all enchantments dispelled
relogged
/buff
appraise character
/getenchantments
StartTime ticks backwards

LtRipley36706: it wasn't incorrect, your test just didn't expose the issue because it needed the character to be complete void of any enchantments and only cooldowns being used.
The issue was a StartCooldown down didn't clear the cache so if there were no enchantments, and you fired something that would start a cooldown, it would never tick unless an enchantment from another category was used. Edge case but was easily reproduced under those circumstance. I removed the code as I didn't understand how to fix it at the time, but you later pointed out the correct fix which I've added, so we're back to all good =)

PR #1851, opened 2019-05-04
Set player masteries at character creation

Theran578:
- Set player masteries at character creation for melee and ranged
- Add ability for summoning mastery to be displayed in client but not set at character creation, as that should be controlled by player choice, starting at level 50


Source/ACE.Server/Factories/PlayerFactory.cs
```diff
@@ -453,6 +460,65 @@ private static WorldObject GetClothingObject(uint weenieClassId, uint palette, d
...
+        /// <summary>
+        /// Set Heritage based Melee and Ranged Masteries
+        /// </summary>
+        /// <param name="heritageGroup"></param>
+        /// <param name="meleeMastery"></param>
+        /// <param name="rangedMastery"></param>
+        /// <param name="summoningMastery"></param>
+        private static void GetMasteries(HeritageGroup heritageGroup, out WeaponType meleeMastery, out WeaponType rangedMastery, out SummoningMastery summoningMastery)
+        {
+            summoningMastery = (SummoningMastery)ThreadSafeRandom.Next(1, 3);
```
mcreedjr: Summoning Mastery should not be set at character creation, but rather post creation by using one of three altars present in Arwic. Additionally, one must be over level 50 and I believe be specialized in summoning before being allowed to use one of the altars to choose a specialization. Source: https://asheron.fandom.com/wiki/Summoning

Theran578: Removed.  I'm assuming that until the mastery is chosen, players shouldn't be able to use the loot dropped essences and that the mastery will lock out the other two classes of summons that the one chosen?

mcreedjr: Correct on both accounts. The mastery will need checked on use of the Essence.

PR #1864, opened 2019-05-07
standardized spells and cantrips for armor pieces

dgarson:
There are currently many missing cantrip spells from armor pieces. For example, only head/gaunts can come with cantrips for war magic, or have focus on them. All of the armor pieces should be eligible for all of the spells that can come on any armor coverage, rather than select spells being candidates for only some armor coverage.


dgarson:
Seems reasonable. Shall I introduce another static array and reuse it for all the coverage slots?

Theran578:
If each armor piece has the same potential spells and/or cantrips, then those separate arrays should be compressed down to fewer.  Jewelry obviously needs to have its own, as those items don't have impen/banes.

Theran578:
On the right track.  ClothingCantrips looks to me much the same, also.  I imagine that the the normal spells would be in similar need of compression and standardization.  Lastly, if those spell arrays are being condensed down to single arrays, the better option would be to remove the switch statements in LootGenerationFactory_Clothing.cs in favor of spells = LootTables.ArmorSpells and cantrips = LootTables.ArmorCantrips

dgarson:
Thanks for the feedback, Theran! Let me know if there's any more changes you'd recommend.

Theran578:
Looks good

dgarson:
Woops, didn't mean to close without merging. Didn't realize I have to wait for you to merge :-)

PR #1872, opened 2019-05-10
Skill XP Rank increases

Theran578:
- Rough fix to SpendSkillXP for arbitrary Rank increases


Theran578:
Still need to test removing the "usage" flag, as it is probably not needed

PR #1881, opened 2019-05-12
Add OnGeneration Emote handling

LtRipley36706:


Source/ACE.Server/WorldObjects/WorldObject_Tick.cs
```diff
@@ -77,7 +77,7 @@ public void GeneratorHeartbeat(double currentUnixTime)
             if (cachedRegenerationInterval > 0)
                 NextGeneratorHeartbeatTime = currentUnixTime + cachedRegenerationInterval;
             else
-                NextGeneratorHeartbeatTime = double.MaxValue;
+                NextGeneratorHeartbeatTime = currentUnixTime + 1;
```
Mag-nus: You should prob have a < 0 check here that sets the next to double.MaxValue and only do the + 1 explicitly on cachedregenint == 0

LtRipley36706: reverting change for now.. needs broader change

PR #1892, opened 2019-05-13
Adjustment to loot percentages

Theran578:
If my math is correct and the code is doing what I expect it should be doing with the comparisons, this PR should match Epic drop rates to the rough estimates provided by Mag-nus in dev.  I somewhat matched the rarity of multi-majors to be similar to Epics. If further adjustments or refinements are needed, they can be performed in SpellDistribution(), GetNumMajorCantrips(), GetNumEpicCantrips, and GetNumLegendayCantrips().

Currently, there isn't any other kind of ordering for cantrip spell additions than minor, major, epic, and lastly legendary, nor is there any coding to preclude lesser or greater cantrip or adjusting their percentages based upon what was rolled beforehand, if that type of process was performed.

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -1013,9 +1013,9 @@ namespace ACE.Server.Factories
             {
                 case 1:
                     // 1-3 w/ chance of minor cantrip
-                    if (nonCantripChance < 50)
+                    if (nonCantripChance <= 50)
                         numNonCantrips = 1;
-                    else if (nonCantripChance < 90)
+                    else if (nonCantripChance <= 90)
                         numNonCantrips = 2;
                     else
                         numNonCantrips = 3;
@@ -1027,7 +1027,7 @@ namespace ACE.Server.Factories
 
                 case 2:
                     // 3-4 w/ chance of either minor or major
-                    if (nonCantripChance < 50)
+                    if (nonCantripChance <= 50)
                         numNonCantrips = 3;
                     else
                         numNonCantrips = 4;
@@ -1041,7 +1041,7 @@ namespace ACE.Server.Factories
 
                 case 3:
                     // 4-5 w/ chance of either major or minor
-                    if (nonCantripChance < 50)
+                    if (nonCantripChance <= 50)
                         numNonCantrips = 4;
                     else
                         numNonCantrips = 5;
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1212,15 +1212,18 @@ private static int GetNumEpicCantrips(int tier)
...
+            // 0.1% chance for 3 Epics, 0.001% chance for 4 Epics 
+            if (epicCantripChance < 20000)
                 numEpics = 1;
-            else if (epicCantripChance > 95)
+            else if (epicCantripChance < 1000)
                 numEpics = 2;
-            else if (epicCantripChance > 98)
+            else if (epicCantripChance < 100)
                 numEpics = 3;
+            else if (epicCantripChance < 1)
```
Mag-nus: This should be <= 1 i think

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1234,10 +1237,10 @@ private static int GetNumLegandaryCantrips(int tier)
 
             int legendaryCantripChance = ThreadSafeRandom.Next(1, 100);
 
-            // 10% chance for a legendary, 5% chance for 2 legendaries
-            if (legendaryCantripChance > 90)
+            // 10% chance for a legendary, 1% chance for 2 legendaries
+            if (legendaryCantripChance < 10)
                 numLegendaries = 1;
-            else if (legendaryCantripChance > 95)
+            else if (legendaryCantripChance < 1)
```
Mag-nus: This should also be <= 1 i think.

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1164,40 +1164,40 @@ private static int GetNumMajorCantrips(int tier)
...
+                    else if (majorCantripChance < 1000)
                         numMajors = 2;
-                    else if (majorCantripChance > 90)
+                    else if (majorCantripChance > 100)
                         numMajors = 3;
-                    else if (majorCantripChance > 95)
+                    else if (majorCantripChance > 10)
                         numMajors = 4;
-                    else if (majorCantripChance > 98)
+                    else if (majorCantripChance > 1)
```
Mag-nus: >= 1? not sure.

Just thinking about what happens when the random is the min or max values in the range provided.

Theran578: Forgot to flip the > symbol to <

PR #1900, opened 2019-05-15
 adding dispells, summon inepts, calming gaze, flame bolts to scroll loot table, also fix raise vital message

dirtyelf:


Source/ACE.Server/WorldObjects/Entity/CreatureVital.cs
```diff
@@ -76,7 +76,7 @@ public uint Base
         {
             get
             {
-                var attr = AttributeFormula.GetFormula(creature, Vital);
+                var attr = AttributeFormula.GetFormula(creature, Vital, false);
```
gmriggs: hm that looks reasonable for Base maybe, but it breaks Current/MaxValue
dunno if you are familiar with the /showstats command, but it is very useful for testing these scenarios
MaxValue would also need to be changed to this:
```
        public uint MaxValue
        {
            get
            {
                var attr = AttributeFormula.GetFormula(creature, Vital, true);

                uint total = StartingValue + Ranks + attr;
```

dirtyelf: makes sense, thanks for reviewing

Source/ACE.Server/WorldObjects/Entity/CreatureVital.cs
```diff
@@ -92,12 +92,9 @@ public uint MaxValue
         {
             get
             {
-                uint total = Base;
+                var attr = AttributeFormula.GetFormula(creature, Vital, true);
 
-                var multiplier = creature.EnchantmentManager.GetVitalMod_Multiplier(this);
```
gmriggs: this should still be getting the enchantments here

```
var attr = AttributeFormula.GetFormula(creature, Vital, true);

                uint total = StartingValue + Ranks + attr;

                var multiplier = creature.EnchantmentManager.GetVitalMod_Multiplier(this);
                var additives = creature.EnchantmentManager.GetVitalMod_Additives(this);

                total = (uint)Math.Round(total * multiplier + additives);
```

dirtyelf: what do those account for? are they not included in the AttributeForumula? my char vitals were properly updating with endurance/willpower without them

dirtyelf: looks like these may be cantrips, essence glutton, life giver, etc...

gmriggs: EnchantmentManager would be all of the enchantments

GetVitalMod_Multiplier would be things like asheron's benediction
GetVitalMod_Additive would be things like essence glutton

PR #1905, opened 2019-05-17
Add AdjustDungeon to teleports via magic

LtRipley36706:


Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -794,7 +794,13 @@ protected EnchantmentStatus ItemMagic(WorldObject target, Spell spell, WorldObje
                                 ActionChain portalRecall = new ActionChain();
                                 portalRecall.AddAction(targetPlayer, () => targetPlayer.DoPreTeleportHide());
                                 portalRecall.AddDelaySeconds(2.0f);  // 2 second delay
-                                portalRecall.AddAction(targetPlayer, () => targetPlayer.Teleport(portal.Destination));
+                                portalRecall.AddAction(targetPlayer, () =>
+                                {
+                                    var teleportDest = new Position(portal.Destination);
+                                    targetPlayer.AdjustDungeon(teleportDest);
```
gmriggs: should this be done in .Teleport(), instead of having all of these calls to it?

i seem to remember .AdjustDungeon() being done in Teleport originally, but im not sure why that was removed...

PR #1908, opened 2019-05-18
reverting 1874 for RegenOnClose chest generators

gmriggs:


Mag-nus:
#1917 fixes the issue caused by #1874

PR #1910, opened 2019-05-19
Update LootTables.cs

LtRipley36706:
Remove Mana Scarab from loot table


LtRipley36706:
#1927 has this as well

PR #1914, opened 2019-05-20
Confirm both parties can accept traded items

LtRipley36706:


Source/ACE.Server/Network/GameAction/Actions/GameActionAcceptTrade.cs
```diff
@@ -6,6 +6,13 @@ public static class GameActionAcceptTrade
         [GameAction(GameActionType.AcceptTrade)]
         public static void Handle(ClientMessage message, Session session)
         {
+            uint partnerGuid = message.Payload.ReadUInt32();
+            ulong tradeStamp = message.Payload.ReadUInt64();
+            uint tradeStatus = message.Payload.ReadUInt32();
+            uint initiatorGuid = message.Payload.ReadUInt32();
+            bool initatorAccepts = message.Payload.ReadBoolean();
```
gmriggs: is the structure of this packet confirmed?

this would be something unseen before in ac protocol: whenever a boolean is sent, it's usually sent as an int

tradeStamp - have never seen a field sent as 64bit before

the total packet size here would also be 162 bits, which would need an .Align() to get it back in sync with the 4-byte dword boundary

Source/ACE.Server/WorldObjects/Player_Trade.cs
```diff
@@ -161,6 +162,52 @@ public void HandleActionAcceptTrade(Session session, ObjectGuid whoAccepted)
...
+                        var wo = target.GetInventoryItem(itemGuid);
+
+                        if (wo == null)
+                            wo = target.GetEquippedItem(itemGuid);
+
+                        if (wo != null)
+                            playerBitems.Add(wo);
+                    }
+
+                    var playerACanAddToInventory = session.Player.CanAddToInventory(playerBitems);
```
gmriggs: what would happen in this scenario?

both players have every pack slot full, and they each try to trade 1 item with each other

LtRipley36706: As coded, trade would fail. It doesn't subtract inventory/burden from items the trade partner is trading away from their inventory

PR #1918, opened 2019-05-20
Partial #1902 Revert

Mag-nus:
Switch back to int to avoid the type casting.

We now check to see if amount is <= 0 instead of just == 0 to avoid packet crafting exploits.

The additional messages remain.

A new logging message has been added for <= 0 events detected.

### DIFF
--- a/Source/ACE.Server/Network/GameAction/Actions/GameActionStackableMerge.cs
+++ b/Source/ACE.Server/Network/GameAction/Actions/GameActionStackableMerge.cs
@@ -12,7 +12,7 @@ namespace ACE.Server.Network.GameAction.Actions
         {
             uint mergeFromGuid = message.Payload.ReadUInt32();
             uint mergeToGuid = message.Payload.ReadUInt32();
-            uint amount = message.Payload.ReadUInt32();
+            int amount = message.Payload.ReadInt32();
 
             session.Player.HandleActionStackableMerge(mergeFromGuid, mergeToGuid, amount);
         }
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -107,7 +107,9 @@ namespace ACE.Server.WorldObjects
                 if (stack == null || stackRootOwner == null)
                     return false;
 
-                AdjustStack(stack, -amount, stackFoundInContainer, stackRootOwner);
+                if (!AdjustStack(stack, -amount, stackFoundInContainer, stackRootOwner))
+                    return false;
+
                 Session.Network.EnqueueSend(new GameMessageSetStackSize(stack));
             }
 
@@ -533,8 +535,14 @@ namespace ACE.Server.WorldObjects
         /// <summary>
         /// If you want to subtract from a stack, amount should be negative.
         /// </summary>
-        private void AdjustStack(WorldObject stack, int amount, Container container, Container rootContainer)
+        private bool AdjustStack(WorldObject stack, int amount, Container container, Container rootContainer)
         {
+            if (stack.StackSize + amount <= 0 || stack.StackSize + amount > stack.MaxStackSize)
+            {
+                log.WarnFormat("Player 0x{0:X8}:{1} tried to adjust stack by an invalid amount amount ({4}) 0x{2:X8}:{3}.", Guid.Full, Name, stack.Guid.Full, stack.Name, amount);
+                return false;
+            }
+
             stack.StackSize += amount;
             stack.EncumbranceVal = (stack.StackUnitEncumbrance ?? 0) * (stack.StackSize ?? 1);
             stack.Value = (stack.StackUnitValue ?? 0) * (stack.StackSize ?? 1);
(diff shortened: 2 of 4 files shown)


Source/ACE.Server/Network/GameAction/Actions/GameActionStackableMerge.cs
```diff
@@ -12,7 +12,7 @@ public static void Handle(ClientMessage message, Session session)
         {
             uint mergeFromGuid = message.Payload.ReadUInt32();
             uint mergeToGuid = message.Payload.ReadUInt32();
-            uint amount = message.Payload.ReadUInt32();
+            int amount = message.Payload.ReadInt32();
```
gmriggs: amount is an unsigned data type, as defined by the ac network protocol

Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -1275,10 +1275,11 @@ private WeenieError CheckWieldRequirement(WorldObject item)
         /// - try to split a stack off of the landblock into a container
         /// - try to split a stack into a different container that doesn't already have a stack that can support a merge
         /// </summary>
-        public void HandleActionStackableSplitToContainer(uint stackId, uint containerId, uint placementPosition, uint amount)
+        public void HandleActionStackableSplitToContainer(uint stackId, uint containerId, int placementPosition, int amount)
```
gmriggs: amount can never be negative here... this is inherently unsafe, and leads to the type of critical errors that were previously demonstrated with amounts

amount is represented as an unsigned data type in both acclient, and the original ac server

the only reason the existing change needed to perform casting, is because many other functions throughout the system are still representing amount as a signed data type.

with amount correctly represented as an unsigned data type throughout all of the game code, it will be provable that amount can never go negative, and it will also avoid this type of casting

Mag-nus: What you're saying is correct, but the round about way of handling it.

The simple solution is as i presented, <= 0.

It's much safer to check your input before you process it to make sure that it's valid, rather than to enforce non-negative numbers by using an unsigned data type.

Furthermore, because this value is used in arithmetic operations, the results then become longs which have to then be type casted back down to int (the native data store).

Simply checking for <= 0 and using an int avoids countless type casting, and in practice, is safer.

Remember, we're talking about a value with actual ranges of 0-10,000.

gmriggs: "The simple solution is as i presented, <= 0."

Which fixes it for this function... with signed data types, it differs from: the ac network protocol, acclient, and the original ac server

The rest of the codebase would need to be audited, everywhere amounts are used, to ensure the correct bounds checks are being performed

In the very first function I see being called, SetStackSize(), it accepts a negative number

"It's much safer to check your input before you process it to make sure that it's valid, rather than to enforce non-negative numbers by using an unsigned data type."

One is not exclusive of the other. The input would always be validated first.

"Furthermore, because this value is used in arithmetic operations, the results then become longs"

This is incorrect. For this data type, you are never mixing uints and ints. Amounts are always uints. You are either setting the stack size directly (unsigned), adding another amount (unsigned), or subtracting another amount (unsigned).

"Simply checking for <= 0 and using an int avoids countless type casting, and in practice, is safer."

See above. Amount really is a simple unsigned field with only 1-2 fundamental operations, not "countless arbitrary arithmetic operations" as being incorrectly implied here. If it was, and uints/ints had to be constantly mixed together, and longs cast back, then yes, this would be valid in this scenario. However, that is not the case here.

"Remember, we're talking about a value with actual ranges of 0-10,000."

Amounts can be much larger than 10k. Mansion purchase requires a pyreal amount of 10 million. There could be even larger amounts, or as you mentioned, someone could write a mod here using anything that is valid in acclient.

Deviating from: ac network protocol, acclient, and the original ac server is probably not the best idea...

Mag-nus: Please remember that we're talking about JUST the stack size amounts for these 3 functions. Your reply broadens the scope.

"The rest of the codebase would need to be audited, everywhere amounts are used, to ensure the correct bounds checks are being performed"

That is incorrect. The very first line of the HandleAction*() function does the <= 0 check. All 3 of these have that check.

"One is not exclusive of the other. The input would always be validated first."

You miss the point. You're saying that we should be using uint to ensure non-negative numbers. What i'm saying, and what is implemented here, is that you don't rely on your data type to constrain your values, you rely on conditions you implement.

"This is incorrect. For this data type, you are never mixing uints and ints. Amounts are always uints. You are either setting the stack size directly (unsigned), adding another amount (unsigned), or subtracting another amount (unsigned)."

(uint)4 - (uint)5 = rollover, so you still have issues, but now you're checking if your value is > someValue instead of <= 0 (again)

"Amounts can be much larger than 10k. Mansion purchase requires a pyreal amount of 10 million. There could be even larger amounts, or as you mentioned, someone could write a mod here using anything that is valid in acclient."

Those amounts aren't stack sizes. The scope of this PR just applies to stack sizes, and these 3 functions, not other unsigned/signed network originated values.

"Deviating from: ac network protocol, acclient, and the original ac server is probably not the best idea..."

How the value comes across the wire, and how we use it are completely separate.

PR #1924, opened 2019-05-21
add god admin command

dirtyelf:
add /god admin command functionality

- sets level to 999
- specializes every skill
- sets attributes to 9999
- level up and BZ particle effects


dirtyelf:
any idea what the retail command did specifically?

LtRipley36706:
> any idea what the retail command did specifically?

from what I remember it just set whatever currently spec/trained skills to a specific level (not sure what), and i *think* changed the level to match that same number...

it's probably not that big of a deal being an admin only function

dirtyelf:
this is definitely more fun, and serves the same purpose.

i can not figure out why i need two calls to setmaxvitals() to get the hp to update in the client. it was working back when i was first looking at this, now it needs the second call. stam and mana seem to update just fine with the first call. also it appears to only be the client, calling /showstats reports the correct values in the terminal for health - just not in the client.

ultimately not a huge deal, just leave the second call to setmaxvitals in there... but it may be something that affects other areas so worth noting

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1484,13 +1484,54 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+                session.Player.Session.Network.EnqueueSend(new GameMessagePrivateUpdateVital(session.Player, v.Key, playerVital.Ranks, playerVital.StartingValue, playerVital.ExperienceSpent, playerVital.Current));
+            }
+
+            session.Player.SetMaxVitals();
+
+            session.Player.PlayParticleEffect(PlayScript.LevelUp, session.Player.Guid);
+            session.Player.PlayParticleEffect(PlayScript.BaelZharonSmite, session.Player.Guid);
 
-            // output: You are now a god!!!
+            ChatPacket.SendServerMessage(session, $"{session.Player.Name} is now a god!", ChatMessageType.WorldBroadcast);
```
LtRipley36706: revert this back to `You are now a god!!!` and `ChatMessageType.Broadcast` please as that part was correct as is.

you may also want to add a `session.Player.ChangesDetected=true;` line somewhere before save. since we've kinda gone outside the normal scope for updating skill/stats/vitals

dirtyelf: i was wondering if i used the save function correctly, thanks

PR #1984, opened 2019-06-10
add characters to be deleted to name in use check

dirtyelf:
prevent character creation with names of characters that are not fully deleted and subsequently restoring those characters for a duplicate name


dirtyelf:
So turns out I am mistaken and in retail if you deleted a character you were able to create another character with the same name immediately and get an error when trying to restore if someone has taken that name.

PR #1991, opened 2019-06-11
[WIP] New player movement, advanced spellcasting

gmriggs:
- To enable detailed runtime state info, set DebugPlayerMoveToStatePhysics=true
- All of the advanced spellcasting techniques should now be available (accurate slidecasting, fastcast, holdcast)
- Server physics / motion system now has precise hooks into the current state, which should match up closely with the client (minus lag)
- Server positioning should be more accurate, and should match the client closely at any given point in time
- Fall damage should now be more accurate and consistent
- Hotspot and pressure plate collision detection should now be more accurate


gmriggs:
glad to hear it

there are still a lot of very precise bugs that need to be ironed out with this pr, and it is still a wip that should not be checked into master yet

also of concern is the slightly higher cpu usage. for a server like CE to be running this, mag's dungeon multithreading should be checked in first

Mag-nus:
I need some help verifying what i believe is a new, unintended behavior.

Location: Pit of Egg Orchard. Using this map:
https://vignette.wikia.nocookie.net/asheron/images/7/7c/002B.gif/revision/latest?cb=20141117102648
The location is in the octagon shaped room, on the bottom part.
Your location is: 0x002B01A2 [184.551422 -96.884575 -53.993999] -0.460788 0.000000 0.000000 -0.887510

VTank options:
Ghost detection off
Character type: Melee

In current master, you can sticky to this location and macro endlessly.
- Monsters that land on your head find a safe way down.
- Corpses end up on the floor
- No ghost monsters remain in the client
- About 1-2 creatures per day will attack the player that the client has either not detected, wasn't informed about, or forgot about. Manually attacking or smiting these creatures works just fine. The client otherwise just works around these creatures.

In this PR:
- Monsters may remain on your head and not find their way down safely
- Corpses of those creatures may end up at head level
- The client ends up with ghost monsters where the monster is in the floor. Doing a /smite all does not delete the creature from the client.

gmriggs:
This PR has been merged into master branch

PR #2000, opened 2019-06-14
fixing oak salvage

gmriggs:


dirtyelf:
Doesn't it also need to be applied to missile weapons? Is that taken care of somewhere else?

gmriggs:
Can you link an example of a Missile Weapon with a speed?

gmriggs:
you are correct, missile weapons do indeed have a speed

for some reason, the /ci yumi that i examined did not have a speed shown in appraisal, even though it appears to have a speed in the data

in addition to this, i don't think any ace code is actually evaluating the speed of missile weapons yet...

dirtyelf:
missile wep speed is used for the reload animation

gmriggs:
is it used for the 'launch arrow' animation + the reload animation, or just the reload anim?

PR #2012, opened 2019-06-17
Hookers can only be used on Hooks

LtRipley36706:


Source/ACE.Server/WorldObjects/Hooker.cs
```diff
@@ -33,5 +35,22 @@ public override void ActOnUse(WorldObject activator)
         {
             // handled in base.OnActivate -> EmoteManager.OnUse()
         }
+
+        public override ActivationResult CheckUseRequirements(WorldObject activator)
+        {
+            if (!(activator is Player player))
+                return new ActivationResult(false);
+
+            if (Owner != null && Owner.WeenieType != WeenieType.Hook)
```
gmriggs: base requirements should be checked first?

PR #2027, opened 2019-06-22
add /ungod command

dirtyelf:
The god command now only lasts until you log out. When you relog you will be at lifestone and any items you obtained while a god will be lost.

This works by saving to the db and flagging the character as 'DoNotSave' before altering the stats. When the player logs out the stats are reloaded from the db before adding to the offline players list.

Definitely open to feedback and suggestions.


Source/ACE.Server/WorldObjects/WorldObject_Database.cs
```diff
@@ -132,6 +155,9 @@ public bool IsStaticThatShouldPersistToShard()
         /// <returns></returns>
         public bool IsDynamicThatShouldPersistToShard()
         {
+            if (DoNotSave)
```
Mag-nus: DoNotSave is not a condition for this function

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
```diff
@@ -95,6 +115,9 @@ public void RemoveBiotaFromDatabase(bool enqueueRemove = true)
         /// </summary>
         public bool IsStaticThatShouldPersistToShard()
         {
+            if (DoNotSave)
```
Mag-nus: DoNotSave is not a condition for this function

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
```diff
@@ -18,7 +18,25 @@ partial class WorldObject
         /// This variable is set to true when a change is made, and set to false before a save is requested.<para />
         /// The primary use for this is to trigger save on add/modify/remove of properties.
         /// </summary>
-        public bool ChangesDetected { get; set; }
+        private bool changesDetected;
```
Mag-nus: This property shouldn't be checking DoNotSave

Source/ACE.Server/WorldObjects/Player_Spells.cs
```diff
@@ -71,7 +71,7 @@ public void HandleActionMagicRemoveSpellId(uint spellId)
                 return;
             }
 
-            ChangesDetected = true;
+            if (!DoNotSave) { ChangesDetected = true; }
```
Mag-nus: ChangesDetected should still be set, and shouldn't be conditional on DoNotSave

Source/ACE.Server/WorldObjects/Player_Database.cs
```diff
@@ -82,6 +102,8 @@ public void SavePlayerToDatabase()
 
         public void SaveCharacterToDatabase()
         {
+            if (DoNotSave) { return; }
```
Mag-nus: Character should still be saved right?

Doesn't God only affect the biota?

Source/ACE.Server/WorldObjects/Player_Database.cs
```diff
@@ -56,7 +74,9 @@ public void RushNextPlayerSave(int seconds)
         /// </summary>
         public void SavePlayerToDatabase()
         {
-            if (CharacterChangesDetected)
+            if (DoNotSave) { return; }
```
Mag-nus: The check shouldn't be here, when it already exists at the WorldObject level

Source/ACE.Server/WorldObjects/Player_Database.cs
```diff
@@ -18,7 +18,25 @@ partial class Player
         /// This variable is set to true when a change is made, and set to false before a save is requested.<para />
         /// The primary use for this is to trigger save on add/modify/remove of properties.
         /// </summary>
-        public bool CharacterChangesDetected { get; set; }
+        private bool characterChangesDetected;
+        public bool CharacterChangesDetected
```
Mag-nus: ChangesDetected should still be set even if DoNotSave is set

Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -40,6 +40,7 @@ public partial class Player : Creature, IPlayer
 
         public bool LastContact = true;
         public bool IsJumping = false;
+        //public bool DoNotSave = false;
```
Mag-nus: ???

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
dirtyelf: What do you mean is not a condition for this function? It checks the WorldObject property DoNotSave which is set in the god function. Not being hostile - legitimately trying to learn. Are these static (and the dynamic one below) not related to players, only for items placed on hooks or in chests?

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
dirtyelf: Is there a better way to accomplish this? Why should it not check DoNotSave?

Source/ACE.Server/WorldObjects/Player_Spells.cs
dirtyelf: I forgot to remove this - it was added in some of my early testing.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: Yes I believe only ignoring the biota could work. I thought it might be better to not save anything instead of saving some things and not others.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: Yeah this ends up being a catch all I guess. If I never set changes detected to true it should never run anyways, is that what you are referring to?

Source/ACE.Server/WorldObjects/Player.cs
dirtyelf: Left over from initial testing, meant to remove.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: It would always be set by the else case to the value, right? It should only force false if DoNotSave evaluates to true.

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
Mag-nus: This function returns a bool that indicates if the wo is the type of object that should persist to shard.

It does not return the bool that indicates both should persist to shard and can persist to shard.

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
Mag-nus: You don't mess with ChangesDetected.

You manage your Save/Don'tSave in SaveBiota

Changes SHOULD still be detected even if DoNotSave is set. ChangesDetected has nothign to do with save conditions

gmriggs:
what about a completely different approach to this?

when a player enters /god mode, serialize the current state of all their affected properties, and save it to a new PropertyString

add an /ungod command, which restores the character state

dirtyelf:
what happens if the server crashes before you /ungod?

gmriggs:
then you just /ungod at any point in the future, after the server has restarted

that's the point of saving it in a PropertyString. it's persisted to the db

dirtyelf:
When you say PropertyString do you mean creating a new entry in the db? Or do you mean serializing the data and saving it in something that is already existing? Is there an example of something that is currently saved like this?

That could also work but isn't sort of the same thing - just on a different (user controllable) timescale? The way it is setup now it forces the reload on log out (or total reload if the server crashes) which should create shorter 'unsaved' time periods. /ungod would just force the reload when the user chooses.

I appreciate the feedback. I'm open to implementing this however we agree is best - or totally dropping it if we agree it is useless, or just too dangerous.

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
dirtyelf: Understood. This check is not needed here. Thank you.

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
dirtyelf: After looking back through it you are right. We can and should still detect changes. Removed this logic.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: This logic is not needed, the check is done at the SaveCharacterToDatabase function.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: What all is contained in character vs biota? I tried settings like /day and side by side vitals, even with this check in there the settings were retained - as a god.

gmriggs:
in PropertyString.cs:

```
        /* custom */
        [ServerOnly]
        AllegianceMotd                  = 9001,
        [ServerOnly]
        AllegianceMotdSetBy             = 9002,
        [ServerOnly]
        AllegianceSpeakerTitle          = 9003,
        [ServerOnly]
        AllegianceSeneschalTitle        = 9004,
        [ServerOnly]
        AllegianceCastellanTitle        = 9005,
        [ServerOnly]
        GodState                        = 9006,
```

gmriggs:
It's just that not saving 1 object in the world, but saving all the others, could possibly leave the world in a very inconsistent state

For example, consider this scenario: i go into god mode, then I go through a lot of inventory state changes... I drop an item into the landblock, I give an item to another player. Then I log off.

The state of the world would be: the item would be updated, and it would now be owned by the landblock/other player, but there would now be duplicate references to the item: the item would exist in the landblock / other player's inventory, but my own inventory would still contain a reference to it, so it might still appear there as well. There won't actually be a duplicate item, but it will be 2 inconsistent references to the same item, and this can throw all of the world data into an inconsistent state.

gmriggs:
It's providing similar functionality, but doing it in a much cleaner / consistent way. The current implementation is bending over backwards to do very weird things that doesn't match up with the way things are designed. The state of the world should be a singular, consistent concept -- the idea of "persist the entire state of the world, EXCEPT for these particular objects" is not a good idea, and gets into lots of possible loopholes and inconsistent states in the data. There is no "don't save this particular object to the db" flag for individual objects for a good reason.

This really should be a simple feature, and shouldn't require tendrils into all of the core / underlying systems on the server. The player goes into god mode, and at some time they might want to revert. So give them that ability to revert. Save the state of all the fields affected by god mode. Restore that state.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: removed

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: removed

dirtyelf:
i liked gm's approach so i ran with it

i think this now gives the appropriate level of functionality without the danger of not saving to the db

Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -69,6 +69,8 @@ public Player(Weenie weenie, ObjectGuid guid, uint accountId) : base(weenie, gui
             Attackable = true;
 
             SetProperty(PropertyString.DateOfBirth, $"{DateTime.UtcNow:dd MMMM yyyy}");
+
+            SetProperty(PropertyString.GodState, $"0={DateTime.UtcNow}");
```
Mag-nus: I'm not sure this is needed.

When you ungod, you should RemoveProperty(). You don't need to set a default property here. The absence of the property should mean that there is no UnGod state

Source/ACE.Server/WorldObjects/Player_Database.cs
```diff
@@ -18,7 +18,7 @@ partial class Player
         /// This variable is set to true when a change is made, and set to false before a save is requested.<para />
         /// The primary use for this is to trigger save on add/modify/remove of properties.
         /// </summary>
-        public bool CharacterChangesDetected { get; set; }
+        public bool CharacterChangesDetected;
```
Mag-nus: You can probably undo this change

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
```diff
@@ -18,7 +18,7 @@ partial class WorldObject
         /// This variable is set to true when a change is made, and set to false before a save is requested.<para />
         /// The primary use for this is to trigger save on add/modify/remove of properties.
         /// </summary>
-        public bool ChangesDetected { get; set; }
+        public bool ChangesDetected;
```
Mag-nus: You can probably undo this change

Source/ACE.Server/WorldObjects/Player.cs
dirtyelf: I was thinking it could be useful to have the timestamps for both directions in case of the need to throttle the requests, or debug in case of errors.

RemoveProperty() would remove the entry from the DB completely? So I would need to check for null, or just that the entry does not exist, instead of StartsWith("0")?

Removing the property from the DB completely when not in godmode would help keep things tidy.

Source/ACE.Server/WorldObjects/Player_Database.cs
dirtyelf: yep...

Source/ACE.Server/WorldObjects/WorldObject_Database.cs
dirtyelf: yep...

Source/ACE.Server/WorldObjects/Player.cs
Mag-nus: Correct, you would just check if the property was null. A null property is one that does not exist.

dirtyelf: cleaned it up and changed it to remove the property instead of set to "0=xxxx"

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+        [CommandHandler("god", AccessLevel.Sentinel, CommandHandlerFlag.RequiresWorld, 0,
+            "Turns current character into a god!",
+            "Sets attributes and skills to higher than max levels.\n"
+            + "To return to a mortal state, use the /ungod command.\n"
+            + "Use the /god command with caution. While unlikely, there is a possibility that the character that runs the command will not be able to return to normal or will end up in a state that is unrecoverable.")]
         public static void HandleGod(Session session, params string[] parameters)
         {
             // @god - Sets your own stats to a godly level.
+            // need to save stats so that we can return with /ungod
+            var biotas = new Collection<(Biota biota, ReaderWriterLockSlim rwLock)>();
```
gmriggs: This looks like it should work, and is probably actually good way to do it, just caught me off guard since it deviates from the standard session.Player.SaveBiotaToDatabase() (which doesn't have a callback)

I would have probably cheaped out here and just called session.Player.SaveBiotaToDatabase(), and then put the Ungod call in an ActionChain that runs a small amount of time later, but this is probably actually technically better

Now that I think about it though... why does the player need to be explicitly saved before entering god mode?

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+        private static void DoGodMode(bool playerSaved, Session session, bool exceptionReturn = false)
+        {
+            if (!playerSaved)
+            {
+                ChatPacket.SendServerMessage(session, "Error saving player. Godmode not available.", ChatMessageType.Broadcast);
+                Console.WriteLine($"Player {session.Player.Name} tried to enter god mode but there was an error saving player. Godmode not available.");
+                return;
+            }
+
+            Biota biota = DatabaseManager.Shard.GetBiota(session.Player.Guid.Full);
```
gmriggs: why is a full copy of the Player Biota being re-fetched here?

The current player state is already accessible in memory, session.Player

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+
+            var godString = biota.BiotaPropertiesString.FirstOrDefault(s => s.Type == (ushort)PropertyString.GodState);
```
gmriggs: should be simply session.Player.GodState

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+                        ChatPacket.SendServerMessage(session, "You are already a god.", ChatMessageType.Broadcast);
+                        return;
+                    }
+                }
+
+                string returnState = "1=";
+                returnState += $"{DateTime.UtcNow}=";
+
+                // need level, available skill credits
+                foreach (var i in biota.BiotaPropertiesInt)
```
gmriggs: seems like this should be accessible simply through session.Player.Level and session.Player.AvailableSkillCredits?

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+
+                // need total xp, unassigned xp
+                foreach (var iSixFour in biota.BiotaPropertiesInt64)
+                {
+                    returnState += $"{iSixFour.Type}=";
+                    returnState += $"{iSixFour.Value}=";
+                }
+
+                // need all attributes
+                foreach (var att in biota.BiotaPropertiesAttribute)
```
gmriggs: similar, why not just use session.Player.Attributes?

accessing a copy of the raw Biota for the player seems very odd here

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+        [CommandHandler("ungod", AccessLevel.Sentinel, CommandHandlerFlag.RequiresWorld, 0,
+            "Returns character to a mortal state.",
+            "Sets attributes and skills back to the values they were when you became a god.\n"
+            + "If the command fails to revert your state it will default to godmode.\n"
+            + "In the event of failure, contact a server administrator to sort it out.")]
+        public static void HandleUngod(Session session, params string[] parameters)
+        {
+            // @ungod - Returns skills and attributues to pre-god levels.
+            Player currentPlayer = session.Player;
+            Biota biota = DatabaseManager.Shard.GetBiota(currentPlayer.Guid.Full);
```
gmriggs: again, shouldn't be any reason to fetch a copy of the raw Biota here. session.Player already has the master reference to the raw biota if needed, and ideally session.Player gamevars should just be dealt with directly

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+            var returnString = biota.BiotaPropertiesString.FirstOrDefault(s => s.Type == (ushort)PropertyString.GodState);
```
gmriggs: session.Player.GodMode

this should be added to Player_Properites.cs, using the existing format:

```
        public string GodMode
        {
            get => GetProperty(PropertyString.GodMode);
            set { if (!value.HasValue) RemoveProperty(PropertyString.GodMode); else SetProperty(PropertyString.GodMode, value); }
        }
```

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1584,59 +1584,268 @@ public static void HandleFumble(Session session, params string[] parameters)
...
+                    if (returnStringArr.Length != 240)
+                    {
+                        Console.WriteLine($"The returnString was not set to the correct length while {currentPlayer.Name} was attempting to return to normal from godmode.");
+                        ChatPacket.SendServerMessage(session, "Error returning to mortal state, defaulting to godmode.", ChatMessageType.Broadcast);
+                        return;
+                    }
+
+                    for (int i = 2; i < returnStringArr.Length;)
+                    {
+                        switch (i)
```
gmriggs: serialization / deserializing looks good, which is the core of it, so good job with the core piece

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: I figured it was a good idea to save a snapshot to db before tinkering with all of the stats.

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: good point, no need to hit the db again

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: added GodState to player_properties.cs and altered this line

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: yeah, good idea i can grab level, sks, total exp, and available exp this way. thanks.

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: right, no need to hit db. just grabbing session.player.biota

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: strings dont have a HasValue method so I changed it to if(value == null)

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: thanks!

Source/ACE.Server/Command/Handlers/AdminCommands.cs
dirtyelf: it was in the format i needed and was easy to loop through this way. i see what you mean though. you are suggesting looping through like i do with the stats increase?
`foreach (var a in currentPlayer.Attributes)
            {
                add values to string here
            }`

how is this different that accessing the "raw" biota as you call it?

dirtyelf:
anything else this needs?

PR #2033, opened 2019-06-23
Update Healer.cs

nwacks:
Check if the healing target's attribute is at max, without assuming it's always health. Covers the rare, gauntlet, and casino Stamina/Mana kits.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Healer.cs
+++ b/Source/ACE.Server/WorldObjects/Healer.cs
@@ -57,7 +57,9 @@ namespace ACE.Server.WorldObjects
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


gmriggs:
Thanks for the PR

Someone mentioned that stamina / mana healing kits weren't working yet (these types of healing kits for other vitals were news to me), and I was going to investigate these soon

I will playtest this PR for these items

gmriggs:
ah i see some preliminary work for adding other vital types was in there, but the initial check was still always for Health

This looks like a good change, and is moving in the right direction for all of the different healing kits types.

I noticed a few other details for stamina/mana/unlimited use kits, mostly related to messages sent to the player. I can help clean these other messages up after this PR, unless anyone else would like to do that part

Looks good to me, thanks again

PR #2034, opened 2019-06-23
Fix issue with doors that had ActivationResponse on them

LtRipley36706:


Source/ACE.Server/WorldObjects/Door.cs
```diff
@@ -77,6 +77,26 @@ public string LockCode
             set { if (value == null) RemoveProperty(PropertyString.LockCode); else SetProperty(PropertyString.LockCode, value); }
         }
 
+        public override void OnActivate(WorldObject activator)
```
gmriggs: this override only includes a small subset of the logic in WorldObject_Use.OnActivate(), and does not call the base

all of the existing / very important logic in WorldObject_Use.OnActivate() suddenly gets thrown out with this change. For the logic it does include, it also duplicates from the base

the idea is for the base to do everything that needs to be done for all objects, and then for the overrides to build on top of that.

i'm not sure what all of the exact requirements are for doors yet, but it seems like this has the potential to break many other door activations, without careful review of all the data first. for example, how is the ""Beyond lays the second test" message sent to the user, is that a Talk or Emote response?

This code seems to assume that doors should only ever have an ActivationResponse.Use, and that it always executes 100% of the time. If that is the case, why not just add ActivationResponse |= ActivationResponse.Use to SetEphemeralValues?

Source/ACE.Server/WorldObjects/Door.cs
```diff
@@ -148,8 +161,9 @@ private void Reset()
                 {
                     IsLocked = true;
                     var updateProperty = new GameMessagePublicUpdatePropertyBool(this, PropertyBool.Locked, IsLocked);
-                    var sound = new GameMessageSound(Guid, Sound.OpenFailDueToLock, 1.0f); // TODO: This should probably come 1.5 seconds after the door closes so that sounds don't overlap
-                    EnqueueBroadcast(updateProperty, sound);
+                    //var sound = new GameMessageSound(Guid, Sound.OpenFailDueToLock, 1.0f); // TODO: This should probably come 1.5 seconds after the door closes so that sounds don't overlap
```
gmriggs: why commenting out the locked sound?

Source/ACE.Server/WorldObjects/WorldObject_Use.cs
```diff
@@ -123,10 +119,22 @@ public virtual void OnActivate(WorldObject activator)
...
             {
                 if (player != null)
-                    EmoteManager.OnUse(player);
+                {
+                    //target.EmoteManager.OnActivation(player); // found a few things with Activation on them but not ActivationResponse.Emote...
+                    target.EmoteManager.OnUse(player);
+                }
+                else if (activator is Creature creature)
+                {
+                    //target.EmoteManager.OnActivation(creature); // found a few things with Activation on them but not ActivationResponse.Emote...
```
gmriggs: why the duplicate logic here?
just the 2nd one is needed, would handle both players and creatures?

PR #2038, opened 2019-06-23
Generator system tweaks

LtRipley36706:
* Generator system tweaks
  - Carry over Delay from profile template correctly.
  - Update RemoveQueue info in GeneratorDump.
  - Adjust CurrentCreate to be calculated to prevent out-of-sync issues.
     - Added GeneratedTreasureItem to GeneratorProfile to track when TreasureGenerator is used (random, mutated lootgen)

### DIFF
--- a/Source/ACE.Server/Entity/GeneratorProfile.cs
+++ b/Source/ACE.Server/Entity/GeneratorProfile.cs
@@ -50,9 +50,28 @@ namespace ACE.Server.Entity
         public bool IsPlaceholder { get => Biota.WeenieClassId == 3666; }
 
         /// <summary>
+        /// TRUE if this Profile generated treasure using TreasureGenerator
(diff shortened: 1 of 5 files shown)


Source/ACE.Server/Entity/GeneratorProfile.cs
```diff
@@ -49,10 +49,29 @@ public class GeneratorProfile
         /// </summary>
         public bool IsPlaceholder { get => Biota.WeenieClassId == 3666; }
 
+        /// <summary>
+        /// TRUE if this Profile generated treasure using TreasureGenerator
+        /// </summary>
+        public bool GeneratedTreasureItem { get; private set; } = false;
```
gmriggs: false is already the default, so the = false bit is redundant

PR #2042, opened 2019-06-25
adding json import, live editing for content creators

gmriggs:


Source/ACE.Server/Managers/PropertyManager.cs
```diff
@@ -482,6 +482,7 @@ public static void LoadDefaultProperties()
 
         public static readonly ReadOnlyDictionary<string, string> DefaultStringProperties =
             DictOf(
+                ("content_folder", ".\\Content"),   // for content creators to live edit weenies. defaults to Content in the netcoreapp2.1 folder
```
Mag-nus: content_folder seems like a system property, so I think it should go in Config.js, not in the database.

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
```diff
@@ -139,5 +145,119 @@ public static List<string> OutputShortcutSQLCommand(string playerName, uint char
 
             return strings;
         }
+
+        [CommandHandler("import", AccessLevel.Developer, CommandHandlerFlag.None, 1, "Imports a weenie from the Content folder", "<wcid>")]
+        public static void HandleImport(Session session, params string[] parameters)
```
Mag-nus: Instead of DeveloperDatabaseCommands, i think this stuff should go into ContentCreationCommands

gmriggs:
config.js is for things that are required to start up the server

content_folder is not required to start up the server. it is also something that won't be used for normal server operation, only for content editing

PropertyManager also gives the user the ability to change content_folder on the fly

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
```diff
@@ -139,5 +145,119 @@ public static List<string> OutputShortcutSQLCommand(string playerName, uint char
...
+
+        public static void ImportSQL(string sqlFile)
+        {
+            var sqlCommands = File.ReadAllText(sqlFile);
+
+            using (var ctx = new WorldDbContext())
+                ctx.Database.ExecuteSqlCommand(sqlCommands);
+        }
+
+        public static void SendMessage(Session session, string msg)
```
Mag-nus: We already have a helper function for this. CommandHandlerHelper.WriteOutputX

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
```diff
@@ -139,5 +145,119 @@ public static List<string> OutputShortcutSQLCommand(string playerName, uint char
...
+
+            if (files == null || files.Length == 0)
+            {
+                SendMessage(session, $"Couldn't find {json_folder}{prefix}*.json");
+                return;
+            }
+
+            var fi = files[0];
+
+            // convert json -> sql
```
Mag-nus: You don't need this step.

You can go direct from LSD model to ACE Weenie, then add or update that weenie to the context

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
gmriggs: the structure of Command/Processors is already getting a little unruly

Handlers/ should ideally be organized by AccessLevel. If we want to further divide up a particular access level into multiple files, then it should still be prefixed with the base AccessLevel, ie. AdminCommands.cs, AdminShardCommands.cs, etc.

Currently there is no 'ContentCreator' access level (although this could be a possible idea in the future). If I am looking for all of the Developer commands, and they are split between Developer*Commands.cs and a bunch of other unrelated files, this could also be confusing

The Command folder and UI will be getting a refactoring pass very soon in general

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
gmriggs: You could do it that way, but that way is also less flexible and adaptable.

This was part of the whole discussion that was had awhile back. The determination was that the process should work as json->sql, and sql->json, as that is logically the model that most content editors have when working with the data.

The key here is that the .sql and .json files are always kept in tandem with each other. I update 1 or the other, and I now have representation of the data in both formats. I can then update the SQL to ACE-Patches, or the JSON to LSD, regardless of which native format I prefer to work in.

Obscuring the model conversion process, and doing it all invisibly behind-the-scenes with no .json or .sql files as the result is definitely not the way to go here

These could even be separated into standalone commands/executables that could be run outside of the server. ie, there could be json2sql and sql2json standalone progams to further make converting between the 2 models as easy as possible

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
gmriggs: config.js is for things that are required to start up the server

content_folder is not required to start up the server. it is also something that won't be used for normal server operation, only for content editing

PropertyManager also gives the user the ability to change content_folder on the fly

Source/ACE.Server/Managers/PropertyManager.cs
gmriggs: config.js is for things that are required to start up the server

content_folder is not required to start up the server. it is also something that won't be used for normal server operation, only for content editing

PropertyManager also gives the user the ability to change content_folder on the fly

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
Mag-nus: You don't go from .json to .sql

you go from .json -> LSD model -> ACE model -> (database or .sql)

gmriggs: which is exactly what it's doing here

.json -> LSD model -> ACE model -> (database or .sql) literally is .json to .sql. From the perspective of the user, it's converting json to sql.

The requirements here are:

 - the .json file describing the gdle model is converted into a .sql file describing the ace model
- The user has both the .json and .sql files on their system at that point, and they can view/edit both
- The updated information is added directly to the ACE DB
- The updated information is available live in-game, ie. it clears out the existing cache for that object, so that I can spawn the item with the updated data

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
Mag-nus: Access level really should be a flags enum, and grouping of commands should be based on function, not access level.

In the future, commands should support multiple access levels.

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
Mag-nus: I see that now. The function and var naming threw me off.

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
gmriggs: ah i wasn't aware of the existing function, updated to WriteOutput

Source/ACE.Server/Command/Handlers/DeveloperDatabaseCommands.cs
gmriggs: should this go into DeveloperContentCreatorCommands.cs?

gmriggs: added DeveloperContentCommands.cs

PR #2045, opened 2019-06-25
Log and Abort from SelectProfilesInit after 1000 tries

LtRipley36706:


Mag-nus:
Is there a better way to do this that doesn't involve looping anywhere between 1-1000 times every time?

PR #2054, opened 2019-06-27
fixing inventory stack bugs

gmriggs:
This PR fixes 2 types of inventory stack bugs:

Bug 1:

- Have a partial stack of an item (249/250 stack size, for example) in your main pack
- Have a partial stack of the same item in the landblock, where merging the stacks would exceed the MaxStackSize
- Have the main pack full
- Attempt to pick up the partial stack from the landblock

Result before patch:

 - Error, and the landblock stack poofs completely

Bug 2:

- Have a partial stack of a wielded item, such as arrows or phials wielded
- Have another partial stack of the same item in main pack, where merging the 2 stacks would still be < MaxStackSize
- Try to dequip the wielded stack by double clicking on it

Result before patch:

The client sends a packet to try to merge the wielded stack into the inventory stack, which resulted in an error before, from the equipped source stack not being found in inventory

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -1714,6 +1714,12 @@ namespace ACE.Server.WorldObjects
                 return;
             }
 
+            if (!CanAddToInventory(sourceStack))
+            {
+                Session.Network.EnqueueSend(new GameEventInventoryServerSaveFailed(Session, sourceStack.Guid.Full, WeenieError.None));
+                return;
+            }
+
             if ((sourceStackRootOwner == this && targetStackRootOwner != this)  || (sourceStackRootOwner != this && targetStackRootOwner == this)) // Movement is between the player and the world
             {
                 if (sourceStackRootOwner is Vendor)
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -1818,7 +1824,7 @@ private bool DoHandleActionStackableMerge(WorldObject sourceStack, Container sou
                 }
                 else if (sourceStackRootOwner != null)
                 {
-                    if (!sourceStackRootOwner.TryRemoveFromInventory(sourceStack.Guid, out _))
+                    if (!sourceStackRootOwner.TryRemoveFromInventory(sourceStack.Guid, out _) && (!(sourceStackRootOwner is Player playerContainer) || !playerContainer.TryDequipObjectWithNetworking(sourceStack.Guid, out _, DequipObjectAction.DequipToPack)))
```
Mag-nus: My only thought here is that we're doing TryRemoveFromInventory (without networking), but then TryDequipObject (WITH networking).

We have this situation often where we try both cases, but they should mostly (if not all) be either both with or both without. Stacks could be an exception though.

gmriggs: i have no idea the proper fix for either of these bugs

all i know is there are tons of inventory/stackable/full mainpack bugs, and this fixes these particular ones

gmriggs: Just verified that TryDequipObjectWithNetworking is the proper function to use here, and a regular dequip leads perma-buff vulnerabilities open

For bug 2, repro with Phials and TryDequipObjectWithBroadcasting instead, and you will be permabuffed forever.

PR #2071, opened 2019-07-04
Aetheria LootGenerationFactory plus LootGen magic code refactor

Theran578:
- Add code to generate a Coealesced Aetheria, with a drop chance of 10% that seems a reasonable amount of chance for its appearance in loot, in my personal testing


LtRipley36706:
Hold off on final merge until DB update includes a few changes to aetheria base objects, code wise checks with me however

PR #2076, opened 2019-07-04
Misc. Aetheria updates

LtRipley36706:


LtRipley36706:
more work to do on a few edge case areas

Source/ACE.Server/Entity/Aetheria.cs
```diff
@@ -133,13 +142,18 @@ public static void UseObjectOnTarget(Player player, WorldObject source, WorldObj
 
             actionChain.AddAction(player, () => ActivateSigil(player, source, target));
 
+            player.EnqueueMotion(actionChain, MotionCommand.Ready);
```
gmriggs: player.IsBusy = false should be after the return to ready state animation

IsBusy=true/false should also be kept together in the same function for consistency and clarity

LtRipley36706:
think this is in a good state now for aetheria improving support

Source/ACE.Server/WorldObjects/Hotspot.cs
```diff
@@ -152,32 +152,35 @@ private void Activate(Player player)
             var amount = DamageNext;
             var iAmount = (int)Math.Round(amount);
 
+            var pAmount = 0;
```
gmriggs: are 3 amounts really needed here?

the whole point of iAmount was to track the integer amount through the pipeline, for the final output message

LtRipley36706: iAmount wasn't changing on 3 of the 4, i can make it do that of course, but at the same time i didn't see the harm either way

Source/ACE.Server/WorldObjects/Hotspot.cs
```diff
@@ -152,32 +152,35 @@ private void Activate(Player player)
...
+
             switch (DamageType)
             {
                 default:
                     if (player.Invincible) return;
                     amount *= (float)player.GetLifeResistance(DamageType);
                     iAmount = (int)Math.Round(amount);
                     player.TakeDamage(this, DamageType, amount, Server.Entity.BodyPart.Foot);
+                    pAmount = Math.Abs(iAmount);
```
gmriggs: iAmount is always positive here, so this statement is effectively pAmount = iAmount, ie. redundant

Source/ACE.Server/WorldObjects/Hotspot.cs
```diff
@@ -152,32 +152,35 @@ private void Activate(Player player)
...
-                    player.UpdateVitalDelta(player.Health, -iAmount);
+                    pAmount = player.UpdateVitalDelta(player.Health, -iAmount);
                     break;
             }
 
             if (!Visibility)
                 EnqueueBroadcast(new GameMessageSound(Guid, Sound.TriggerActivated, 1.0f));
 
-            if (!string.IsNullOrWhiteSpace(ActivationTalk))
+            if (!string.IsNullOrWhiteSpace(ActivationTalk) && pAmount != 0)
```
gmriggs: all of this code can be nuked, an replaced with 1 update: && iAmount != 0

PR #2078, opened 2019-07-06
ObjectDescriptionFlags cleanup

gmriggs:


Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -1121,6 +1121,12 @@ public WieldRequirement WieldRequirements4
             set { if (!value.HasValue) RemoveProperty(PropertyInt.MaxStackSize); else SetProperty(PropertyInt.MaxStackSize, value.Value); }
         }
 
+        public bool IsSellable
```
Mag-nus: A property like this with a default that could go either way should describe the default via a summary.

I'd prefer a property like this not return a default.

gmriggs: This stems from a misunderstanding of the data and the system. The property *can't* go either way; as defined by the AC data, if an item doesn't have the IsSellable property, then it is sellable.

"I'd prefer a property like this not return a default."

That's fine, but you have to realize that this view does not match the reality of the system. GetProperty() is still there if you want to access the property without returning a default.

Can you give an example of where you'd have 3 separate branches depending on if this property is null/false/true, or an example of where you'd have varying defaults for this property?

For every property that has been given the correct default, it has been proven: that is indeed the default used for that property as per the data, and that every use of that property all throughout the server code previously had the same Property ?? Default code scattered throughout the system.

This normalizes those defaults into 1 location, to provide a clear indication of what the default for that field actually is, which eliminates redundant code and the potential for errors.

Mag-nus: "This stems from a misunderstanding of the data and the system. The property can't go either way; as defined by the AC data, if an item doesn't have the IsSellable property, then it is sellable."

If you learn these things, and want others to understand the system as you do, then you should comment these findings with summaries on both the Property enum as well as the property wrapper.

A consumer of the property shouldn't have to F12 or search through commit history to find a statement such as this.

Summaries make it clear to the consumers the function and reasoning of a property, especially in cases where the default of a property isn't the same as the native default (false for bool, 0 for int, etc..)

This type of documentation is very easy to do and helps future developers and content creators understand what all states of the property mean. In the case of this bool, null and true = true, false = false.

LtRipley36706:
I did test this mostly via the Obj Maint 3.0 pr, but should still be valid

PR #2084, opened 2019-07-07
Aetheria LootGenerationFactory plus LootGen magic code refactor

Theran578:
- Add code to generate Coealesced Aetheria, with a drop chance of 1%, with a server option to mod the drop rate
- Refactor LootGen magic code to consolidate redundant code
- Current RNG chances for cantrips in master maintained but with a server option to mod the drop rates
- Default of 1.0 for server option mods set drop rates to coded defaults; values greater than one decrease the drop rate and values less than one increase the drop rate.


Theran578:
As mentioned in the first PR of the same name, will hold off on merging, if approved before Aetheria Quest content is added, as it is probably wouldn't be a good idea to have Coealesced Aetheria drop for players without a means to use them.

LtRipley36706:
The other factor for merging before DB is updated is that the current "Coealesced Aetheria" dropping would have the name of "Aetheria" when looted, and makes it a bit more difficult to distinguish it needs be used with the Aetheria Mana Stone. As coded now, only Coealesced Aetheria can be "bathed in mana" so once the Aethera quest db is merged, later this week, we can bring this in to go hand-in-hand with. :) Thanks for adding it to loot for us! =)

PR #2085, opened 2019-07-07
Add support for version command response

LtRipley36706:


Mag-nus:
Is this a table that just holds one row that represents the version?

It might make more sense to create a Property table that can expose the master version as a record, but also allow additional properties to be defined in the future.

For the world version, it might make more sense to store that as a Build Date instead of a Build Version, so you might have a PropertiesTimestamp table., or you could put it in a PropertiesString table, similar to how we do it in the Shard db.

PR #2099, opened 2019-07-12
Revert "Only track Player biotas (#1813)"

Mag-nus:
This reverts commit ce9791703b7acbaeea433bb1cbadff6e3cbf17e0.

Don't merge just yet, I want to review it a little more thoroughly as well as add notes.


Mag-nus:
Ok, this is correct and is as it was before we removed EF tracking for non-player based biotas.

This re-enables EF tracking for all biotas.
Get/Delete performance is no different.
Save performance is substantially improved. By using EF tracking (as we were), it could take 10% - 50% as much time as it was without the tracking.

With the tracking, every time a non-player biota was saved, the entire entity would have to be pulled from the database, diffed, and then saved.

This simply saves the changes on save without having to pull and diff.

PR #2103, opened 2019-07-13
WIP: Multi-thread landblock ticking (Landblock Groups)

Mag-nus:
This is the start of "landblock group" ticking.

The idea behind the landblock groups are that each group may contain multiple landblocks that must be ticked on the same thread, but, each group itself can be ticked on independent threads.

Landblock groups describe their area using a rectangle. As landblocks are added (and removed) the rectangle that contains all the landblocks in the group is adjusted.

Landblocks are added to groups based on a minimum distance to the landblock groups rectangle.

In the event a landblock needs to be added and is within range of multiple groups, those groups will be combined into one.

Periodically, landblock groups are checked to see if they can be split. When a landblock group is split, it can result in a new group that has an overlapping rectangle to the parent group. This is not a problem because all the landblocks within each group meet the minimum distance requirements. In the event that a new landblock is added and is close to both rectangles, the groups will then be merged again.

Adding landblocks to groups is a very efficient process
Removing landblocks from groups is a very efficient process
Checking landblock groups for split potential does incur some overhead (~0.5 ms) which is why it's only done on intervals that are twice the Landblock.UnloadInterval.  
  
  
Changes we've had to made for thread safety
- ACE.Server/Network/Sequence/SequenceManager


Mag-nus:
One issue I can think of would be fellowships. Fellowships can contain entities that may exist across thread boundaries. We'll have to look at thread safety for relationships like these, and others.

Mag-nus:
On my test setup, this reduces UpdateGameWorld_LandblockManager_Tick from over 50ms to under 10ms.

Mag-nus:
Replaced by 2303

PR #2112, opened 2019-07-15
Update Creature Combat Skills for MoA

Theran578:
- Update Creature combat skills for MoA, authored by @gmriggs


Theran578:
Tested and works with the inclusion of the null check

PR #2137, opened 2019-07-20
Save Item to DB in TryCreateInInventoryWithNetworking

LtRipley36706:
Attempt at preventing item rollback on crashes

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -105,6 +105,8 @@ namespace ACE.Server.WorldObjects
             if (item.WeenieType == WeenieType.Coin)
                 UpdateCoinValue();
 
+            item.SaveBiotaToDatabase();
+
             return true;
         }
 


Mag-nus:
Please make this server configurable, default to off.

LtRipley36706:
CE running this has no impact on perf.

Mag-nus:
Both the rx and tx saves should be conditional based on the bool property gmriggs mentioned.

That will keep the system 1:1 in both configurations and still optional.

LtRipley36706:
I actually disagree with the optional nature you propose. The purpose of the save is to prevent issues relating to crashes which can't always be prevented/anticipated, where as throughout the code SaveBiotaToDatabase is used to force saves to ensure states are preserved, and not optionally, just for such occasions. 

player_receive_immediate_save was added at some point and actually is more expensive because its doing a full player save instead of just the item saves needed in current use cases. The PR code actually negates the need to save entire players because on both outgoing and incoming items, the item with significantly less properties/dictionaries is saved vs the three instances of this code:
```cs
if (PropertyManager.GetBool("player_receive_immediate_save").Item)
                                player.RushNextPlayerSave(5);
```

Mag-nus:
A server operator should be able to chose if they want to incur the additional overhead with ACID-like behavior, or, are ok with a rollback of approx 0-5m (which is similar to how retail operated)

LtRipley36706:
this isn't a rollback. The item is lost to the void because of the nature of saves. It saved in the item with null owner/container because that save is "required" and then subsequently lost because the update to add owner/container is no longer required? a true rollback is better because the previous full state is preserved, given that isn't the case here, you can't say its an acceptable outcome

Mag-nus:
The concern here is player to player/chest item transfers right? If so, we can solve this in a much more database-efficient manner.

LtRipley36706:
concern here is player to player, player to npc, npc to player, player to container, container to player

Mag-nus:
If you comment out the save on RemoveItemFromContainer, what issues remain that need to be solved?

Here is the reason why I originally added that:

Notes:
Landblocks are each on their own independent 5m save intervals
Players all have their own independent 5m save intervals

Bob is saved
Bob gives sword to Jim
Bob is saved
Server crashes, Jim was not saved.

The sword is still recorded as being possessed by Bob. When Bob relogs, what happens if his inventory is full? Any excess items should be moved into any available slot, or, dropped on the ground. I don't believe this logic is implemented.

Currently, as implemented, if items are being saved without containers while they are being transitioned between containers, that's a problem that we should fix.

We shouldn't then be saving the item again after it's received by the container to bring the item back out of the void.

PR #2140, opened 2019-07-20
Add Curse of Raven Fury spell

Slushnas:
This will likely need more testing to verify everything behaves like retail but from my own tests everything appears to work properly.

### DIFF
--- a/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
+++ b/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
@@ -1750,7 +1750,7 @@ namespace ACE.Server.WorldObjects
 
         public int? EncumbranceVal
         {
-            get => GetProperty(PropertyInt.EncumbranceVal) ?? 0;
+            get => GetProperty(PropertyInt.EncumbranceVal);
             set { if (!value.HasValue) RemoveProperty(PropertyInt.EncumbranceVal); else SetProperty(PropertyInt.EncumbranceVal, value.Value); }
         }
 
--- a/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
+++ b/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
@@ -1750,7 +1750,7 @@ namespace ACE.Server.WorldObjects
 
         public int? EncumbranceVal
         {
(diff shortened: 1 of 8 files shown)


Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -1750,7 +1750,7 @@ public string Inscription
 
         public int? EncumbranceVal
         {
-            get => GetProperty(PropertyInt.EncumbranceVal);
+            get => GetProperty(PropertyInt.EncumbranceVal) ?? 0;
```
Mag-nus: I'm not sure if this will cause any issues throughout the code, so I am not commenting on that.

If you give a property a default, please also give it a /// <summary> describing that default. Also, a property that exposes a default probably shouldn't be nullable.

gmriggs: yeah this is not the correct way to change properties, the entire codebase needs to be audited for everywhere that uses EncumbranceVal

i would recommend modifying this in the place where it needs to be set atm, until the rest of the codebase can be reviewed for it

Slushnas: Yes, I wasn't sure what the best way to change this would be. I noticed some other properties in the file were set the exact same way (BoostValue, UnlimitedUse etc.) so I stuck with that convention.

I believe you can propose changes inline now in Github which I could accept if you want to do that or let me know what to change it to to address the concern.

gmriggs: the difference is things like BoostValue and UnlimitedUse return non-nullable types. the process for this has been to review the entire codebase, and see that everywhere EncumbranceVal is used all throughout the existing gameplay code, it is redundantly setting the default value of '0' all throughout every place in the existing code.

from doing a quick search of the code last night, i can see that this was not the case - there was some code in RecipeManager / tinkering that used a default EncumbanceVal of 1 when modifying the weight of an item.

the purpose of this change is to specify the defaults in 1 location, instead of the same defaults specified multiple times redundantly everywhere they are used in the code

the process also ensures that there is no existing code that makes a differentiation between null and the default value.

PR #2143, opened 2019-07-21
Command to list runtime configuration properties

dgarson:
This PR seeks to improve the way properties are set and the visibility of the different properties.

First is the addition of the "showprops" command which enumerates the different properties (in order by their data type) and then shows both their name, description and default value.

Now validation is also done on values passed in when setting an option, ensuring that the user knows whether the value was successfully set, or whether no setting existed with the given key. It's worth noting that it will also complain a key does not exist if it exists but has a different data type (this could be improved if necessary).


gmriggs:
tested, seems to be a good addition

one suggestion, on the /showprops screen, can it show the current value for that property?

PR #2144, opened 2019-07-21
Include material in give item message

dgarson:
Right now whenever you give an item to a player or an NPC, it omits the material name prefix. This material name prefix should be present at all times when passing around or tinkering items with a material type. This PR adjusts the text to include the material name, but ensures that the item is neither stackable (which should never have a valid material) and that it has a material name specified, prior to including it in the output.


gmriggs:
looks like this should be good. since material+item name might be more of a common thing outside of just RecipeManager, eventually perhaps there should be a WorldObject.SomethingName that automatically builds the correct material+name. RecipeManager.GetMaterialName() was a start at this, but still needs additional logic layered on top, ie. to handle when there is no material name. All of this probably should be wrapped into 1 consistent function eventually

Issue #1621, opened 2019-04-01
Configurable Option For Character Save Time


cyberkiller6:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Desired functionality.
Would like to request a configurable option for character save time. Allow server operators to specify/change the default 5 minute character save time to x amount of minutes they choose.

dgarson:
I took a stab at this due to it being inactive for a long time.

dgarson:
I think this ticket can be closed.

Mag-nus:
I'm closing this.

Much of ACE is balanced around certain conditions, such as the current 5 minute defined save interval. Misunderstanding the gravity of the property, or accidentally changing it, could result in drastic performance and/or data corruption issues.

If such an adjustment is required, it should likely be done by the server operator in their own local code repo.

PR #2146, opened 2019-07-21
Added configurable 'player_save_interval' seconds property

dgarson:
Adds a new configuration property for the number of minutes between the automatic save of player characters. This is in response to the the feature request in https://github.com/ACEmulator/ACE/issues/1621

### DIFF
--- a/Source/ACE.Server/Managers/PlayerManager.cs
+++ b/Source/ACE.Server/Managers/PlayerManager.cs
@@ -27,17 +27,10 @@ namespace ACE.Server.Managers
         private static readonly Dictionary<uint, Player> onlinePlayers = new Dictionary<uint, Player>();
         private static readonly Dictionary<uint, OfflinePlayer> offlinePlayers = new Dictionary<uint, OfflinePlayer>();
 
-        public static readonly long DefaultPlayerSaveIntervalSecs = 300; // default to 5 minutes
-
         /// <summary>
         /// OfflinePlayers will be saved to the database every 1 hour
         /// </summary>
         private static readonly TimeSpan databaseSaveInterval = TimeSpan.FromHours(1);
-
-        /// <summary>
-        /// The timespan between automatic flushing of character changes to the database
-        /// </summary>
-        public static TimeSpan PlayerSaveInterval = TimeSpan.FromSeconds(DefaultPlayerSaveIntervalSecs); // default to 5 until initialized from PropertyManager
 
         private static DateTime lastDatabaseSave = DateTime.MinValue;
 
@@ -53,8 +46,6 @@ namespace ACE.Server.Managers
                 var offlinePlayer = new OfflinePlayer(result);
                 offlinePlayers[offlinePlayer.Guid.Full] = offlinePlayer;
             }
-
-            UpdateAfterPropertyChanges();
         }
 
         public static void Tick()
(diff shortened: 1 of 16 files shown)


Source/ACE.Server/Managers/PlayerManager.cs
```diff
@@ -27,11 +27,18 @@ public static class PlayerManager
         private static readonly Dictionary<uint, Player> onlinePlayers = new Dictionary<uint, Player>();
         private static readonly Dictionary<uint, OfflinePlayer> offlinePlayers = new Dictionary<uint, OfflinePlayer>();
 
+        public static readonly long DefaultPlayerSaveInterval = 5;
```
Mag-nus: When I was thinking about this, I was thinking the unit of measurement would be seconds.

Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -47,6 +47,11 @@ public partial class Player : Creature, IPlayer
 
         public float CurrentRadarRange => Location.Indoors ? 25.0f : 75.0f;
 
+        public TimeSpan PlayerSaveInterval
```
Mag-nus: This can go back in Player_Database.cs

Source/ACE.Server/Managers/PropertyManager.cs
```diff
@@ -469,8 +475,9 @@ public static void LoadDefaultProperties()
                 ("char_delete_time", 3600),         // the amount of time in seconds a deleted character can be restored
                 ("mansion_min_rank", 6),            // overrides the default allegiance rank required to own a mansion
                 ("max_chars_per_account", 11),      // retail defaults to 11, client supports up to 20
-                ("pk_timer", 20)                    // the number of seconds where a player cannot perform certain actions (ie. teleporting)
+                ("pk_timer", 20),                   // the number of seconds where a player cannot perform certain actions (ie. teleporting)
                                                     // after becoming involved in a PK battle
+                ("player_save_interval", 5)         // number of minutes between saving of characters
```
Mag-nus: Seconds make sense here. It's the unit of measurement we use in the majority (all?) of cases in ACE

Source/ACE.Server/Managers/PlayerManager.cs
dgarson: I'll add a suffix indicating it's minutes.

Source/ACE.Server/WorldObjects/Player.cs
dgarson: Woops. This isn't supposed to be in here anymore. Thanks for catching that.

Source/ACE.Server/Managers/PropertyManager.cs
dgarson: I was just using the existing units which was minutes. Seconds works too. I'll switch it over.

Source/ACE.Server/WorldObjects/Player.cs
dgarson: I remember what happened here. It says `PlayerManager` is not in scope for Player_Database

Mag-nus:
Thanks for the seconds change.

My only other thought is that PlayerManager should not be involved.

I think you can just have PlayerSaveInterval in Player_Dabase.cs point to => PropertyManager.GetLong("player_save_interval", DefaultPlayerSaveIntervalSecs).Item;

You can change PlayerSaveInterval to an int so you're not converting every request.

If we want to capture changes from Propertymanager, it should raise an event that also includes the property being changed. Then, objects could subscribe to that event to determine if they need to make an adjustment.

dgarson:
I found that PropertyManager was also inaccessible from Player_Database either. Any other ideas?

I agree about the PropertyManager change callback but that felt outside the scope of this PR.

Theran578:
Adding 'using ACE.Server.Managers;' should make PropertyManager "accessible" from Player_Database.cs.  As far as I can see, the relevant API functions of PropertyManager are all set to public.

dgarson:
@Mag-nus any more changes I should make to this PR before it's ready to merge?

PR #2147, opened 2019-07-21
Unattackable until exited from portal space

dgarson:
Right now it is possible to attack players, and to gain aggro on mobs, while you are still in portal space. Retail behavior was that you were considered unattackable while in portal space, meaning nothing would aggro you and all attempts to attack would fail to do any damage.

This PR adds a check for the Teleporting flag in a few places where we determine whether a target can be attacked, namely a player, since this is not applicable for mobs.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Creature_Melee.cs
+++ b/Source/ACE.Server/WorldObjects/Creature_Melee.cs
@@ -135,7 +135,7 @@ namespace ACE.Server.WorldObjects
                 if (player != null && creature is Player && player.CheckPKStatusVsTarget(player, creature, null) != null)
                     continue;
 
-                if (!creature.Attackable)
+                if (!creature.Attackable || creature.Teleporting)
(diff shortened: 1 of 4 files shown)


dgarson:
There's a chance this is already working as expected. I have to verify.

gmriggs:
This should be working in most places, but it looks like you found 2 missing places:

- Cleaving damage
- There seems to be a bug where sometimes monsters will aggro players while they are still in portal space

gmriggs:
this looks like a good addition, some other notes:

- not sure if ace's concept of 'IsTeleporting' was a thing in retail, or if really the code that determines if something is attackable should be going by the .Hidden (pink bubbles) state

- the checks for 'IsTeleporting' were recently added to player attacks such as melee/missile, but there still might be some gaps for the monsters... if something is in Hidden state, the monsters shouldn't really be trying to even attack it

PR #2150, opened 2019-07-22
 Add full support for Contracts

LtRipley36706:


LtRipley36706:
Of Note, all the Virtual changes came from scaffolding on the latest version of dotnet

Source/ACE.Server/Managers/ContractManager.cs
```diff
@@ -0,0 +1,283 @@
...
+    {
+        public static void Write(this BinaryWriter writer, ContractManager contractManager)
+        {
+            writer.Write(contractManager.ContractTrackerTable);
+        }
+
+        public static void Write(this BinaryWriter writer, Dictionary<uint, ContractTracker> contractTrackerHash)
+        {
+            PackableHashTable.WriteHeader(writer, contractTrackerHash.Count);
+            foreach (var kvp in contractTrackerHash)
```
gmriggs: this table needs to be sorted by key % 32

(confirmed it seems to always use 32 tablesize in the pcaps)

PR #2156, opened 2019-07-23
Add busy checks for recall/PKL commands

dgarson:
Right now it is possible to use multiple recall commands as well as the PKL command simultaneously. You can test on the master branch by typing /mp and then immediately typing /pkl. You will notice that you become PKL just before you teleport to marketplace. You can use these commands while teleporting in portal space which will begin the timer for the action concurrently with the current teleport/action.

This PR adds a common check to determine whether a recall (or entering PKL) is possible at that time. It checks for whether the character is busy or is already teleporting.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -895,7 +895,8 @@ namespace ACE.Server.WorldObjects
                 Session.Network.EnqueueSend(new GameEventWeenieError(Session, WeenieError.OnlyNonPKsMayEnterPKLite));
                 return;
             }
-            if (IsBusy || EmoteManager.IsBusy)
+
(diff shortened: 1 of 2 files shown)


dgarson:
So... for some reason I've noticed that I don't get the "You're too busy" weenie error, but it *does* work in terms of preventing the concurrent recall/pkl commands. Any idea as to why?

Theran578:
There is already precedent set for checking if (IsBusy || Teleporting) throughout the existing codebase that I don't think a new getter to encapsulate this check is really necessary

gmriggs:
> So... for some reason I've noticed that I don't get the "You're too busy" weenie error, but it _does_ work in terms of preventing the concurrent recall/pkl commands. Any idea as to why?

Just tested, I get the "You're too busy!" error

PR #2160, opened 2019-07-25
fixing transferring items from side packs -> main pack at max burden

gmriggs:


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -910,7 +910,9 @@ private bool DoHandleActionPutItemInContainer(WorldObject item, Container itemRo
                 }
             }
 
-            if (!container.TryAddToInventory(item, placement, true))
+            var burdenCheck = itemRootOwner != this && containerRootOwner == this;
```
Mag-nus: I think this should be:

var burdenCheck = itemRootOwner != this || containerRootOwner != this;

That way the code still does burden checks on the target container for items that exit the player.

PR #2168, opened 2019-07-27
RestrictionDB null prevention

gmriggs:


Source/ACE.Server/Network/Structure/RestrictionDB.cs
```diff
@@ -46,7 +46,7 @@ public RestrictionDB(House house)
 
             if (owner == null)
             {
-                Console.WriteLine($"RestrictionDB({house.HouseInstance.Value}): couldn't find house owner {house.HouseOwner.Value}");
+                Console.WriteLine($"RestrictionDB({house.HouseInstance:X8}): couldn't find house owner {house.HouseOwner:X8}");
```
Mag-nus: Please use Log instead of console so these messages/warnings/etc aren't lost.

gmriggs: These messages shouldn't occur under any normal circumstances

While I generally agree, the current logging classes are incredibly inconvenient to use. There are 100s of references to them, and for this class it would need to: add a using, add the 101st reference to log.

The logging interface needs to be fixed, which is a much bigger issue.

PR #2193, opened 2019-08-05
Tailoring

OptimShi:
Some tailoring cleanup. Verified all the properties and text from the (limited number) of pcaps and added in some support to try to clean up potential IconOverlay issues. Also removed the ability to reduce chest armor to places they're not supposed to go (e.g. anywhere but chest)


Source/ACE.Database/ShardDatabase.cs
```diff
@@ -702,7 +702,6 @@ public List<Character> GetCharacters(uint accountId, bool includeDeleted)
                 .Include(r => r.CharacterPropertiesQuestRegistry)
                 .Include(r => r.CharacterPropertiesShortcutBar)
                 .Include(r => r.CharacterPropertiesSpellBar)
-                .Include(r => r.CharacterPropertiesSquelch)
```
gmriggs: bad merge?

OptimShi: Yes, thank you... Will try to clear that up here.

PR #2199, opened 2019-08-06
[WIP] fix olthoi armor AL for existing shard

gmriggs:


gmriggs:
should not be merged until all base #s and bonus RNG has been confirmed

gmriggs:
this patch still is not correct. the base AL for olthoi armor (and many other armors on the server) are still incorrect, and the loot generator functions which apply AL modifiers is using an additive method, whereas the data from retail seems to be using a multiplier

PR #2204, opened 2019-08-10
adding SyncLocation for enter_world()

gmriggs:
This fixes the bug with scatter generator position broadcasting


Theran578:
I can confirm this fixes the shard scatter issue

PR #2211, opened 2019-08-11
Fix issue with spell book probabilities

LtRipley36706:


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -291,11 +291,10 @@ public bool TryEquipObjectWithNetworking(WorldObject item, EquipMask wieldedLoca
                     if (item.HasProcSpell((uint)spell.Spell))
                         continue;
 
-                    if (spell.Probability == 0.0f)
```
gmriggs: not how this code should work, this breaks existing functionality

Source/ACE.Server/WorldObjects/WorldObject.cs
```diff
@@ -246,6 +246,12 @@ private void InitializePropertyDictionaries()
 
             foreach (var x in Biota.BiotaPropertiesSpellBook)
                 BiotaPropertySpells[x.Spell] = x;
+
+            // Remove doubled up proc/use spells from spell book (incorrect data)
```
gmriggs: this should not be done at runtime. the purpose of the above code is to detect those situations so that they can be fixed in the correct place in the db

Source/ACE.Server/WorldObjects/Player_Inventory.cs
gmriggs: this does not fix the issue, as it's still removing the continue

Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -291,13 +291,26 @@ public bool TryEquipObjectWithNetworking(WorldObject item, EquipMask wieldedLoca
...
+                    if (ProcSpell.HasValue && spell.Spell == ProcSpell.Value)
```
gmriggs: this is checking ProcSpell / SpellDID on the player, instead of the item

should just be

```
if (spell.Spell == item.SpellDID || spell.Spell == item.ProcSpell)
    continue;
```

the logging is unneeded, as we now have a query that finds all of these cases across the entire db

PR #2216, opened 2019-08-11
Revert "prototype: GfxObjCache + WeakReference (#1692)"

Mag-nus:
This reverts commit 509ebd6148fb71ef4d884333924eb1d0a7284ef1.

Not needed anymore.

### DIFF
--- a/Source/ACE.Server/Physics/Entity/GfxObjCache.cs
+++ b/Source/ACE.Server/Physics/Entity/GfxObjCache.cs
@@ -1,3 +1,6 @@
+// Uncomment this if you want to use weak reference cache. It could save 10's of MB and might be useful on a micro-server
+//#define USE_WEAK_REFERENCE_CACHE
+
 using System;
 using System.Collections.Concurrent;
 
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Physics/Entity/GfxObjCache.cs
```diff
@@ -8,8 +8,7 @@ namespace ACE.Server.Physics.Entity
 {
     public static class GfxObjCache
     {
-        //public static readonly ConcurrentDictionary<uint, GfxObj> GfxObjs = new ConcurrentDictionary<uint, GfxObj>();
-        public static readonly ConcurrentDictionary<uint, WeakReference<GfxObj>> GfxObjs = new ConcurrentDictionary<uint, WeakReference<GfxObj>>();
+        public static readonly ConcurrentDictionary<uint, GfxObj> GfxObjs = new ConcurrentDictionary<uint, GfxObj>();
```
gmriggs: the caches are designed to ensure there are not multiple instances to the same resource (similar to DatLoader)

however, not similar to DatLoader, they are not intended to keep resources permanently loaded throughout the lifetime of the server. if i am running a smaller server for multiple hours, i will most likely only have a small subset of these resources loaded at any point. i do not want my computer to continually permaload more resources as it comes across them.

the physics caches do not work like the DatLoader in this way, or even the Weenie cache. if there should be a similar option to the Weenie cache to preload all of these resources on startup, and keep them permaloaded throughout the lifetime of the server, then that should be a configurable option. (a lot of the larger servers also run with weenie preloading disabled)

Mag-nus: I can make this optional, but when we tested this, it was tens of MB's we're talking about.

If you want it configurable for 10's of MB's, let me know and I'll make the change

gmriggs: yeah for GfxObjCache in particular, might not add up to much... was making more of a general statement of how the physics entity caches differ from other things like DatLoader / WeenieCache

PR #2217, opened 2019-08-11
Cleanup log.Info level messages

Mag-nus:
Log.info is the default console output and is intended for:
server startup
connection/disconnections
admin initiated command output
server shutdown

Log.Debug is the appropriate level for debug-type messages that are to be logged and audited at a later date, but not output to the console.

To filter specific messages to a separate file for easy auditing, see log4net.config.example


Source/ACE.Server/Managers/HouseManager.cs
```diff
@@ -405,7 +405,7 @@ private static void HandleRentPaid(PlayerHouse playerHouse)
 
             slumlord.SaveBiotaToDatabase();
 
-            log.Info($"HouseManager.HandleRentPaid({playerHouse.PlayerName}): rent payment successful!");
+            log.Debug($"HouseManager.HandleRentPaid({playerHouse.PlayerName}): rent payment successful!");
```
gmriggs: not debug messages.

Source/ACE.Server/WorldObjects/Corpse.cs
```diff
@@ -225,8 +225,8 @@ public void GenerateRare(WorldObject killer)
             var tier = LootGenerationFactory.GetRareTier(wo.WeenieClassId);
             LootGenerationFactory.RareChances.TryGetValue(tier, out var chance);
 
-            log.Info($"[RARE] {Name} ({Guid}) generated rare {wo.Name} ({wo.Guid}) for {killer.Name} ({killer.Guid})");
-            log.Info($"[RARE] Tier {tier} -- 1 / {chance:N0} chance");
+            log.Debug($"[RARE] {Name} ({Guid}) generated rare {wo.Name} ({wo.Guid}) for {killer.Name} ({killer.Guid})");
```
gmriggs: most definitely not debug messages.

Source/ACE.Server/WorldObjects/Creature_Death.cs
```diff
@@ -361,10 +361,10 @@ public void DoCantripLogging(WorldObject killer, WorldObject wo)
             var legendaryCantrips = wo.LegendaryCantrips;
 
             if (epicCantrips.Count > 0)
-                log.Info($"[EPIC] {Name} ({Guid}) generated item with {epicCantrips.Count} epic{(epicCantrips.Count > 1 ? "s" : "")} - {wo.Name} ({wo.Guid}) - {GetSpellList(epicCantrips)} - killed by {killer.Name} ({killer.Guid})");
+                log.Debug($"[EPIC] {Name} ({Guid}) generated item with {epicCantrips.Count} epic{(epicCantrips.Count > 1 ? "s" : "")} - {wo.Name} ({wo.Guid}) - {GetSpellList(epicCantrips)} - killed by {killer.Name} ({killer.Guid})");
```
gmriggs: most definitely not debug messages.

Mag-nus: Don't get caught up in the term "debug".

What you should really be asking is, is this a message that should go to the console output during normal server operation?

In ACE, the way we define that threshold is between Debug and Info log levels. Info goes to the console, Debug does not.

Source/ACE.Server/Managers/HouseManager.cs
Mag-nus: Is this a message that should be printed to the console/server manager when running a server?

I don't think sever operators want to see a live stream of rent being paid.

Source/ACE.Server/WorldObjects/Corpse.cs
Mag-nus: Sever operators may want to see this in their console output, we should ask.

Mag-nus:
I'll add prefixes and update the log4net.config.exmaple

Source/ACE.Server/WorldObjects/Creature_Death.cs
gmriggs: in that case, yes... if the server admin chooses to enable epic info, or any particular info such as corpse logging, then all that info should be displayed on the server console

in fact, i'm not even sure i understand why LogLevel.Debug wouldn't be displayed on the server console. i think of Debug as something only devs would enable mostly, and this would show information such as 'player is moving from landblock XXYY to another landblock XXYY' or things that a live server would never enable under normal circumstances. and even then, if i were developing w/ LogLevel.Debug, i would expect to see those messages on the console. i would expect the console to just display any log messages that i have enabled, including debug

Source/ACE.Server/Managers/HouseManager.cs
gmriggs: sure, if they want to see a stream of events of certain types that describe what is happening on the server, it's very likely important Housing events such as this would be something they want to see

placing messages into a handful of tiered buckets (if you want to see housing messages, you have to enable debug logging, and everything that comes with it, and then filter it to the messages you DO want to see) seems like it might not be the best system to me. deciding which of 2 buckets messages should go into, or which buckets go to the console, is completely arbitrary. Log(Type, Message) where Type is a Flags enum with all the different messages types (housing, corpse, epics, legendaries, rares) that the server admin can selectively choose to enable/disable independently

gmriggs: not debug messages

Source/ACE.Server/WorldObjects/Corpse.cs
```diff
@@ -67,7 +67,7 @@ protected override void OnInitialInventoryLoadCompleted()
...
+                log.Debug($"[CORPSE] {Name} (0x{Guid.ToString()}) Reloaded from Database: Corpse Level: {Level ?? 0} | InventoryLoaded: {InventoryLoaded} | Inventory.Count: {Inventory.Count} | TimeToRot: {TimeToRot} | CreationTimestamp: {CreationTimestamp} ({Time.GetDateTimeFromTimestamp(CreationTimestamp ?? 0).ToLocalTime().ToString("yyyy-MM-dd HH:mm:ss")}) | Corpse should not decay before: {dtTimeToRot.ToLocalTime().ToString("yyyy-MM-dd HH:mm:ss")}, {tsDecay.ToString("%d")} day(s), {tsDecay.ToString("%h")} hours, {tsDecay.ToString("%m")} minutes, and {tsDecay.ToString("%s")} seconds from now.");
```
gmriggs: not debug messages

Source/ACE.Server/WorldObjects/Corpse.cs
```diff
@@ -225,8 +225,8 @@ public void GenerateRare(WorldObject killer)
             var tier = LootGenerationFactory.GetRareTier(wo.WeenieClassId);
             LootGenerationFactory.RareChances.TryGetValue(tier, out var chance);
 
-            log.Info($"[RARE] {Name} ({Guid}) generated rare {wo.Name} ({wo.Guid}) for {killer.Name} ({killer.Guid})");
-            log.Info($"[RARE] Tier {tier} -- 1 / {chance:N0} chance");
+            log.Debug($"[LOOT][RARE] {Name} ({Guid}) generated rare {wo.Name} ({wo.Guid}) for {killer.Name} ({killer.Guid})");
```
gmriggs: definite not debug messages, should not have multiple categories

Source/ACE.Server/WorldObjects/Creature_Death.cs
```diff
@@ -367,10 +367,10 @@ public void DoCantripLogging(WorldObject killer, WorldObject wo)
             var legendaryCantrips = wo.LegendaryCantrips;
 
             if (epicCantrips.Count > 0)
-                log.Info($"[EPIC] {Name} ({Guid}) generated item with {epicCantrips.Count} epic{(epicCantrips.Count > 1 ? "s" : "")} - {wo.Name} ({wo.Guid}) - {GetSpellList(epicCantrips)} - killed by {killer?.Name} ({killer?.Guid})");
+                log.Debug($"[LOOT][EPIC] {Name} ({Guid}) generated item with {epicCantrips.Count} epic{(epicCantrips.Count > 1 ? "s" : "")} - {wo.Name} ({wo.Guid}) - {GetSpellList(epicCantrips)} - killed by {killer?.Name} ({killer?.Guid})");
```
gmriggs: definitely not debug messages, should not have multiple categories

Source/ACE.Server/WorldObjects/Player_Allegiance.cs
```diff
@@ -92,7 +92,7 @@ public void SwearAllegiance(uint targetGuid, bool success, bool confirmed = fals
                 return;
             }
 
-            log.Info($"{Name} swearing allegiance to {patron.Name}");
+            log.Debug($"{Name} swearing allegiance to {patron.Name}");
```
gmriggs: not a debug message

Source/ACE.Server/WorldObjects/Player_Death.cs
```diff
@@ -75,7 +75,7 @@ public override DeathMessage OnDeath(WorldObject lastDamager, DamageType damageT
 
             var broadcastMsg = new GameMessageSystemChat(nearbyMsg, ChatMessageType.Broadcast);
 
-            log.Info(nearbyMsg);
+            log.Debug("[CORPSE] " + nearbyMsg);
```
gmriggs: definitely not a debug message

Issue #2209, opened 2019-08-10
Ability to modify character vitals/attributes/skills in game. 


cyberkiller6:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Desired functionality.
Currently modifying a character's vitals/attributes/skills through the db, requires restarting ACE to apply. 

Some sort of command that would force a reload of a character from the db would allow changes to be made and used without restarting the entire ACE server. 

Or an ingame command that would allow changing vitals/attributes/skills on a character in real time.

This could be used in conjunction with the /morph command to create player controlled monsters.

Issue #2208, opened 2019-08-10
Ability to modify creatures vitals/attributes/skills in game. 


cyberkiller6:
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Desired functionality.

Would like the ability to modify a creatures vitals/attributes/skills in game. 

Example usage, use the command /create to spawn a mob. Then select that mob and increase it's max health and attack/defense skills. 

This would allow creating unique one time creatures in game without having to create new weenies or restarting the server.

PR #2219, opened 2019-08-11
Commands to modify max-vitals and skills

dgarson:
Adds two new admin commands for (1) modifying the max vitals for an NPC or player, (2) modifying the skill levels for an NPC or player

https://github.com/ACEmulator/ACE/issues/2209 and https://github.com/ACEmulator/ACE/issues/2208

### DIFF
--- a/Source/ACE.Server/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE.Server/Command/Handlers/AdminCommands.cs
@@ -22,6 +22,7 @@ using ACE.Server.Network;
 using ACE.Server.Network.GameMessages.Messages;
 using ACE.Server.WorldObjects;
 using ACE.Server.Network.Enum;
+using ACE.Server.WorldObjects.Entity;
 
 using Position = ACE.Entity.Position;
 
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1865,6 +1866,111 @@ public static void HandleMagicGod(Session session, params string[] parameters)
             ChatPacket.SendServerMessage(session, "You are now a magic god!!!", ChatMessageType.Broadcast);
         }
 
+        
+        [CommandHandler("modifyvital", AccessLevel.Admin, CommandHandlerFlag.None, 2, "Adjusts the maximum vital attribute for the last appraised mob/player and restores full vitals", "<Health|Stamina|Mana> <delta>")]
+        public static void HandleModifyVital(Session session, params string[] parameters)
+        {
+            var lastAppraised = CommandHandlerHelper.GetLastAppraisedObject(session);
+            if (lastAppraised == null || !(lastAppraised is Creature)) return;
```
dirtyelf: Might be useful to output text to the admin letting them know why the command failed.

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1865,6 +1866,160 @@ public static void HandleMagicGod(Session session, params string[] parameters)
             ChatPacket.SendServerMessage(session, "You are now a magic god!!!", ChatMessageType.Broadcast);
         }
 
+        
+        [CommandHandler("modifyvital", AccessLevel.Admin, CommandHandlerFlag.None, 2, "Adjusts the maximum vital attribute for the last appraised mob/player and restores full vitals", "<Health|Stamina|Mana> <delta>")]
+        public static void HandleModifyVital(Session session, params string[] parameters)
+        {
+            var lastAppraised = CommandHandlerHelper.GetLastAppraisedObject(session);
```
gmriggs: my only concern about GetLastAppraisedObject() is how decal can automatically appraise other objects invisibly / behind-the-scenes, and unintentionally change the intended target of the player

there is also a field in ACE which points to the 'currently selected creature target' (anything with vitals), and this selection is not automatically changed by decal, as far as i know... perhaps it would be worth considering to use that variant?

dgarson: This seems like a good idea. I found the `selectedTarget` field in Player class but it's a Guid. What's the easiest way to convert that to a Creature instance?

mcreedjr: Something like this should work:
Creature targetCreature = CurrentLandblock?.GetObject(Guid) as Creature;

PR #2223, opened 2019-08-12
Support Purging of deleted Characters from database

### LINKED ISSUE #2220: Remove deleted chars and associated entries from shard db
### Bug Report or Feature Request (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
```

### Desired functionality.

This is not a high priority item. More of a wish list sort of feature. 

Currently when a character is deleted, the character is marked as deleted in the shard db and that's it. The actual entries in the db aren't removed/deleted.

What I propose would be a way to remove these deleted characters along with any other associated db entries like biotas / inventory items. 

This could be done through an SQL file (ran manually against the db by the server operator) or a function builtin to ACE. Being built into ACE, would allow automatic clean up. 

This would be a configurable option in days that would default to off. So you could set the configuration to say 30 to remove any deleted characters from the db older than 30 days. This would still allow being able to look back at recently deleted characters. 

The function in ACE would cleanly and properly delete the character and all related entries from the shard db by the date of deletion set by the configurable option. This could either run on server startup or a set interval while the server is running. 

One thing to consider would be ACE has builtin GUID recycling. So when these entries are deleted from the database, would their GUIDs then be freed to use again?

Main reason for this feature would be to reduce the growing size of shard db's on high population servers and to keep the db as efficient as possible.

LtRipley36706:
Resolves #2220


Source/ACE.Database/SerializedShardDatabase.cs
```diff
@@ -372,6 +372,25 @@ public List<Biota> GetHousesOwned()
...
+        {
+            return _wrappedDatabase.CanPurgeCharacter(characterId);
+        }
+
+        public bool PurgeCharacter(uint characterId)
+        {
+            return _wrappedDatabase.PurgeCharacter(characterId);
+        }
+
+        public bool PurgeCharacters(int daysLimiter, out int numberOfCharactersPurged)
```
Mag-nus: this function shouldn't be in the serializedshard class, as it shouldn't be run at run time.

Source/ACE.Database/SerializedShardDatabase.cs
```diff
@@ -372,6 +372,25 @@ public List<Biota> GetHousesOwned()
...
+        {
+            return _wrappedDatabase.PurgeCharacters(daysLimiter, out numberOfCharactersPurged);
+        }
+
+        public bool PurgeOrphanedBiotas(out int numberOfBiotasPurged)
```
Mag-nus: this function shouldn't be in the serializedshard class, as it shouldn't be run at run time.

Source/ACE.Server/Command/Handlers/AdminShardCommands.cs
```diff
@@ -179,5 +180,19 @@ public static void HandleHelp(Session session, params string[] parameters)
             else if (close)
                 WorldManager.Close(session == null ? null : session.Player, bootPlayers);
         }
+
+        // purgeorphans
+        [CommandHandler("purgeorphans", AccessLevel.Admin, CommandHandlerFlag.None, 0,
+            "Purges all orphaned objects found in the database")]
+        public static void HandlePurgeOrphans(Session session, params string[] parameters)
```
Mag-nus: Any type of db purge should be done at startup, or, completely separate from ACE, while ACE is offline.

Mag-nus:
My only remaining feedback is that this feature should be enabled via the config.js, not a shard property.

To me, this is a feature specific to a physical server, and should travel with the shard as it moves across servers, or gets shared for use on another server.

PR #2226, opened 2019-08-12
able to create stacks and multiple objects with /create

dirtyelf:
/create now takes a second argument (defaults to one) and will create a stack if possible or spawn that many of the weenie you specify

### DIFF
--- a/Source/ACE.Server/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE.Server/Command/Handlers/AdminCommands.cs
@@ -1358,7 +1358,7 @@ namespace ACE.Server.Command.Handlers
 
             weenieToSpawn = CreateObjectForCommand(session, weenieClassDescription);
 
-            // Only allows for spawning multiple weenies of type create or stackable.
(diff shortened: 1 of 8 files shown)


Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1315,56 +1315,100 @@ static WorldObject CreateObjectForCommand(Session session, string weenieClassDes
 
         // create wclassid (number)
         [CommandHandler("create", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 1,
-            "Creates an object in the world.", "wclassid (string or number)")]
+            "Creates an object or objects in the world.", "wclassid (string or number), Amount to Spawn (default:1)")]
```
gmriggs: does the description include all of the parameters?
numToSpawn / palette / stackSize?

dirtyelf: numToSpawn essentially equals stacksize here... palette and shade are left out.. not sure what they do really and they seem to be rarely used

PR #2229, opened 2019-08-13
Update rares

Theran578:
- Remove a few retired rares and updated one that changed wcids


gmriggs:
looks like a bunch of rares removed... with the wcids, it is tough to verify this list, but i will assume the updated data is correct..

PR #2242, opened 2019-08-17
Add some config.js.example comments

Mag-nus:


Source/ACE.Server/Config.js.example
```diff
@@ -16,9 +16,33 @@
     },
     "DatFilesDirectory": "c:\\ACE\\",
```
gmriggs: DatFilesDirectory should be commented

One of the most common pitfalls is that people will replace this with single \, even though the default folder is double \'s

Source/ACE.Server/Config.js.example
```diff
@@ -16,9 +16,33 @@
...
+    // Enable this to cache the world database at startup.
+    // This will add about 1-2 minutes to server startup time and will consume about 1.7 GB of RAM
+    // With this disabled, caching happens on an as-requested basis and can cause the server to feel less responsive.
     "WorldDatabasePrecaching": false,
+
+    // LandblockPreloading, defaut: true
+    // Development recommendation: true
+    // Server recommendation: true
+    // This makes sure that perma-load landblocks are loaded.
+    // Perma-load lanblocks are landblocks that can contain server-wide mechanics and should always be active.
```
gmriggs: lanblocks typo

Source/ACE.Server/Config.js.example
```diff
@@ -16,9 +16,33 @@
...
     "LandblockPreloading": true,
+
+    // PreloadedLandblock
+    // These are the default pre-loaded and perma-loaded landblocks.
+    // In most cases you should not need to adjust them.
```
gmriggs: One confusing part about this section is that 'enabled: false' disables the entire section, even if permaload is set to true. Maybe some comments about enable

Source/ACE.Server/Config.js.example
fartwhif: an overall comment should be included to familiarize the user with the escape sequences that are in use.

Mag-nus: Please provide a suggested comment and I can add it.

gmriggs:
```
The folder that contains AC DAT files (client_cell_1.dat, client_portal.dat, client_highres.dat, client_local_English.dat)
Note that for servers running in Linux, the ~ home folder must be fully expanded, ie. /home/user/ace/, instead of ~/ace/
Note that for servers running on Windows, a path like c:\ace\ MUST be entered as c:\\ace\\, replacing the single backslashes (\) with double backslashes (\\)
```

Source/ACE.Server/Config.js.example
```diff
@@ -16,9 +16,34 @@
...
+    // Development recommendation: true
+    // Server recommendation: true
+    // This makes sure that perma-load landblocks are loaded.
+    // Perma-load landblocks are landblocks that can contain server-wide mechanics and should always be active.
     "LandblockPreloading": true,
+
+    // PreloadedLandblock
+    // These are the default pre-loaded and perma-loaded landblocks.
+    // In most cases you should not need to adjust them.
+    // This section requres LandblockPreloading to be true
```
gmriggs: typo: requres

PR #2245, opened 2019-08-18
Add StuckGuidAllocator

LtRipley36706:
StuckGuidAllocator:
- StuckGuidAllocator is used for any object that cannot be picked up/owned by a player.
- It restarts at Max and the range is 0x60000000 to 0x6FFFFFFF.

- Fixed an issue with creatures dropping wielded treasure to their corpses.


Mag-nus:
I'm strongly against this kind of separation.

It will absolutely lead to an error due to misuse, and you'll end up with guid errors.

Even with the current dynamic guid recycling disabled, guid allocation in ACE is no longer an issue.

PR #2251, opened 2019-08-19
fixing dual wielding

gmriggs:


LtRipley36706:
changed code in #2256 to be checking for validlocations not being null or not being 0x0 to start with

PR #2260, opened 2019-08-20
Add global configuration for SpellComponentsRequired

LtRipley36706:


Mag-nus:
Didn't we already have a burn multiplier?

LtRipley36706:
> Didn't we already have a burn multiplier?

not a comp burn change, this is a global override to force client/server to not require any spell components for casting

Source/ACE.Server/Managers/PropertyManager.cs
```diff
@@ -502,6 +502,7 @@ public static void LoadDefaultProperties()
                 ("quest_info_enabled", new Property<bool>(false, "toggles the /myquests player command")),
                 ("salvage_handle_overages", new Property<bool>(false, "in retail, if 2 salvage bags were combined beyond 100 structure, the overages would be lost")),
                 ("show_dot_messages", new Property<bool>(false, "enabled, shows combat messages for DoT damage ticks. defaults to disabled, as in retail")),
+                ("spell_requires_spell_comps", new Property<bool>(true, "if FALSE players casting spells does not require spell components in inventory. defaults to enabled, as in retail")),
```
gmriggs: can this name be shortened?

require_spell_comps perhaps?

Mag-nus:
> 
> 
> > Didn't we already have a burn multiplier?
> 
> not a comp burn change, this is a global override to force client/server to not require any spell components for casting

I thought you could change the burn multiplier to not consume any comps.

LtRipley36706:
> > > Didn't we already have a burn multiplier?
> > 
> > 
> > not a comp burn change, this is a global override to force client/server to not require any spell components for casting
> 
> I thought you could change the burn multiplier to not consume any comps.

even if you can, that is not the same thing as this. This boolean makes so the client (and server) doesn't do any checking if your inventory contains comps

Source/ACE.Server/WorldObjects/Player_Magic.cs
```diff
@@ -1230,6 +1231,8 @@ public bool HasComponentsForSpell(Spell spell)
         {            
             spell.Formula.GetPlayerFormula(this);
 
+            if (!PropertyManager.GetBool("spell_requires_spell_comps").Item) return true;
```
gmriggs: get name does not match up with property name

Source/ACE.Server/WorldObjects/Player_Networking.cs
```diff
@@ -103,6 +106,12 @@ private void SendSelf()
             SendContractTrackerTable();
         }
 
+        private void SendPropertyUpdatesAndOverrides()
+        {
+            if (!PropertyManager.GetBool("spell_requires_spell_comps").Item)
```
gmriggs: get name does not match up with property names

PR #2261, opened 2019-08-20
Limit database from taking all the threads

Mag-nus:
This adds limits to the database thread consumption.

It also should allow the world to now consume threads easier.

The way it is done is as follows.

- We determine the number of available threads using Environment.ProcessCount
- We allocate (int)Math.Max(Environment.ProcessorCount * .34, 1) to the World, and the remainder tothe database.

It breaks down as follows
1 vCPU = ACE will not start
2 vCPU = 1 thread world, 1 thread database
3 vCPU = 1 thread world, 2 thread database
4 vCPU = 1 thread world, 3 thread database
5 vCPU = 1 thread world, 4 thread database
6 vCPU = 2 thread world, 4 thread database
7 vCPU = 2 thread world, 5 thread database
8 vCPU = 2 thread world, 6 thread database
9 vCPU = 3 thread world, 6 thread database
10 vCPU = 3 thread world, 7 thread database
11 vCPU = 3 thread world, 8 thread database
12 vCPU = 4 thread world, 8 thread database

I'd like to get some feedback from this PR on various sized servers.

What you may notice is that loading a player may take slightly longer (very slightly).

What you will probably notice is no discernible difference in-game.

What I want to make sure happens is that the world doesn't end up feeling more choppy due to the parallel processing of outbound network traffic. Hopefully the more fair thread distribution will help prevent thread starvation.


Mag-nus:
So far results have been positive.

Theran578:
Been using this PR on my local server since about the time it was posted and have not experienced any ill effects.

Mag-nus:
> 
> 
> Need to support even those computers that have 1 cpu

That's just not realistic. For example, let's say you want to load a player on a 1 CPU system. Loading the player and all of their inventory might take 1s - 5s. During that time, the entire world will be frozen. Every time a player saves, the entire world will be frozen.

Any time a new player logs in, the entire world will be frozen.

1 cpu is just not a realistic environment that can run ACE.

PR #2262, opened 2019-08-20
Add UseCreateItem + UseCreateQuantity support to Gems

LtRipley36706:


Source/ACE.Server/WorldObjects/Gem.cs
```diff
@@ -14,6 +14,8 @@
 using ACE.Server.Network.Structure;
 using ACE.Server.Network.GameEvent.Events;
 using ACE.Server.Network.GameMessages.Messages;
+using ACE.Database;
```
gmriggs: usings not sorted

Source/ACE.Server/WorldObjects/Gem.cs
```diff
@@ -101,6 +103,87 @@ public override void ActOnUse(WorldObject activator)
                     player.Session.Network.EnqueueSend(new GameMessageSystemChat($"{Name} accepted. Click on the quill icon in the lower right corner to open your contract tab to view your active contracts.", ChatMessageType.Broadcast));
             }
 
+            if (UseCreateItem.HasValue && UseCreateItem > 0)
```
gmriggs: many of these UseCreateItem.HasValues are redundant

this is equivalent to if (UseCreateItem > 0)

PR #2263, opened 2019-08-20
Update TakeItems msg to match retail messaging

LtRipley36706:


Source/ACE.Server/Managers/EmoteManager.cs
```diff
@@ -1066,7 +1066,7 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
                             var itemTaken = WorldObjectFactory.CreateWorldObject(DatabaseManager.World.GetCachedWeenie(emote.WeenieClassId ?? 0), ObjectGuid.Invalid);
                             if (itemTaken != null)
                             {
-                                var msg = $"You hand over {emote.StackSize ?? 1} of your {itemTaken.GetPluralName()}.";
+                                var msg = $"You hand over {(player.GetNumInventoryItemsOfWCID(emote.WeenieClassId ?? 0) == 0 ? "all" : (emote.StackSize ?? 1).ToString())} of your {itemTaken.GetPluralName()}.";
```
gmriggs: is it confirmed that 'all' was sent even when the player only has 1 item?

LtRipley36706: yes

PR #2285, opened 2019-08-25
Update Landblock.cs

LtRipley36706:
- Update Logging messages
- Notify Generator of spawn failure for regen retry

### DIFF
--- a/Source/ACE.Server/Entity/Landblock.cs
+++ b/Source/ACE.Server/Entity/Landblock.cs
@@ -766,10 +766,13 @@ namespace ACE.Server.Entity
                     wo.CurrentLandblock = null;
 
                     if (wo.Generator != null)
-                        log.Debug($"AddWorldObjectInternal: couldn't spawn 0x{wo.Guid}:{wo.Name} from generator {wo.Generator.WeenieClassId} - 0x{wo.Generator.Guid}:{wo.Generator.Name}");
+                    {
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -766,10 +766,13 @@ private bool AddWorldObjectInternal(WorldObject wo)
                     wo.CurrentLandblock = null;
 
                     if (wo.Generator != null)
-                        log.Debug($"AddWorldObjectInternal: couldn't spawn 0x{wo.Guid}:{wo.Name} from generator {wo.Generator.WeenieClassId} - 0x{wo.Generator.Guid}:{wo.Generator.Name}");
+                    {
+                        log.Debug($"AddWorldObjectInternal: couldn't spawn 0x{wo.Guid}:{wo.Name} [{wo.WeenieClassId} - {wo.WeenieType}] from generator {wo.Generator.WeenieClassId} - 0x{wo.Generator.Guid}:{wo.Generator.Name}");
+                        wo.NotifyOfEvent(RegenerationType.PickUp); // Notify generator the generated object is effectively destroyed, use Pickup to catch both cases.
```
gmriggs: is it possible this could send some generators into an infinite loop?

LtRipley36706: it would keep re-trying spawn per regen rate + delay rate rules

LtRipley36706: depending on generator it also could select another profile

PR #2287, opened 2019-08-25
adding server option for aetheria healing over time message color

gmriggs:
This option was requested, it seems that a number of players (including myself) would like an option for the aetheria / healing over time message color to be green or blue, instead of the red combat text color


Mag-nus:
This can also be done via: http://www.virindi.net/wiki/index.php/Virindi_Chat_System_5

PR #2288, opened 2019-08-26
Guid Recycle Redux

LtRipley36706:
- Added code to send DestroyObject messages from objects being destroyed to those that know about it when it is not on a landblock (Corpse/Chest/Container)
- Re-enabled Guid Recycling


Mag-nus:
Can you name a pcap/line # that shows this actually happening?

This will add a lot of network overhead, and I still doubt this is what happened in retail.

PR #2303, opened 2019-09-16
Multi-thread landblock ticking (Landblock Groups w/Thread Safety)

Mag-nus:
This is the start of "landblock group" ticking.

The idea behind the landblock groups are that each group may contain multiple landblocks that must be ticked on the same thread, but, each group itself can be ticked on independent threads.

Landblock groups describe their area using a rectangle. As landblocks are added (and removed) the rectangle that contains all the landblocks in the group is adjusted.

Landblocks are added to groups based on a minimum distance to the landblock groups rectangle.

In the event a landblock needs to be added and is within range of multiple groups, those groups will be combined into one.

Periodically, landblock groups are checked to see if they can be split. When a landblock group is split, it can result in a new group that has an overlapping rectangle to the parent group. This is not a problem because all the landblocks within each group meet the minimum distance requirements. In the event that a new landblock is added and is close to both rectangles, the groups will then be merged again.

Adding landblocks to groups is a very efficient process
Removing landblocks from groups is a very efficient process
Checking landblock groups for split potential does incur some overhead (~0.5 ms) which is why it's only done on intervals that are twice the Landblock.UnloadInterval.

  
  
Things still to do: Investigate more thread safety in the following scenarios
- Actions that come in from the client do not need to be checked for thread safety. Client actions are processed in a single thread.
- Actions are the result of a landblock tick, or world object tick need to be checked for thread safety.
- World actions that result in fellowship adjustments should be investigated
- World actions that result in allegiance adjustments should be investigated

Creature OnDeath needs to examined.
- Players in the DamageHistory list can be on separate threads.
- This includes player xp, fellowship xp, allegiance xp.
- The target actor of an emote action can be on a separate thread.
  
Additional Changes we've had to made for thread safety
- ACE.Server/Network/Sequence/SequenceManager
- Source/ACE.Server/Physics/Common/ObjectMaint (Need to safely manage/maintain relationships)

ToDo:
- ACE\Source\ACE.Server\Entity\Fellowship.cs
- ACE\Source\ACE.Server\Managers\AllegianceManager.cs
- ACE\Source\ACE.Server\WorldObjects\Allegiance.cs
- ACE\Source\ACE.Server\Managers\QuestManager.cs
- ACE\Source\ACE.Server\Managers\HouseManager.cs


shadowroot-ac:
Have been using this WIP branch on my personal server for about seven days or so and have not encountered any issues with its inclusion.  Admittedly, my server isn't very heavily used; however, the with the use it has gotten, no stability issues or bugs have been experienced.

Mag-nus:
> been using this WIP branch on my per

Thank you for the feedback. The config options have now been moved to config.js, see config.js.example for the new properties.

PR #2315, opened 2019-09-23
Support SpellSets for Items that don't level

### LINKED ISSUE #2314: ItemSet spells applied to armor do not cast
### Bug Report or Feature Request (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
Add an EquipmentSetID value from 13 to 29 to a loot armor item piece.  Debug 'level' at location https://github.com/ACEmulator/ACE/blob/master/Source/ACE.Server/WorldObjects/WorldObject_Set.cs#L157, when equipping armor piece with an EquipmentSetID. One will notice that 'level' is always zero, so spells that should be cast by the item set are never get applied.

### The log given by the failure.
No log given by the failure

### Desired functionality.
A given spell's level from an Armor EquipmentSetID should be based upon the count of the pieces equipped with a given EquipmentSetID.

### Mention any other details that might be useful.
The code at the location linked above works for Aetheria.  However, it does not work for armor spell sets.  While Aetheria spell set spells base their level on the cumulative ItemLevel of similar Aetheria equipped, armor spell set level are based upon the count of armor pieces equipped that contain the same spell set.

Whether correct or not, I found that changing the line of code at 157 to the following, allows the set spells to cast, and their spell level increase and decrease, as pieces with the same EquipmentSetID are added or removed.
            uint level;
            if (equipmentSet >= EquipmentSet.Soldiers && equipmentSet <= EquipmentSet.Lightningproof)
                level = (uint)setItems.Count;
            else
                level = (uint)(setItems.Sum(i => i.ItemLevel ?? 0) + levelDiff);

LtRipley36706:
Resolves #2314

### DIFF
--- a/Source/ACE.DatLoader/Entity/SpellSet.cs
+++ b/Source/ACE.DatLoader/Entity/SpellSet.cs
@@ -12,13 +12,17 @@ namespace ACE.DatLoader.Entity
         // client calls this m_PieceCount
         public SortedDictionary<uint, SpellSetTiers> SpellSetTiers = new SortedDictionary<uint, SpellSetTiers>();
 
+        public uint HighestTier = 0;
+
         public void Unpack(BinaryReader reader)
         {
             SpellSetTiers.UnpackPackedHashTable(reader);
 
+            HighestTier = SpellSetTiers.Keys.LastOrDefault();
+
             SpellSetTiers lastSpellSetTier = null;
 
-            for (uint i = 0; i < SpellSetTiers.Keys.LastOrDefault(); i++)
+            for (uint i = 0; i < HighestTier; i++)
             {                
                 if (SpellSetTiers.ContainsKey(i))
                     lastSpellSetTier = SpellSetTiers[i];
@@ -27,8 +31,7 @@ namespace ACE.DatLoader.Entity
                     if (lastSpellSetTier != null)
                         SpellSetTiers.Add(i, lastSpellSetTier);
                 }
-
-            }
+            }            
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Source/ACE.Server/WorldObjects/Player_Spells.cs
```diff
@@ -141,9 +141,9 @@ public void OnItemLevelUp(WorldObject item, int prevItemLevel)
 
             var levelDiff = prevItemLevel - (item.ItemLevel ?? 0);
 
-            var prevSpells = GetSpellSet((EquipmentSet)item.EquipmentSetId, setItems, levelDiff);
+            var prevSpells = GetSpellSet((EquipmentSet)item.EquipmentSetId, item.ItemXpStyle ?? 0, setItems, levelDiff);
```
gmriggs: maybe just pass in item here, and have GetSpellSet fetch (EquipmentSet)Item.EquipmentSetId and item.ItemXpStyle ?? 0 in 1 place?

gmriggs: or not sure if item is even needed, if it should just verify setItems.Count > 0, get the first item, and use the EquipmentSetId/ItemXpStyle from that? (would require the calling the function to ensure that the items in setItems match up)

Source/ACE.DatLoader/Entity/SpellSet.cs
```diff
@@ -11,9 +12,26 @@ public class SpellSet : IUnpackable
...
+            SpellSetTiers lastSpellSetTier = null;
+
+            for (uint i = 0; i < HighestTier; i++)
+            {                
+                if (SpellSetTiers.ContainsKey(i))
+                    lastSpellSetTier = SpellSetTiers[i];
+                else
+                {
+                    if (lastSpellSetTier != null)
+                        SpellSetTiers.Add(i, lastSpellSetTier);
```
gmriggs: should not modify base dat structures

PR #2326, opened 2019-09-28
PurgeOrphanedBiotas

Mag-nus:
Need to verify that all the proper WeenieTypes that CAN exist without a wielder, container, or location are not being pruned.

Currently, that list only includes:
WeenieType.Allegiance,


Mag-nus:
@LtRipley36706 At this point the PR is done from my side.

What's left to do is determine what other weenie types should be excluded. Need your help there.

Mag-nus:
While this may be approved, I think we should get some feedback from server operators before it gets checked in.

Hopefully a few server operators can run this, and then we can wait a week or two to see if anyone complains.

PR #2339, opened 2019-10-20
updated monster spellbook probabilities

gmriggs:


Source/ACE.Server/WorldObjects/Monster_Magic.cs
```diff
@@ -60,7 +60,7 @@ public float GetSpellProbability()
             var probability = 0.0f;
 
             foreach (var spell in Biota.BiotaPropertiesSpellBook)
-                probability += spell.Probability;
+                probability += spell.Probability - 2.0f;
```
Mag-nus: Would be good to have a comment describing why - 2.0f is needed.

Source/ACE.Server/WorldObjects/Monster_Magic.cs
```diff
@@ -242,10 +242,10 @@ public BiotaPropertiesSpellBook GetRandomSpell()
             var currentSpell = 0.0f;
             foreach (var spell in Biota.BiotaPropertiesSpellBook)
             {
-                if (rng < currentSpell + spell.Probability)
-                    return spell;
+                currentSpell += spell.Probability - 2.0f;
```
Mag-nus: Same here regarding - 2.0f

Source/ACE.Server/WorldObjects/Monster_Magic.cs
gmriggs: All of the data uses a base 2.0, i.e a 5% chance would be 2.05 instead of 0.05

Why it actually does this, not sure

PR #2359, opened 2019-10-26
Ensure that Impen spell is present on magical Covenant Armor

ziang4891:
Per a discussion on LSD discord, it was determined that Covenant Armor that contains spells will have Impenetrability as one of them.  While I tested the added code in game, it should probably be tested by others before merging to ensure that I didn't miss something, especially in regards to array indexing, if the added code is deemed acceptable.


ziang4891:
I cannot push

PR #2375, opened 2019-10-31
Cleanup ActOnUse for Ammunition

LtRipley36706:


LtRipley36706:
Logging wasn't removed, error still occurs. but unless a server is in debug mode, theres no reason to tell a player a weenietype is undefined.. for whatever ones remain. they've all been added as it was certain they either did something, or nothing

gmriggs:
Public servers are almost never running in debug mode

Sure there is, this error message has found tons of items where no UseHandler was defined for. Players have seen this message and then reported the message to us.

LtRipley36706:
ok then, i've looked at all the weenie types, there are none that remain left to do.

PR #2382, opened 2019-11-02
Adjusting AddPickupChainToMoveToChain

LtRipley36706:
aka contributing nothing

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Move.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Move.cs
@@ -50,7 +50,7 @@ namespace ACE.Server.WorldObjects
             // fix bug in magic combat mode after walking to target,
             // crouch animation steps out of range
             if (useRadius == null)
-                useRadius = target.UseRadius ?? 1f;
+                useRadius = target.UseRadius ?? 0.6f;
 
             if (CombatMode == CombatMode.Magic)
                 useRadius = Math.Max(0.0f, useRadius.Value - 0.2f);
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/Player_Move.cs
```diff
@@ -50,7 +50,7 @@ public void CreateMoveToChain(WorldObject target, Action<bool> callback, float?
             // fix bug in magic combat mode after walking to target,
             // crouch animation steps out of range
             if (useRadius == null)
-                useRadius = target.UseRadius ?? 0.6f;
+                useRadius = target.UseRadius ?? 1f;
```
gmriggs: this is a client hardcoded value. this should not be changed arbitrarily.

LtRipley36706: ok, can you provide the source point for it? 

Issue is at less than 1f, items on desks, tables, and similar furniture cannot be reachable without getting on top of them.

gmriggs: LODWORD(this->distance_to_object) = 1058642330;
public static readonly float Default_DistanceToObject = 0.6f;

this can also be seen in all of the pcaps where it sends 0.6f for the default DistanceToObject for almost everything in the game.

LtRipley36706: very well. How do we know that UseRadius == Default_DistanceToObject, is it at all possible they are different?

gmriggs: just examine the pcaps for any object. it sends the UseRadius for DistanceToObject in the MoveToObject MovementParameters

PR #2388, opened 2019-11-05
add valid skills hashset, check valid skills entering godmode

dirtyelf:
added a valid skills hashset to the skillhelper so that it could be checked against when entering godmode. also added hard coded checks against the attribute and attribute 2nd types... figured they didnt need a helper

this change will only add entries to the godstring if they are valid. 

According to ripley, the bug is with the GetOrAddSkill function that adds a skill to a players biota_properties_skill table when checking for invalid skill types. if a player had an invalid skill (typically this was a skill with type 0) the string would be too long and cause the error. now when entering godmode invalid skill and attribute values will be ignored and everything should work as intended

the only way an error when attempting to god/ungod should happen now is if the db is externally corrupted or manually changed... or until the next bug is uncovered :)

### DIFF
--- a/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
+++ b/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using System.Threading;
 
 using ACE.Entity;
+using ACE.Entity.Enum;
 using ACE.Entity.Enum.Properties;
 
 namespace ACE.Database.Models.Shard
@@ -1170,9 +1171,18 @@ namespace ACE.Database.Models.Shard
                 try
                 {
                     entity = new BiotaPropertiesSkill { ObjectId = biota.Id, Type = type, Object = biota };
-                    biota.BiotaPropertiesSkill.Add(entity);
-                    skillAdded = true;
-                    return entity;
+
+                    if (SkillHelper.ValidSkills.Contains((Skill)type))
+                    {
+                        biota.BiotaPropertiesSkill.Add(entity);
+                        skillAdded = true;
+                        return entity;
+                    }
+                    else
+                    {
+                        skillAdded = false;
+                        return entity;
+                    }
                 }
                 finally
                 {
--- a/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
+++ b/Source/ACE.Database/Models/Shard/BiotaExtensions.cs
@@ -3,7 +3,6 @@ using System.Linq;
 using System.Threading;
 
 using ACE.Entity;
-using ACE.Entity.Enum;
 using ACE.Entity.Enum.Properties;
 
 namespace ACE.Database.Models.Shard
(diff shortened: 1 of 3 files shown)


dirtyelf:
figured i would fix the root of the problem too by checking against the valid skills hashset before adding new skills with GetOrAddSkill

not sure if this fix is appropriate... based on my understanding i think this would still return a creature skill object if the call is made with an inappropriate skill value, but it wouldnt add the value to the db. the onus is on the user of the function to make sure they arent checking or using invalid skills at that point... with 99+ references to GetCreatureSkill i cant possibly check them all

Source/ACE.Database/Models/Shard/BiotaExtensions.cs
```diff
@@ -1170,9 +1171,18 @@ public static BiotaPropertiesSkill GetOrAddSkill(this Biota biota, ushort type,
                 try
                 {
                     entity = new BiotaPropertiesSkill { ObjectId = biota.Id, Type = type, Object = biota };
-                    biota.BiotaPropertiesSkill.Add(entity);
-                    skillAdded = true;
-                    return entity;
+
+                    if (SkillHelper.ValidSkills.Contains((Skill)type))
```
gmriggs: This fix is not appropriate, and will break a ton of functionality.

Almost all monsters in the game will basically have 0 attack skill with this change.

gmriggs: This fix is not appropriate, and will break a ton of functionality.

Almost all monsters in the game will basically have 0 attack skill with this change.

dirtyelf: Could you elaborate at all? Do mobs still use the retired skills? 

For all skills in the valid skills hashset this should provide the same functionality as before. 

The difference is now if a call to GetOrAddSkill is made with an invalid skill (i.e. type = 0).

The first conditional...
```
var entity = biota.BiotaPropertiesSkill.FirstOrDefault(x => x.Type == type);
                if (entity != null)
                {
                    skillAdded = false;
                    return entity;
                }
```
...will be passed over because entity will be set to null (default = null?) when a skill of type 0 is not found in the biota.

The second conditional...
```
entity = new BiotaPropertiesSkill { ObjectId = biota.Id, Type = type, Object = biota };

                    if (SkillHelper.ValidSkills.Contains((Skill)type))
                    {
                        biota.BiotaPropertiesSkill.Add(entity);
                        skillAdded = true;
                        return entity;
                    }
                    else
                    {
                        skillAdded = false;
                        return entity;
                    }
```
...will populate entity with the invalid skill type = 0 and then pass over the BiotaPropertiesSkill.Add block because type = 0 is not in the valid skills hashset. Instead it will fall to the else case and set skillAdded = false and return the entity with the invalid skill type... just like it would have before these changes, except without the call to BiotaPropertiesSkill.Add and setting skillAdded to true.

thank you for the review, please help me to understand the error with this logic

Mag-nus: To add to the review, The BiotaExtensions should not be performing this specific type of validation. These BiotaExtensions are simply used for interfacing safely with the records themselves.

If you want to perform validation on the input, or result, it should be done by your caller or consumer of the function.

dirtyelf: Blindly adding skills of invalid types is considered a "safe" way to interact with the records?

You are suggesting that the validation should happen at the GetCreatureSkill level?

gmriggs: "Could you elaborate at all? Do mobs still use the retired skills?"

yes, a ton of the mobs still use UnarmedCombat, for example

dirtyelf:
Ah. Definitely didn’t realize that mobs used deprecated skills.

I’ll back these changes out of this PR when I get home and possibly look into fixing this elsewhere in the call chain like suggested. 

Thanks guys.

PR #2391, opened 2019-11-07
Assign armor level

ziang4891:
- Update AssignArmorLevel() to account for updated base AL weenie values and align AL mod with armor type, ie, cloth, leather, chainmail, metal, etc.
- Maintains original AssignArmorLevel code for a compatibility layer for databases that have not been updated, yet
- Should be paired with an [update](https://github.com/ACEmulator/ACE-World-16PY-Patches/pull/253) to ACE-World-Patches but should be safe to use without updated armor objects


ziang4891:
Closing apparently unwanted change

PR #2392, opened 2019-11-08
Allow long running OnDeath emotes to complete

LtRipley36706:


Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1342,7 +1340,12 @@ public void Enqueue(BiotaPropertiesEmote emoteSet, WorldObject targetObject, int
                         Nested--;
 
                         if (Nested == 0)
+                        {
                             IsBusy = false;
+                            if (OnDeathEmoteInProgress)
```
gmriggs: this has a chance to fail, and OnDeathEmoteInProgress to remain true forever

as mentioned, IsBusy and Nested state can get trashed with simultaneous emotes running, and the state of these 2 vars cannot be trusted reliably after OnDeath sets IsBusy=false and forces itself to run.

Source/ACE.Server/WorldObjects/WorldObject.cs
```diff
@@ -890,6 +890,20 @@ public virtual void Destroy(bool raiseNotifyOfDestructionEvent = true)
             if (raiseNotifyOfDestructionEvent)
                 NotifyOfEvent(RegenerationType.Destruction);
 
+            if (!(this is Creature))
```
gmriggs: @Mag-nus should probably check this part

Source/ACE.Server/WorldObjects/WorldObject.cs
```diff
@@ -898,6 +912,19 @@ public virtual void Destroy(bool raiseNotifyOfDestructionEvent = true)
                 GuidManager.RecycleDynamicGuid(Guid);
         }
 
+        private void CheckForOnDeathCompletion()
```
gmriggs: polling is not the best design here, should probably use a callback

Source/ACE.Server/WorldObjects/WorldObject.cs
```diff
@@ -898,6 +912,19 @@ public virtual void Destroy(bool raiseNotifyOfDestructionEvent = true)
...
+        {
+            if (!EmoteManager.OnDeathEmoteInProgress)
+                RemoveObjectAndRecycleGuid();
+            else
+            {
+                var recheck = new ActionChain();
```
gmriggs: polling is not the best design here, should probably use a callback

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
gmriggs: update: OnDeathEmoteInProgress will not remain true forever as originally thought, it just might be delayed if a Nested emote is already running when OnDeath is called. very unlikely / not a realistic scenario with any emote data i'm aware of, so not a big issue

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1304,6 +1306,13 @@ public void Enqueue(BiotaPropertiesEmote emoteSet, WorldObject targetObject, int
             if (emoteSet == null)
             {
                 Nested--;
+
+                if (OnDeathEmoteInProgress)
```
gmriggs: this branch should technically be unneeded

what this should be doing, is simply setting OnDeathEmoteInProgress = HasEmoteCategory(EmoteCategory.Death);

gmriggs: actually, this part is completely unsafe, and should be removed

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1343,6 +1343,13 @@ public void Enqueue(BiotaPropertiesEmote emoteSet, WorldObject targetObject, int
 
                         if (Nested == 0)
                             IsBusy = false;
+
+                        if (emoteSet.Category == (uint)EmoteCategory.Death)
```
gmriggs: this should only run if Nested == 0

gmriggs: redacted, this is using Category==Death instead of OnDeathEmoteInProgress

this should only be true in the base branch of OnDeath, so this should be fine

PR #2395, opened 2019-11-08
Separate Armor from MiscClothing in loot gen

ziang4891:
Separate MiscClothing items out from being lumped into armor in general, without creating code duplication, by adding a flag to CreateArmor() that specifies either one or the other.  MiscClothing items should now have a fixed one in five chances of being generated, instead of a variable 1 in 6 to 1 in 28 chances, depending upon loot tier.

Code compiles but have not run/play tested

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -237,31 +237,35 @@ namespace ACE.Server.Factories
                     type = 2;
                     break;
                 case LootBias.Weapons:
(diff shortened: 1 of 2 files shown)


harliq:
From Magnus's data, clothing has a 5.1-7.8% chance of dropping depending on tier, its basically half of armor drop chance. By adding clothing chance to this roll, it will greatly reduce the number of drops of armor/weapons, and overflowing the server with a lot of clothing.  This one of the main area's (lootbias) that needs to be updated and aligned with what drop rate data we have.

PR #2398, opened 2019-11-09
Removing unused pattern from CryptoSystem for slightly better performance.

fartwhif:
Adding .NET IDisposable pattern to CryptoSystem
Slightly improved CryptoSystem documentation

After considering the theory more it seems that the preventing of the search window from changing upon finding a key isn't actually needed by the system's current implementation, especially since it's been **well tested**.

Removing the unneeded parameter.  [This will reduce the memory usage on the stack and improve performance.](https://www.dotnetperls.com/parameter-optimization)  The modified source code is clearer and the new implementation more direct further improving performance as fewer stack frames are needed.


Mag-nus:
I think everything in here is great, except for the need to add IDisposable. Disposable is used to clean up unmanaged resources. https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose?redirectedfrom=MSDN

In this PR, CyrptoSystem is cleaning up managed event handlers, and collections, for which CryptoSystem is the root. The GC will take care of this for us.

I agree that the crypto system to this point is very well tested, and I'm all for improving performance.

Source/ACE.Common/Cryptography/CryptoSystem.cs
```diff
@@ -8,10 +8,10 @@ namespace ACE.Common.Cryptography
 {
     /// <summary>
     /// Forward looking cache for ISAAC calculations.<para />
-    /// Manages ISAAC stream cipher and key searches.<para />
+    /// Manages ISAAC stream cipher and performs key searches.<para />
     /// Manages a range of encryption keys generated by the stream cipher.<para />
     /// </summary>
-    public class CryptoSystem
+    public class CryptoSystem : IDisposable
```
Mag-nus: Only change requested would be to remove IDisposable. See my comment on the main PR conversation.

fartwhif: I'm sorry but that's wrong.  The IDisposable pattern absolutely works for managed objects and the habit of cleaning up reference objects in the **managed section** of the IDisposable pattern is encouraged and follows Microsoft best practices.   [It is highly inefficient for an application to force garbage collection as part of the service control flow.](https://lostechies.com/chrispatterson/2012/11/29/idisposable-done-right/)

Mag-nus: In the link you provide, he also references MS documentation that is titled "Implementing Finalize and Dispose to Clean Up **Unmanaged** Resources"

In this case, we're talking about implementing excessive cleanup for single EventHandler subscription per Session/NetworkSession, aka, 1 per client connection.

When a Session drops, the tie to root will be lost, and GC will clean these up. I also know this because I've spent considerable amount of time analyzing memory profiles from very populated ACE servers in effort to help reduce memory consumption.

fartwhif: i would like a second opinion

fartwhif: > In this case, we're talking about implementing excessive cleanup for single EventHandler subscription per Session/NetworkSession, aka, 1 per client connection.

you are wrong

PR #2406, opened 2019-11-10
Update EmoteType.DeleteSelf

LtRipley36706:


Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -257,7 +257,14 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
 
                 case EmoteType.DeleteSelf:
 
-                    WorldObject.Destroy();
+                    if (player != null)
+                    {
+                        var wo = player.FindObject(WorldObject.Guid.Full, Player.SearchLocations.Everywhere, out _, out Container rootOwner, out bool wasEquipped);
```
gmriggs: This player blocks looks very strange.

What is the point of this? Are there any examples of data that would use this?

If an item in the player's inventory, calls DeleteSelf, then it's going to delete the entire player.

LtRipley36706: it isn't deleting the player, its deleting the WorldObject that has the emote DeleteSelf

gmriggs: function call is wo.DeleteObject(optional root owner), and not static WorldObject.DeleteObject(WorldObject obj);

PR #2412, opened 2019-11-12
Added NoAccess as an AccessLevel

TheKnobGoblin:
Accounts set to AccessLevel 0 will now be rejected.

0 - NoAccess
1 - Player
2 - Advocate
3 - Sentinel
4 - Envoy
5 - Developer
6 - Admin

Useful for servers that require email validation before a player can login.

### DIFF
--- a/Database/Updates/Authentication/2019-11-11-00-Add_NoAccess_To_Accesslevel_Table.sql
+++ b/Database/Updates/Authentication/2019-11-11-00-Add_NoAccess_To_Accesslevel_Table.sql
@@ -1,0 +1,26 @@
+DROP TABLE IF EXISTS `accesslevel`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+ SET character_set_client = utf8mb4 ;
+CREATE TABLE `accesslevel` (
+  `level` int(10) unsigned NOT NULL DEFAULT '0',
+  `name` varchar(45) NOT NULL,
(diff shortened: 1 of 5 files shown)


Mag-nus:
I like the idea, but, i don't think shifting the AccessLevel by 1 will work. I think that would demote existing characters and account levels.

Furthermore, it's probably not a good idea to drop the entire table in
DROP TABLE IF EXISTS `accesslevel`;
Instead, it would probably be safer to update existing rows, and add the new row.

TheKnobGoblin:
It works, you need to run the sql so that existing account levels are converted to the new ones.

Also you can't simply update the rows in the accesslevel table due to one of them being a primary key. I see nothing wrong with dropping this table and remaking it but your call.

Mag-nus:
I like the idea of this, but the implementation needs to be thought out some more.

For example, if a server adds this PR without the mysql update, their accounts won't have the appropriate access level, and many will be rejected.

if the do the SQL update before the PR, then their players have more access than they should.

ALso, the UPDATE statement should be a single query that adds 1 to all existing values that are between 0-5

TheKnobGoblin:
I've tested this and it works perfectly. You're defeating the entire purpose of mandatory sql updates when they're made. Server owners always need to install them.

Also who cares about whether the sql is on one line or not? It's just a simple update script that will be eventually moved to the base anyway.

Anyway, I'll keep this to myself and make sure not to contribute to this project anymore.

PR #2419, opened 2019-11-14
adding Storage.Default_ChestResetInterval override

gmriggs:


LtRipley36706:
can this made to be -1? Why does a limit on storage need to exist?

gmriggs:
whatever is appropriate, just wanted to get the option in to set Storage separately

gmriggs:
ok this has been updated for Chests to default to ResetInterval=double.PositiveInfinity, and special logic to skip queueing the action chain when ResetInterval is this value

another way to do this might simply be for the check to be && (!this is Storage), however if someone does wish to modify their server for Storage chests to have ResetIntervals (such as 5 mins. or 10 mins.) then this code gives them the option to easily change the default value

PR #2421, opened 2019-11-15
Update Instance IDs upon swearing in Allegiance

LtRipley36706:
Boot from housing on break

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Allegiance.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Allegiance.cs
@@ -125,6 +125,11 @@ namespace ACE.Server.WorldObjects
             AllegianceXPGenerated = 0;
             AllegianceOfficerRank = null;
 
+            // update instance ids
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Player_Allegiance.cs
```diff
@@ -125,6 +125,11 @@ public void SwearAllegiance(uint targetGuid, bool success, bool confirmed = fals
             AllegianceXPGenerated = 0;
             AllegianceOfficerRank = null;
 
+            // update instance ids
+            Session.Network.EnqueueSend(new GameMessagePrivateUpdateInstanceID(this, PropertyInstanceId.Allegiance, Allegiance.Guid.Full));
```
gmriggs: should this be setting PropertyInstanceId.Allegiance on the server?
with this current code, it looks like as soon as the player relogs, their PropertyInstanceId wouldn't be set anymore

also, consider using UpdateProperty or writing wrapper methods for common methods such as updating and broadcasting properties... this is a lot of redundant code

PR #2427, opened 2019-11-18
Support QuestManager on Creatures

LtRipley36706:


ziang4891:
Tested working on my local server

Source/ACE.Server/Managers/QuestManager.cs
```diff
@@ -20,8 +20,38 @@ public class QuestManager
 
         public Player Player { get; }
```
gmriggs: there doesn't need to be separate fields for Player and Creature

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1218,6 +1209,124 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
             return delay;
         }
 
+        private static void CheckForQuestManager(Creature creature)
```
gmriggs: this method is unneeded. suggested pattern posted to ce channel

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1218,6 +1209,124 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
...
+        {
+            if (creature != null && creature.QuestManager == null)
+                creature.QuestManager = new QuestManager(creature);
+        }
+
+        private void UpdateQuest(BiotaPropertiesEmoteAction emote, WorldObject targetObject, Creature creature)
```
gmriggs: EmoteManager should not contain special methods for particular emotes... if these needed to be methods, they belong in QuestManager

However, there is no need for these to be separate methods

PR #2428, opened 2019-11-18
ThreadSafeRandom Cleanup

Mag-nus:
This fixes the namespace for ThreadSafeRandom, from ACE -> ACE.Common.

It also adds better Summary descriptions.

The rest of the file changes are simply adding ACE.Common references.

This is mainly commenting and namespace preparation for adding a non-thread safe random class, in effort to reduce lock contention.

### DIFF
--- a/Source/ACE.Common/ThreadSafeRandom.cs
+++ b/Source/ACE.Common/ThreadSafeRandom.cs
@@ -1,21 +1,23 @@
-namespace ACE
+using System;
+
+namespace ACE.Common
 {
     // important class, ensure unit tests pass for this
+    // todo: implement exactly the way AC handles it.. which we'll never know unless we get original source code
     public static class ThreadSafeRandom
     {
         private static readonly object randomMutex = new object();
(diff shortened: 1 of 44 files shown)


Source/ACE.Common/ThreadSafeRandom.cs
```diff
@@ -1,21 +1,23 @@
...
+    // todo: implement exactly the way AC handles it.. which we'll never know unless we get original source code
     public static class ThreadSafeRandom
     {
         private static readonly object randomMutex = new object();
-        private static readonly System.Random random = new System.Random();
+        private static readonly Random random = new Random();
+
         /// <summary>
-        /// Returns a random number between min and max
+        /// Returns a random floating-point number between min and max, inclusive
```
gmriggs: Maybe add some more details here about how 0.0-1.0 can never return 1.0, as per the Random.NextDouble() api

This some detail can be important for writing property RNG that handles both the 0.0 and 1.0 cases correctly (ie. the callers should always use <, and not <=)

PR #2432, opened 2019-11-18
Switch ThreadSafeRandom from lock to ThreadLocal<Random>

Mag-nus:
This builds on the ThreadSafeRandom cleanup, which should be merged first, to allow easy rollback.


fartwhif:
`ThreadLocal` is cool but [do you know about this?](https://particular.net/blog/the-dangers-of-threadlocal)

fartwhif:
Suggestion to [copy this](https://github.com/grmartin/Mersenne-Twister), it's pure c# no underlying object so instantiation of it has real meaning as opposed to .NET Random.  And use `AsyncLocal` for futureproofing, it has the same benefits of `ThreadLocal` but is compatible with [async/await](https://particular.net/blog/the-dangers-of-threadlocal).

Mag-nus:
Ok thanks.

I don't feel comfortable straying from the native .NET implementation of Random.

I'll investigate the performance impact of AsyncLocal vs ThreadLocal. Being that this class is in ACE.Common, it could be consumed outside of ACE in an Async manner.

fartwhif:
> Ok thanks.
> 
> I don't feel comfortable straying from the native .NET implementation of Random.
> 
> I'll investigate the performance impact of AsyncLocal vs ThreadLocal. Being that this class is in ACE.Common, it could be consumed outside of ACE in an Async manner.

I take it back, I don't know enough about it all to make that recommendation.

PR #2433, opened 2019-11-18
Another CurrentMovementData.Invalid is null fix

Mag-nus:
This fixes an exception.

This does not fix the root issue

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -769,7 +769,7 @@ namespace ACE.Server.WorldObjects
         {
             //Console.WriteLine($"{Name}.HandleRunRateUpdates()");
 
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -769,7 +769,7 @@ public bool HandleRunRateUpdate()
         {
             //Console.WriteLine($"{Name}.HandleRunRateUpdates()");
 
-            if (CurrentMovementData.MovementType != MovementType.Invalid)
+            if (CurrentMovementData.MovementType != MovementType.Invalid || CurrentMovementData.Invalid == null)
```
gmriggs: this does indeed fix the root issue

as mentioned, this code incorrectly assumed that when MovementType == Invalid, then Invalid would always be populated

Invalid is the default state of MovementType, so this isn't true. this is the correct fix

PR #2440, opened 2019-11-20
CreateCaster (wand/staff/obs) refactored, TestLootGen Updated

harliq:
LootFactory CreateCaster refactored.  Items now created via wields not tier.  Updated values for the correct wields (values are based off of data from Optim and Magnus).

TestLootGen added more details for casters, and the format.

PR has been tested on private server without issue.


Mag-nus:
Instead of constantly updating ConsoleCommands.cs, please instead create a ConsoleCommands_Loot.cs where you can move and update your lot testing code too. Note: You may need to make the class Partial, if it's not already.

Also, in the future, please separate actual loot drop rate changes in a separate PR (when possible) to allow easier rollbacks.

Thanks!

harliq:
Pulling TestLootGen updates out, closing this PR, will create a new PR

PR #2449, opened 2019-11-22
Test Loot Generator

harliq:
Moved Test Loot Generator (Loot Simulator) to its own class (LootGenerationFactory_Test.cs) and moved Test Loot Gen console commands to its own class (DeveloperLootCommands.cs)


harliq:
Just seen I left ConsoleCommands as a partial class (forgot to change it back to just a static class).  Fixed

Source/ACE.Server/Command/Handlers/DeveloperLootCommands.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                Console.WriteLine("Number of items to generate " + numberItemsGenerate);
+            }
+            else
+            {
+                Console.WriteLine("numbr of items is not an integer");
+                return;
+            }
+
+            if (Int32.TryParse(parameters[1], out int itemsTier))
```
gmriggs: the following command throws an exception:

testlootgen 1

The 4th parameter to the 'CommandHandler' should be 2 instead of 1

This exception does not crash the server, as it is caught elsewhere

Whenever accessing a particular offset of an array, it should be ensured that the array is actually of that length... ie. for an array of length 100, it will have indices 0-99. If i try to access index 100 of the array, it will throw an IndexOutOfRangeException

ziang4891: I'm always forgetting that array indexes start at zero...

PR #2451, opened 2019-11-22
Caster update

harliq:
LootFactory CreateCaster refactored. Items now created via wields not tier. Updated values for the correct wields (values are based off of data from Optim and Magnus).

PR has been tested on private server without issue.


ziang4891:
Your PR isn't really clean.  There appears to still be code changes from the Test Loot Generator PR, along with changes that are duplicated in the Missile Loot gen PR.  This one should really only have just the two files being touched, IMO.

harliq:
> 
> 
> PR isn't really presenting clean. There appears to still be code changes from the Test Loot Generator PR, along with changes that are duplicated in the Missile Loot gen PR. This one should really only have just the two files being touched, IMO. Still gives the appearance of being all inclusive...

Once the new PR is merged, then those should issues should resolve.  This is how I was instructed to do this, per Magnus and Gmriggs.

ziang4891:
OK.  I guess it is just my OCD acting up, when seeing all of those merge branch commits and incremental change commits from the get go, instead of a clean, single commit ( or very few initial commits ) PR.

Source/ACE.Database/Models/World/TreasureDeath.cs
```diff
@@ -23,5 +23,29 @@ public partial class TreasureDeath
         public int MundaneItemMaxAmount { get; set; }
         public int MundaneItemTypeSelectionChances { get; set; }
         public DateTime LastModified { get; set; }
+
```
LtRipley36706: going to need to find a different way to do this part, scaffolding will blow this out every time

gmriggs: yeah my bad, i completely forgot this was a scaffolding generated file

this method should be easily moveable to a static method in another class, or an extension method: deathTresure.Clone()

Source/ACE.Server/ACE.Server.csproj
```diff
@@ -10,6 +10,10 @@
     <ServerGarbageCollection>true</ServerGarbageCollection>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Factories\TestLootGenerationFactory.cs" />
```
gmriggs: This should not be modifying the .csproj file.

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -2250,5 +2250,160 @@ private static float GetTotalProbability(List<TreasureMaterialGroups> list)
             return totalSum;
         }
 
-    }
+        /// <summary>
+        /// Will return correct meleeMod for bow wields (some debate on what 375 top out at, leaving at 18 for now). HarliQ 11/17/19
+        /// </summary>
+        private static double GetWieldReqMeleeDMod(int wield)
+        {
+            double meleeMod = 0;
```
gmriggs: this should not have the changes from 2452 merged in

should just be TestLootGen + 2451

Source/ACE.Database/Models/World/TreasureDeath.cs
harliq: Removed

Source/ACE.Server/ACE.Server.csproj
harliq: Fixed

PR #2452, opened 2019-11-22
Missile weapon updates

harliq:
Missile Updates
If tier is 4 or 5 then a chance for non elemental missile weapon with wield reqs
Tier 6 and up, Only elemental missile weapons are supposed to drop per retail.
Added 10% variance for Damage Modifier. Full Modifier was rare in retail
Added +19 Elemental (like retail) and readjusted odds (odds are approximate, no hard data).
Made MeleeDefenseMod appropriate to Bow Reqs.


Source/ACE.Server/ACE.Server.csproj
```diff
@@ -10,6 +10,10 @@
     <ServerGarbageCollection>true</ServerGarbageCollection>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Factories\TestLootGenerationFactory.cs" />
```
gmriggs: what is this line?

harliq: Fixed

PR #2459, opened 2019-11-23
Set all Shard non-zero defaults to zero

Mag-nus:
There appears to be a bug in EF where, if the models value is not nullab,e and is 0, and the SQL columns default is not 0, the EF.SaveChanges() command will INSERT the record but omit the columns that are 0 value.

The DB will then add default values for the omitted columns.

The entire record is then returned as a result, and EF uses this to update the model, now setting those fields that were 0 to the new non-zero default.

This is the cause for the Jive Turkey emote delay issues.

The next step will be to create an update script that will set existing shard objects to their weenie defaults. The following tables will need to be investigated:
biota_properties_create_list
biota_properties_emote
biota_properties_emote_action


Database/Base/ShardBase.sql
```diff
@@ -418,13 +418,13 @@ DROP TABLE IF EXISTS `biota_properties_generator`;
...
-  `max_Create` int(10) NOT NULL DEFAULT '1' COMMENT 'Maximum amount of objects to generate',
-  `when_Create` int(10) unsigned NOT NULL DEFAULT '2' COMMENT 'When to generate the weenie object',
-  `where_Create` int(10) unsigned NOT NULL DEFAULT '4' COMMENT 'Where to generate the weenie object',
+  `init_Create` int(10) NOT NULL DEFAULT '0' COMMENT 'Number of object to generate initially',
+  `max_Create` int(10) NOT NULL DEFAULT '0' COMMENT 'Maximum amount of objects to generate',
+  `when_Create` int(10) unsigned NOT NULL DEFAULT '0' COMMENT 'When to generate the weenie object',
+  `where_Create` int(10) unsigned NOT NULL DEFAULT '0' COMMENT 'Where to generate the weenie object',
   `stack_Size` int(10) DEFAULT NULL COMMENT 'StackSize of object generated',
   `palette_Id` int(10) unsigned DEFAULT NULL COMMENT 'Palette Color of Object Generated',
   `shade` float DEFAULT NULL COMMENT 'Shade of Object generated''s Palette',
```
gmriggs: missing biota_properties_spell_book update?

gmriggs:
All of the code that creates new biotas and adds them to the database should be reviewed for these records

If the DB/EF defaults were being relied on before, the code will need to explicitly specify these defaults now

For example, if a new spellbook entry was being created at runtime, and the code was relying on the DB/EF defaults of 2, after this change those new entries would have 0. These default values would now move to becoming the code's responsibility

gmriggs:
I can add a command to fix the shard emote_action delays

emote probability is not a factor -- there is nothing with probability 0

create_list is not a factor -- create_list, emote, generator, and spell_book biotas that are saved and restored from the database are already majorly broken in other ways, specifically they are lacking an order column, and if they get written to the shard db, EF is usually writing them in reverse order / in an undefined order. So any biotas that get written/restored from the shard db for these 4 tables are already majorly broken already until the order column is added.

Database/Base/ShardBase.sql
Mag-nus: added

PR #2465, opened 2019-11-25
EquipmentSetID

ziang4891:
- Add method to apply a random EquipmentSetID to tier 7 & 8 armor, with a modifiable base line of 10% chance
- Initial introduction controlled via server property, which is set to false by default, allowing server ops to decide for themselves to enable it
- Added option to add Set name to item name


ziang4891:
Closing apparently unwanted change

PR #2488, opened 2019-12-02
Melee overdamage fix

harliq:
Fixes Melee Weapons being created that over max damage for that wield.


Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -721,7 +721,7 @@ private static int GetMeleeMaxDamage(Skill weaponType, int wieldDiff, LootWeapon
             }
 
             // To add a little bit of randomness to Max weapon damage
-            int maxDamageVariance = ThreadSafeRandom.Next(-4, 2);
+            int maxDamageVariance = ThreadSafeRandom.Next(-4, 0);
```
gmriggs: any idea where this original 6-damage variance came from?

should this be -6, 0 now?

harliq: The comment says to add variance.  But it was adding 2 over max damage.  Thats all I am removing for the moment to stop getting over max weapons.  Eventually will add variance based on wield data, but with melee weapons, its 32-36 data points per wield req.  Its a lot of research and time to get the correct variance for each melee weapon, let alone code it.

PR #2491, opened 2019-12-03
mana conversion mod tweaks

gmriggs:
The main purpose of this patch was to fix a bug where appraisal was showing Mana Conversion Bonus: +0% for a many casting items

At first I thought this was a bug in the loot generator, but after discussing it with @harliq, it turns out there are actually many casting items with base weenies that have PropertyFloat.ManaConversionMod=0 in the data

So to account for this, Appraisal is now checking if ManaConversionMod exists, but is 0, it removes it.

I also noticed some possible bugs with Hermetic Void, which sounds like it is cast on the wand, similar to the other item debuffs. The logic for ManaConversionMod enchantments should be taking both auras and spells cast directly on the item into account for this.

Update: it has been discovered from retail pcaps that retail also had this bug.

Retail also didn't normalize ManaConversionMod on the items -- some of the base weenies have 0, some of them have null. The retail pcaps had ManaConversionMod=0 in many instances (even while unbuffed), but there were also a few rare instances of ManaConversionMod not being sent.

The 'show_mana_conv_bonus_0' server option has been added, which defaults to true to match retail behavior.


Warloxx01:
Hi, I added this and the (Null mana c) to a clean server build and this happened when I was playing in Metos...
[Error.txt](https://github.com/ACEmulator/ACE/files/3923092/Error.txt)

This didnt crash my server.

gmriggs:
Hey @Warloxx01,

Thanks for the bug report!

I have checked in an update that should hopefully fix this error. Please let me know if you find anything else

ziang4891:
I can confirm that the null checks fix the exception, on appriasal

PR #2515, opened 2019-12-09
Network improvements

Mag-nus:
Disregard the commit where I was testing Fartwhifs CryptoSystem changes. That work is not in this PR.

The objective here, which is still on going

- Reduce "new byte[]"
-- At least two "new byte[]" have been reduced per client packet

- Make the ClientPacket and ServerPacket objects easier to reuse outside of ACE, in either a client or server side app
-- Move "Pack" and "Unpack" work out of the ctor and into public functions

- Clean up some of the variable and function names
-- "Pack", "Unpack", "Payload" -> "DataReader"

This PR is good to go and offers improvements over current master. There are still improvements to be had, and I will likely work on those in the near future and submit them in follow up PR's.


ziang4891:
I tried the PR out, and appears to be causing wild fluctuations of CPU and odd memory usage behavior vs current master.

ziang4891:
This PR does appear to release most, if not all, of the memory allocated by ACE released at program termination, unlike how current master has been operating.  In master, it appears that allocated memory is held for a time after the program ends.  The other difference is that the PR appears to postpone the initial memory caching, until after the caching routines declare via console output that they are completed.  ACE appears to continues to load its memory even after the server states that the world is open, causing the server to apparently initially ignore client connection attempts for a time.  This phenomenon can be observed via Resource Manager, in the graphed rise and fall of allocated memory, during the starting and stopping stages of the server.

Mag-nus:
Can you do some more testing? While this PR does look rather large, the bulk of it is renaming and code shuffling. The actual functional code changes are very small.

Thanks.

ziang4891:
Does have a conflict with master, now...

Mag-nus:
fixed! :)

ziang4891:
Never mind above.  This phenomenon isn't a result of this PR.  I just built a plain, current master, and that build is exhibiting this affect.

PR #2552, opened 2019-12-20
Magic and Missile Defense Modifier Bonus simplifcation/fix

harliq:
Made getting MagicD and MissileD modifier the same across all weapon types.  Also fixes Missile Weapons not receiving a chance of getting a MagicD modifier bonus.

Also neatened up Melee Factory to comply with StyleCop

Tested with loot sim to make sure drops are correct.

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -1390,12 +1390,12 @@ namespace ACE.Server.Factories
                         magicMissileDefenseMod = 0;
                         break;
                     case 3:
-                        int chance = ThreadSafeRandom.Next(0, 100);
+                        int chance = ThreadSafeRandom.Next(1, 100);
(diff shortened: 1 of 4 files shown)


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1373,96 +1373,101 @@ private static double GetManaCMod(int tier)
...
-                        missileMod = 0;
-                    break;
+                switch (tier)
+                {
+                    case 1:
+                    case 2:
+                        magicMissileDefenseMod = 0;
+                        break;
+                    case 3:
+                        int chance = ThreadSafeRandom.Next(0, 100);
```
gmriggs: technically this is rolling between 0-100 here, inclusive, meaning 101 possible values

should probably be 1-100

harliq: Fixed and pushed

PR #2563, opened 2019-12-22
switch BiotaPropertiesInt to Composite Key

Mag-nus:
To use ShardDatabaseWIthoutCaching (not recommended unless you really really need this due to EF tracking memory consumption issues)...

in DatabaseManager.cs
change this: var shardDb = new ShardDatabase();
to this: var shardDb = new ShardDatabaseWithoutCaching();


Mag-nus:
Maybe before this is merged into master, we should switch the other property tables to Composite key as well.

gmriggs:
testing on ce, looks good so far

Mag-nus:
#2569 should go in first.

PR #2594, opened 2020-01-06
Update default dats path on windows

Mag-nus:
This lines up with the ACE install on windows wiki. I've seen others using the same directory structure as well before this documentation update:

https://github.com/ACEmulator/ACE/wiki/ACE-Hosting-Windows


gmriggs:
This seems rather arbitrary / unnecessary, and could be seen as worse..

Mag-nus:
The idea is to sort the different assets under their own sub-directories under ACE.

This is a scalable pattern. The alternative is root becomes bloated with various assets unrelated to each other, but required by ACE.

PR #2601, opened 2020-01-10
improving /createinst command for content dev

gmriggs:


Thomasy388:
is it possible to have a /loc flag to add location with text?
May we have an option 
/createinst 51961 /loc 0x7AE80002 [18.392700 25.705500 120.005000] 0.668462 0.000000 0.000000 0.743747

-Dark Uncle

Thomasy388:
Can generator respawns rate be changed (or are all generators on world spawn) ?

gmriggs:
> is it possible to have a /loc flag to add location with text?
> May we have an option
> /createinst 51961 /loc 0x7AE80002 [18.392700 25.705500 120.005000] 0.668462 0.000000 0.000000 0.743747
> 
> -Dark Uncle

What would the use case for this be?

You're supposed to go to the location you want the object to spawn, and then /createinst

gmriggs:
> Can generator respawns rate be changed (or are all generators on world spawn) ?

Sure, but this doesn't have anything to do with /createinst or landblock instances.

You would simply go into the generate weenie, and adjust the properties for the generator.

Thomasy388:
I have tried to place a book and other things by dropping and sometimes more precision is needed

gmriggs:
> I have tried to place a book and other things by dropping and sometimes more precision is needed. It can however as you suggest done through SQL

Precision placement of objects, and moving them around afterwards, is not the intended usage of /createinst. For something like that, you would be better off with a 3d editor, and clicking and dragging items around.

If you have the /loc already, then simply /teleloc there, and place the object.

gmriggs:
btw, for future reference for precision placement of objects, some things that might be useful:

- appraise yourself, /setproperty PropertyFloat.DefaultScale 0.2, relog
- maybe give yourself a very high jumping skilll (/god?)
- /cloak on to make yourself ethereal / pass into normally unreachable areas can also be useful

PR #2603, opened 2020-01-12
remove network test helpers

fartwhif:


Mag-nus:
I found these useful. Why remove them?

fartwhif:
external tools are more effective

Yonneh0:
@Mag-nus using in-process tools to scramble packets is fundamentally flawed.

When disabled, there is still a performance impact
it currently doesn't scramble the headers
fixing those 2 would take more dev time from actual useful features, since tools like clumsy exist.

PR #2616, opened 2020-01-14
improved database persistence for biota_properties_generator

gmriggs:


Mag-nus:
Should something similar be applied to the WorldDatabase where the weenies are read?

gmriggs:
> Should something similar be applied to the WorldDatabase where the weenies are read?

Technically when a bunch of the of the weenie* tables are read in, it should be ordering by id, ensuring the records are returned in the order they were entered into the db.

The World DB does not have the same problems as the Shard DB though. The Shard DB biota tables are written incorrectly, as a result of EF writing collections / sets in its own internal way, providing no interface to specify or guarantee the ordering.

So if the question is 'should be sorting the weenie generators table by probability when it is read in', the answer is no, most definitely not. In fact, we really even shouldn't be doing it for the shard database -- it is only a hack to avoid having to place the order column on the biota_generators table. The order column would be the 'real' way to do it.

Source/ACE.Database/ShardDatabase.cs
```diff
@@ -223,7 +223,7 @@ public static Biota GetBiota(ShardDbContext context, uint id)
             if (populatedCollectionFlags.HasFlag(PopulatedCollectionFlags.BiotaPropertiesEnchantmentRegistry)) biota.BiotaPropertiesEnchantmentRegistry = context.BiotaPropertiesEnchantmentRegistry.Where(r => r.ObjectId == biota.Id).ToList();
             if (populatedCollectionFlags.HasFlag(PopulatedCollectionFlags.BiotaPropertiesEventFilter)) biota.BiotaPropertiesEventFilter = context.BiotaPropertiesEventFilter.Where(r => r.ObjectId == biota.Id).ToList();
             if (populatedCollectionFlags.HasFlag(PopulatedCollectionFlags.BiotaPropertiesFloat)) biota.BiotaPropertiesFloat = context.BiotaPropertiesFloat.Where(r => r.ObjectId == biota.Id).ToList();
-            if (populatedCollectionFlags.HasFlag(PopulatedCollectionFlags.BiotaPropertiesGenerator)) biota.BiotaPropertiesGenerator = context.BiotaPropertiesGenerator.Where(r => r.ObjectId == biota.Id).ToList();
+            if (populatedCollectionFlags.HasFlag(PopulatedCollectionFlags.BiotaPropertiesGenerator)) biota.BiotaPropertiesGenerator = context.BiotaPropertiesGenerator.Where(r => r.ObjectId == biota.Id).OrderBy(r => r.Probability).ToList();
```
Mag-nus, line 226: Your reasoning makes sense, and this change should work.

I'd like to see a comment above this line though. It should describe why we're doing the OrderBy here. It would also act as an indicator to note that this line is different among all the other queries.

In the future, if we do add an order column to this table, we could get caught out by not realizing this band-aid exists.

PR #2625, opened 2020-01-17
Switch delayed LogOut_Inner due to PKLogoutActive to PlayerManager.Tick

LtRipley36706:


Mag-nus:
Instead, should log out be delayed until the player is back on a landblock?

gmriggs:
> Instead, should log out be delayed until the player is back on a landblock?

They never return to the landblock... that's the bug

CurrentLandblock is somehow going null for the player, even though the Landblock.worldObjects list still contains references to the player

Mag-nus:
I'm still thinking that this should be a function of the Session, not PlayerManager.

LtRipley36706:
Doesn't matter where it is, as long as it can't go null and can tick without being "in game" to completion

PR #2631, opened 2020-01-20
Loot gen changes

ziang4891:
- Add text descriptions to EquipmentSet enum
- Add option (modifybool equipmentsetid_enabled) to enable assigning EquipmentSet IDs to top tier armor loot and is set to disabled by default, for now
- Add option (modifybool equipmentsetid_name_decoration) to prefix EquipmentSet name to item name and is set to disabled by default
- Re-align the loot profiles to be **generic** items, which are non-mutable items, **mundane** (mutable items with no magic), and **magical**, mutable items
- Move Aetheria drop to mundane loot profile ( https://asheron.fandom.com/wiki/Announcements_-_2010/04_-_Shedding_Skin :: May 5th, 2010 entry )
- Separate out Dinnerware from generic items
- Update loot biasing to support Legendary Chest loot profiles
- Add passing whole TreasureDeath profile object into various LootGen methods, instead of merely the tier component
- Add usage of LootQualityMod from TreasureDeath profile data to increase cantrip likelyhood
- Add updated AL assignment method, which will be used by updated weenies per yotes reconciliation (http://ac.yotesfan.com/json/validate.php), and current AL assignment method switched to a compatibility function for databases without the armor weenie updates
- Make use of switch expressions vs. switch/case statement blocks
- Correct warning messages on pattern matching switch statement blocks


Source/ACE.Entity/Enum/EquipmentSet.cs
```diff
@@ -16,22 +18,39 @@ public enum EquipmentSet
         ArmMindHeart                     = 10,
         ArmorPerfectLight                = 11,
         ArmorPerfectLight2               = 12,
+        [Description("Soldier's")]
```
gmriggs: should not be using reflection lookups here
reflection lookups are slow, and this pattern should be avoided

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -62,40 +62,26 @@ public static List<WorldObject> CreateRandomLootObjects(TreasureDeath profile)
 
             switch (profile.TreasureType)
             {
-                case 1001:  // Mana Forge Chest, Advanced Equipment Chest, and Mixed Equipment Chest
+                case var n when (n == 1001 || n == 2001):  // Mana Forge Chest, Advanced Equipment Chest, and Mixed Equipment Chest
```
gmriggs: case 1001:
case 2001:

Source/ACE.Entity/Enum/EquipmentSet.cs
ziang4891: Might have to see if there is a better way to insert the apostrophe into the name of the first six enum names

ziang4891: removed; solved with string manipulation

gmriggs:
for the actual lootgen parts, @harliq would probably be the best person to review that

harliq:
> 
> 
> for the actual lootgen parts, @harliq would probably be the best person to review that

I will pull it and run it through loot sim tonight

harliq:
I pulled and tested with @ziang4891 updated weenies.  Loot Sim shows correct armor levels, and weapons appear to be fine.

ziang4891:
Adjustments made to the loot drop rates to better approximate those found in current master

ziang4891:
Messed up the conflict resolution.  Closing and will re-build in clean master.

PR #2634, opened 2020-01-20
Upped Melee D Bonus for no wield wands upto 10% from 5%. 

harliq:
Upped Melee D Bonus for no wield wands up to 10% from 5%.  That is what data supports from Optim.

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory.cs
@@ -2266,6 +2266,27 @@ namespace ACE.Server.Factories
             switch (wield)
             {
                 case 0:
+                    if (chance < 20)
+                        meleeMod = 0.01;
+                    else if (chance < 30)
+                        meleeMod = 0.02;
+                    else if (chance < 40)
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -2266,6 +2266,27 @@ private static double GetWieldReqMeleeDMod(int wield)
             switch (wield)
             {
                 case 0:
+                    if (chance < 20)
```
gmriggs: with the way these are written, the first one only has 19% chance, all of the middle ones have a 20% chance, and the last one has a 6% chance

harliq: there isn't a a good reference on what the drop rate is, just want the top 3 to have a lower chance.

gmriggs: what i'm saying is, with the way the code is written, it looks like the intention was for the first one to be 20%, and the last one to be 5%

the first and last entries have off-by-1 errors

PR #2654, opened 2020-01-25
syncing up emotes for scream and newenemy

gmriggs:


gmriggs:
ideally would like to get dictionary / fast lookup support into EmoteManager collections before merging this

LtRipley36706:
current master good enough for this or needs to be #2697?

PR #2668, opened 2020-01-29
Move OnHomeSick call to MoveToHome

LtRipley36706:


DridsGhost:
Tested. It seems work pretty consistent for the issue I had, with the local signaling. Does seem to kick the local signal twice, but the doors can at least hear it now and act accordingly.

PR #2683, opened 2020-02-03
adding landblock instance support to import-sql and import-json

gmriggs:


DridsGhost:
I have tested and this works very well. The only caveat is for the "key" at the top of the JSON file it creates when exporting from sql, it used either the file name, or just the 0x8765 where as GDLE requires the file to be converted from 0x87650000.  So you have to go back and edit the JSON version of the landblocks.

PR #2691, opened 2020-02-05
Step 1. Add new adapter classes and models

Mag-nus:
All classes and objects added are in new namespaces.

This add has 0 interraction on existing ACE code.

This sets the groundwork for Step 2. Moving World DB caching to the new models

### DIFF
--- a/Source/ACE.Database/Adapter/BiotaConverter.cs
+++ b/Source/ACE.Database/Adapter/BiotaConverter.cs
@@ -1,0 +1,890 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+
+using ACE.Database.Models.Shard;
+using ACE.Entity.Enum;
+using ACE.Entity.Enum.Properties;
+using ACE.Entity.Models;
+
+namespace ACE.Database.Adapter
+{
+    public static class BiotaConverter
+    {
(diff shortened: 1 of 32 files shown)


Mag-nus:
It was just an easy way to go from what we have to the model system.

If you'd like the names changed, please suggest new names in the following format (or similar)
OldName -> NewName

gmriggs:
my suggestion is that only the things that are actually properties in ac should be listed as properties:

https://acemulator.github.io/protocol/Types/ACBaseQualities.html

ie., all the PropertyTypes, where types are primitive types

EnchantmentRegistry it not a property. it should be just EnchantmentRegistry, not PropertyEnchantmentRegistry

gmriggs:
technically it shouldn't even be ICollection<EnchantmentRegistry>, it should be ICollection<Enchantment>

Mag-nus:
This needs to be thought through more.

You'll end up with some properties that become BodyPart, Emotes (EmoteSet -> Emote), AnimPart (which already has a simliar named AnimData in entity)...

And then you get to things like Attribute and Attribute2nd, Skill, etc.. where if you remove the Properties prefix, you can get collisions with other enums or entity objects that are similar, but not the same.

I would suggest waiting to switch the names and/or theme of the names until after this is all pushed into master.

Trying to switch the names now adds another layer of confusion and room for error.

PR #2694, opened 2020-02-06
adding quest support to import-json and import-sql

gmriggs:


DridsGhost:
As discussed when converting two single quotes used for the message field SQL insert to JSON, it leaves  two single quotes. Works great when going from JSON to SQL. 

It will add the extra single quote to your SQL to avoid INSERT problems with the message field.

PR #2695, opened 2020-02-06
Loot gen changes

ziang4891:
- Add text descriptions to EquipmentSet enum
- Add option (modifybool equipmentsetid_enabled) to enable assigning EquipmentSet IDs to top tier armor loot and is set to disabled by default, for now
- Add option (modifybool equipmentsetid_name_decoration) to prefix EquipmentSet name to item name and is set to disabled by default
- Move Aetheria drop to mundane loot profile ( https://asheron.fandom.com/wiki/Announcements_-_2010/04_-_Shedding_Skin :: May 5th, 2010 entry )
- Separate out Dinnerware from generic items
- Update loot biasing to support Legendary Chest loot profiles
- Add passing whole TreasureDeath profile object into various LootGen methods, instead of merely the tier component
- Add usage of LootQualityMod from TreasureDeath profile data to increase cantrip likelyhood
- Add updated AL assignment method, which will be used by updated weenies per yotes reconciliation (http://ac.yotesfan.com/json/validate.php), and current AL assignment method switched to a compatibility function for databases without the armor weenie updates
- Make use of switch expressions vs. switch/case statement blocks
- Correct warning messages on pattern matching switch statement blocks


ziang4891:
Recreated from a clean backup of origin PR, against current master.  While giving it a once-over, I found and corrected one unintended change.

ziang4891:
Someone else will need to push, as I cannot

PR #2697, opened 2020-02-07
Change EmoteManager to use dictionary lookup for Emotes

LtRipley36706:


Mag-nus:
The changes in ExecuteEmote are a step backwards.

Prior, it was utilizing a linq query which was very efficient. The only issue was it would enumerate over more records when not using a keyed dictionary.

Use linq to build the query, and the return emoteSet.FirstOrDefault() will perform the work and return only what's needed (a single element, or null), without creating any additional lists.

Mag-nus:
Also, I still stand by my statements that this will provide no measurable performance improvement.

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -1367,30 +1397,35 @@ public float ExecuteEmote(BiotaPropertiesEmote emoteSet, BiotaPropertiesEmoteAct
...
+
+            if (!hasValidEmoteSets)
+                return null;
 
             // optional criteria
             if (questName != null)
-                emoteSet = emoteSet.Where(e => e.Quest.Equals(questName, StringComparison.OrdinalIgnoreCase));
-            if (vendorType != null)
-                emoteSet = emoteSet.Where(e => e.VendorType != null && e.VendorType.Value == (uint)vendorType);
+                emoteSet = emoteSet.Where(e => e.Quest.Equals(questName, StringComparison.OrdinalIgnoreCase)).ToList();
```
gmriggs: this shoudn't be casting .ToList() every single clause

this is effectively turning 1 query into 4-5 separate queries

Source/ACE.Server/WorldObjects/Managers/EmoteManager.cs
```diff
@@ -47,6 +47,39 @@ public class EmoteManager
...
             _worldObject = worldObject;
+            InitEmoteCaches();
+        }
+
+        private Dictionary<EmoteCategory, IEnumerable<BiotaPropertiesEmote>> emoteCache = new Dictionary<EmoteCategory, IEnumerable<BiotaPropertiesEmote>>();
+        private Dictionary<VendorType, IEnumerable<BiotaPropertiesEmote>> emoteCacheVendor = new Dictionary<VendorType, IEnumerable<BiotaPropertiesEmote>>();
+
+        private void InitEmoteCaches()
+        {
+            emoteCache.Clear();
```
Mag-nus: I think you also want a emoteCacheVendor .Clear() here as well.

Mag-nus:
I'm testing this against current master (that has the fixed orderby), and am not seeing a noticeable difference with the following setup:
Landblocks: 1,415 active (3 dungeons), 635 dormant (1 dungeons), Landblock Groups: 14 - Players: 154, Creatures: 17,053, Missiles: 10, Other: 24,439, Total: 41,656

I still plan to do more controlled testing. Will report those results when I have them.

PR #2700, opened 2020-02-11
adding z-bump to scatter gen

gmriggs:


DridsGhost:
I've tested and confirmed this updates resolves my issue at Neftet plateaus.

PR #2707, opened 2020-02-12
Updating project files

LtRipley36706:


Mag-nus:
I would still rename /art/ to /images/, or, alternatively put it in the repo root as /images/.

Other projects may want to references these assets. More generically you'd call these "resources" and they'd be in a folder with the same name, however, i think a specific folder for /images/ is appropriate.

I've never seen the term /art/ used to define such assets so it feels odd to me.

LtRipley36706:
moving to top level solution folder looked to be problematic in project file. maybe can revisit later

PR #2714, opened 2020-02-17
Update OnActivate for deep-linked object activation propagation

LtRipley36706:


DridsGhost:
This seems to work correctly. Tested with 3 doors and a switch.

PR #2721, opened 2020-02-19
adding reload-landblock command for content devs

gmriggs:


DridsGhost:
Seems to work very well. I use it all the time now. Its especially handy after importing a large volume of statically place weenies and not having to restart your server to try changes.

PR #2727, opened 2020-02-21
update AdjustCells to ConcurrentDictionary

gmriggs:


Mag-nus:
Please hold off on this.

The exception you posted is a single threaded stack, direct from WorldManager.

Rather than add the increased overhead from a ConcurrentDictionary, I'd like to find out what is actually messing with AdjustCells in a different thread.

Also, this is the first time I've ever seen this exception as a result of multi-threading. I spent a lot of time making sure that the various threads and entities (like landblock) can be thread safe without the use of heavier objects like ConcurrentDictionary and excessive locking. I think the current stability of ACE in mulit-thread env is pretty solid. I'd like to know more about this specific issue before we apply a patch that may be masking a bigger architectural issue that I may have missed.

I'm also not ruling out something specific to the server in questions codebase. As I mentioned, this is the first time I've seen this exception.

Mag-nus (approved):

PR #2728, opened 2020-02-21
fix acclient skipping over ethereal animation hooks for known doors that are opened > 96 distance

gmriggs:


DridsGhost:
Tested around 8 or 9 times. Seems to work without fail.

gmriggs:
For a good example between the 1st and 2nd version of this patch:

In first version of patch:

- start running constantly into a door
- open door while running into it

expected:

to match up with acclient, the doors to the buildings in yaraq for example, have a ~0.3s delay before they turn ethereal

actual:

the player previously started to run through the doors immediately, in the first version of this patch.

gmriggs:
nevermind, i misinterpreted the data.

the yaraq door is not 0.3s, it is 0.03s -- the hook to change ethereal state is on the second frame of the animation. so ethereal is set to false 1/30 of a second after opening the door, and it is set to true 29/30 of a second after closing the door

in other words, at least for the doors i am testing, the original logic was "close enough", and this would be overcomplicating things, unless there are truly any doors where the ethereal hook happens 1/2 way through the animation.

PR #2731, opened 2020-02-22
Step 3. Convert ShardDatabase/WorldObjects biota from the ACE.Database model to the new ACE.Entity model

Mag-nus:
This builds on #2702 , when that is merged in, this PR will show less changes.

* This is a major change and will require significant testing.

This switches ACE to use a new Biota model that uses dictionaries instead of collections of records. Dictionaries and lists are used to improve lookup performance.

This provides a significant improvement in RAM consumption and Garbage Collection load, and a minor improvement on CPU consumption.

There are further improvements to be made based on this framework. Those will be done in a follow up PR. Some examples:
* Use a hybrid weenie/biota approach for non-database entities to reduce duplication of collections like Emotes and CreateLists.
* Improve EnchantmentManager to use a single lock to perform certain queries, and to do so without generating temporary lists (.ToList()).


gmriggs:
running on ce past few days, looks good

PR #2752, opened 2020-03-01
Pet devices levels tiers

harliq:
Adding a spread of Pet Device levels for each tier - Level 200 pets should only be dropping in T8 Loot - HQ 2/29/2020
The spread is from Optim's Data
T5-T8 20/35/30/15% split 
T8- 200,180,150,125
T7- 180,150,125,100
T6- 150,125,100,80
T5- 125,100,80,50
T4- 100,80,50
T3- 80,50
T2- 50
T1- 50


Source/ACE.Server/Factories/LootGenerationFactory_Test.cs
```diff
@@ -249,8 +249,10 @@ public static LootStats LootStats(WorldObject wo, LootStats ls, bool logstats)
                                 petLevel = 100;
                             else if (petDevice.UseRequiresSkillLevel == 370)
                                 petLevel = 80;
-                            else if (petDevice.UseRequiresSkillLevel == 310)
+                            else if (petDevice.UseRequiresSkillLevel == 310 || petDevice.UseRequiresSkillLevel == 320)
                                 petLevel = 50;
+                            if (petLevel == 0)
+                                Console.WriteLine("Skill= " + petDevice.UseRequiresSkillLevel);
```
gmriggs: debug code?

harliq: Some PetDevice weenies do not have the correct wield reqs, thats why its there

gmriggs: shouldn't that be using log.Debug then, and printing out a full error message?

if Skill=# gets output to the console, it will be unclear what that indicates

gmriggs:
another question: should the function that rolls for pet ratings still be based on tier?

if a T5 rolls a level 50 pet, it doesn't make much sense for it to have lower ratings on average than a T1 rolling that same level 50 pet

PR #2760, opened 2020-03-03
Fix Alt Currency minimum cost

LtRipley36706:


gmriggs:
Not sure if this fix is appropriate.

This leaves the potential for bad data lingering around that never gets detected or fixed..

PR #2776, opened 2020-03-06
LootFactory Melee refactor

harliq:
Refactoring, removing all the SetProperties References, adding Notes & simplifying


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -902,9 +901,8 @@ private static double GetMaxDamageMod(int tier, int maxDamageMod)
                 default:
                     break;
             }
-            double damageMod2 = 1.0 + damageMod;
 
-            return damageMod2;
+            return damageMod + 1;
```
gmriggs: nbd, but since damageMod is an additive float multiplier, the 1.0 + value format seemed clearer imo

PR #2809, opened 2020-03-12
[TEST] network relogin

gmriggs:


gmriggs:
Update:

With https://github.com/ACEmulator/ACE/pull/2809/commits/71c27feee83562330b667028ef1d16d0035cf405 , I am still able to repro the 'client repeatedly tries to log into world very rapidly' bug using a new set of repro steps:

- Start server
- Log into server via thwarg launcher. Close thwarg launcher window so that it doesn't automatically close the ac window
- Open thwarg launcher again
- Set /pktimer on char still logged into world
- Log in again with thwarg launcher, WITHOUT closing the original acclient window

At this point, since ThwargLauncher will be connecting on a new port, it will get a new session, instead of the existing one. So the new session won't be in 'WorldConnected' state, so it won't hit any of that logic

It will however, hit this logic in AuthenticationHandler, on line 176:

```
                if (PropertyManager.GetBool("account_login_boots_in_use").Item)
                {
                    var previouslyConnectedAccount = NetworkManager.Find(account.AccountName);

                    if (previouslyConnectedAccount != null)
                    {
                        previouslyConnectedAccount.Terminate(SessionTerminationReason.AccountLoggedIn, new GameMessageCharacterError(CharacterError.Logon));
                    }
                }
```

Thinking maybe this logic should be moved to NetworkManager.LoginRequest perhaps?

gmriggs:
With the revised implementation in https://github.com/ACEmulator/ACE/pull/2809/commits/593cc5f06a1e63a9062e6f256b3234d8da01b146, the previous bug is fixed

PR #2822, opened 2020-03-15
improved collision detection for projectiles on players

gmriggs:


gmriggs:
This patch has been running on Potato for the past few weeks

I have not heard any reports from players on whether they have perceived any noticeable improvements from this patch -- however, the 2-way collision detects definitely seem like the "right" way to do this part here

PR #2828, opened 2020-03-17
[WIP -Debug] debugging bad weenie scrolls

harliq:
For catching bad scrolls/weenies

### DIFF
--- a/Source/ACE.Database/Adapter/WeenieConverter.cs
+++ b/Source/ACE.Database/Adapter/WeenieConverter.cs
@@ -2,7 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
-
+using log4net;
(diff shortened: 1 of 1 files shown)


gmriggs:
I forwarded this particular error to @Mag-nus, so that he can investigate the best place to insert such error handling and logging code

I think perhaps the inner database lookup code should just be returning null properly / not crashing there, and then it would maybe be the responsibility of the outer game code that fetches the weenie to check if null / handle / error logging

harliq:
I agree, this just for DnF at the moment for helping find issues in their stuff.  Not really meant for master

Mag-nus:
I don't think this is correct.

I believe adapter was working correctly. The null check should be going higher up in the chain, or somewhere else.

harliq:
no need for this PR, closing

harliq:
close

PR #2846, opened 2020-03-22
adding null check to GetScrollWeenie

gmriggs:


gmriggs:
What happens if I import the scroll to my live server with the import content commands?

The weenie cache will be cleared, but the scroll cache will not be... so the scroll cache will still be saying that scroll is 'null'

Mag-nus:
I think clearweeniecache should probably be expanded to be "clearworldcache". If not in name, at least functionally.

I'm fine either way, just bring up something I saw.

PR #2847, opened 2020-03-23
adding null check

gmriggs:
This seems like it should be impossible, but there is a rare crash in Player_Inventory.cs on line 517

No idea how GetKnownObjects() could be returning null


gmriggs:
This caused a stack trace, but not an actual server crash

Closing this PR.

PR #2870, opened 2020-03-29
fixing shade=0, from optimshi

gmriggs:


OptimShi:
There's several other instances of `Shade > 0`, not just `Item.Shade > 0`. Not sure if all are relevant (e.g. where it's used as "probability"), but the others correct some other issues.

Test: Create an NPC with a Faran Robe, WCID 5850, PaletteTemplate 5 and Shade of 0. The created NPC with robe will use the default Shade on the Robe of 0.5 (should be a dark blue/black trim, but is instead a medium blue with brown/greyish trim)

PR #2874, opened 2020-04-02
Added /teleplayerloc Command to teleport a specific player to a specific location

harliq:
Teleport a specific character to a specific location.
```
/teleplayerloc <player name> <loc>
```
Usage
```
@teleplayerloc follows the same number order as displayed from @loc output\n
@teleplayerloc <player name> 0x7F0401AD [12.319900 -28.482000 0.005000] -0.338946 0.000000 0.000000 -0.940806
Example: @teleplayerloc <player name> 0x7F0401AD 12.319900 -28.482000 0.005000 -0.338946 0.000000 0.000000 -0.940806
Example: @teleplayerloc <player name> 7F0401AD 12.319900 - 28.482000 0.005000")]
```

### DIFF
--- a/Source/ACE.Server/Command/Handlers/AdminCommands.cs
+++ b/Source/ACE.Server/Command/Handlers/AdminCommands.cs
@@ -1084,6 +1084,64 @@ namespace ACE.Server.Command.Handlers
             }
         }
 
+
+        // teleplayerloc <player name> cell x y z [qx qy qz qw]
+        [CommandHandler("teleplayerloc", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 4,
+            "Teleports a player to the specified location.",
+            "cell [x y z] (qw qx qy qz)\n" +
+            "@teleplayerloc follows the same number order as displayed from @loc output\n" +
+            "Example: @teleplayerloc <player name> 0x7F0401AD [12.319900 -28.482000 0.005000] -0.338946 0.000000 0.000000 -0.940806\n" +
+            "Example: @teleplayerloc <player name> 0x7F0401AD 12.319900 -28.482000 0.005000 -0.338946 0.000000 0.000000 -0.940806\n" +
+            "Example: @teleplayerloc <player name> 7F0401AD 12.319900 - 28.482000 0.005000")]
+        public static void HandleTeleportPlayerLOC(Session session, params string[] parameters)
+        {
+
+            var playerName = string.Join(" ", parameters[0]);
+            var player = PlayerManager.GetOnlinePlayer(playerName);
+            if (player == null)
+            {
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1084,6 +1084,60 @@ public static void HandleTeleportLOC(Session session, params string[] parameters
...
+        [CommandHandler("teleplayerloc", AccessLevel.Developer, CommandHandlerFlag.RequiresWorld, 4,
+            "Teleports a player to the specified location.",
+            "cell [x y z] (qw qx qy qz)\n" +
+            "@teleplayerloc follows the same number order as displayed from @loc output\n" +
+            "Example: @teleplayerloc <player name> 0x7F0401AD [12.319900 -28.482000 0.005000] -0.338946 0.000000 0.000000 -0.940806\n" +
+            "Example: @teleplayerloc <player name> 0x7F0401AD 12.319900 -28.482000 0.005000 -0.338946 0.000000 0.000000 -0.940806\n" +
+            "Example: @teleplayerloc <player name> 7F0401AD 12.319900 - 28.482000 0.005000")]
+        public static void HandleTeleportPlayerLOC(Session session, params string[] parameters)
+        {
+            var playerName = string.Join(" ", parameters[0]);
```
gmriggs, line 1097: this will not work for players w/ spaces in their names

should be just parameters, without the [0]

Source/ACE.Server/Command/Handlers/AdminCommands.cs
```diff
@@ -1084,6 +1084,60 @@ public static void HandleTeleportLOC(Session session, params string[] parameters
...
+                    string strippedcell = parameters[1].Substring(2);
+                    cell = (uint)int.Parse(strippedcell, System.Globalization.NumberStyles.HexNumber);
+                }
+                else
+                    cell = (uint)int.Parse(parameters[1], System.Globalization.NumberStyles.HexNumber);
+
+                var positionData = new float[8];
+                for (uint i = 1u; i < 8u; i++)
+                {
+                    if (!float.TryParse(parameters[i + 1].Trim(new Char[] { ' ', '[', ']' }), out var position))
```
gmriggs, line 1121: not sure why this was modified from the original version, but this is very non-standard and contains a few bugs:

- there is no reason to create a 9-slot array for something that requires 8-slots, just to use 1-based indexing. this is very non-standard, and leads to confusion
- float.TryParse is still using + 1 indexing on top of the 1-base, so it starts parsing the location parameters at parameters[2], instead of the intended parameters[1]
- this indexing method doesn't work with characters w/ spaces in their names

PR #2875, opened 2020-04-03
Destroy Laggy Projectiles

Mag-nus:


gmriggs:
Projectiles shouldn't just randomly explode because of server lag, or if 1 tick seems to randomly take too long...

I get these random 'tick took ## ms' messages on my local server all the time. This would severely destabilize the game..

PR #2903, opened 2020-04-10
Network layer bug fix, reduction

fartwhif:
fixed bug allowing a corrupt or malformed packet to crash the server when the fragment parser tries to read past the end of the data
simplified `CryptoSystem`
removed unused counter from `ISAAC`
simplified client packet CRC verification call stack

Tested with full Clumsy suite enabled: passed
Tested with Coldeve, currently active in Coldeve, current pop: 600

_Credit goes to Yonneh for simplified `CryptoSystem`_


gmriggs:
Can confirm this patch has been running on CE the past couple days, no issues reported with it

PR #2905, opened 2020-04-10
Fix ungod failing on SkillAdvancementClass

decaprime:
Currently if you `/god` and you can not `/ungod` you'll hit:

`Exception ( System.Private.CoreLib - Input string was not in a correct format. ) caught while {player} was attempting to return to normal from godmode.` Due to the godstate string having the string value of SkillAdvancementClass.

This PR serializes as uint, but also handles encountering the string in order to restore characters that have `/god` in the past week.


decaprime:
# Validation
1. On clean server, on master I created a character and `/god` and confirmed I was stuck.
2. Shutdown server, switched to this pull request, build and relaunched the server.
3. Ran `/ungod` -> `/god` -> `/ungod`

Result:
> You have returned from your godly state.
> You are now a god!!!
> You have returned from your godly state.

PR #2919, opened 2020-04-12
Added spell_duration_multiplier for global duration multiplier of all…

ahricvirindi:
Added spell_duration_multiplier for global duration multiplier of all spells, positive or negative, before augmentation application.  Modeled after xp_multiplier.


gmriggs:
Any particular reason this increases the debuff timers?

I can imagine there being a use case where an admin would want to increase the buff timers, but keep the debuff timers

ahricvirindi:
I thought it would help for there to still be a matching detriment to a global extension of spell duration from a balance perspective.  I think it would up the need for players to be more prepared, say for getting dispel gems or chorizite, for certain activities.  Then on a PvP server it would potentially be seen as necessary since in that relative context a "negative" spell cast onto players B is a "positive" spell person A who cast it.

PR #2935, opened 2020-04-14
prevent ninja looting chests

gmriggs:
Scenario: another player is standing next to a locked chest

You run up to the chest, and unlock it

The other player immediately opens the chest, looting all of the items in it

Retail supposedly gave the unlocker a 5-10 second window where they had exclusive access to first opening the chest.

Originally tested with a 5s window, but it ended up feeling too quick when testing in-game. Bumped it up to 10s, and it feels much more reasonable.

If players do want to prevent this delay, they can always open the chest, and then immediately close it, at which point other players can then open it immediately.


jkurs:
> If players do want to prevent this delay, they can always open the chest, and then immediately close it, at which point other players can then open it immediately.

Not sure how relevant this is but MFK chests always reset to locked upon closing so this doesn't prevent the delay as its not possible to ever bypass it with chests like those that reset upon closure.

I might suggest adding this as a server option, so that server operators can tweak this setting higher or lower depending on preference. You said 5 seconds is too low but that might be accurate to others. Just my 2cents.

Source/ACE.Server/WorldObjects/Chest.cs
```diff
@@ -82,6 +83,12 @@ private void SetEphemeralValues()
         protected static readonly Motion motionOpen = new Motion(MotionStance.NonCombat, MotionCommand.On);
         protected static readonly Motion motionClosed = new Motion(MotionStance.NonCombat, MotionCommand.Off);
 
+        /// <summary>
+        /// The chest unlocker is given exclusive access to first opening the chest
+        /// for this number of seconds
+        /// </summary>
+        public static readonly float UseLockThreshold = 10.0f;
+
```
LtRipley36706, line 91:
```suggestion
```

PR #2939, opened 2020-04-16
Update GetHouse

LtRipley36706:
Fixes issue with callback and landblock that hasn't completed load of statics

### DIFF
--- a/Source/ACE.Server/Managers/HouseManager.cs
+++ b/Source/ACE.Server/Managers/HouseManager.cs
@@ -721,8 +721,6 @@ namespace ACE.Server.Managers
                 var houseBiota = House.Load(houseGuid);
 
                 RegisterCallback(houseBiota, callback);
-
-                return;
             }
             else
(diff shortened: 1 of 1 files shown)


gmriggs:
Can you describe the repro steps for this, and the bug it fixes in more detail?

Source/ACE.Server/Managers/HouseManager.cs
```diff
@@ -716,6 +716,14 @@ public static void GetHouse(uint houseGuid, Action<House> callback)
                 else
                     callback(house);
             }
+            else if (!loaded.CreateWorldObjectsCompleted)
+            {
+                var houseBiota = House.Load(houseGuid);
+
+                RegisterCallback(houseBiota, callback);
+
+                return;
```
gmriggs, line 725: return unneeded

LtRipley36706, line 725:
```suggestion
```

PR #2944, opened 2020-04-18
Loot missile refactor

harliq:


Source/ACE.Server/Factories/LootGenerationFactory_Missile.cs
```diff
@@ -29,63 +33,68 @@ public static WorldObject CreateMissileWeapon(TreasureDeath profile, bool isMagi
...
+                wo.ItemSpellcraft = null;
+                wo.ItemDifficulty = null;
+                wo.ManaRate = null;
             }
 
+            // Material/Value/Color
             double materialMod = LootTables.getMaterialValueModifier(wo);
             double gemMaterialMod = LootTables.getGemMaterialValueModifier(wo);
-            var value = GetValue(profile.Tier, workmanship, gemMaterialMod, materialMod);
+            var value = GetValue(profile.Tier, GetWorkmanship(profile.Tier), gemMaterialMod, materialMod);
```
gmriggs, line 87: this should be using wo.Workmanship.Value

If it calls GetWorkmanship(tier) again, it will be doing a separate rng roll / getting a possibly different workmanship value than what was previously rolled for the item

harliq: Good Catch, Fixed.

PR #2945, opened 2020-04-18
Update CheckUseRequirements for Portals

LtRipley36706:


Source/ACE.Server/WorldObjects/Portal.cs
```diff
@@ -134,6 +134,77 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
...
                 }
+
+                var playerPkLevel = player.PkLevel;
+
+                if (PropertyManager.GetBool("pk_server").Item)
+                    playerPkLevel = PKLevel.PK;
+                else if (PropertyManager.GetBool("pkl_server").Item)
+                    playerPkLevel = PKLevel.PKLite;
+
+                if ((PortalRestrictions & PortalBitmask.Undef) != 0)
```
gmriggs, line 145: This line doesn't make much sense as written. You are doing something & 0 != 0, which is impossible with a standard bitmask... the only way this might be possible is if PortalRestrictions == null, in which case, you should just be writing PortalRestrictions == null

Source/ACE.Server/WorldObjects/Portal.cs
```diff
@@ -134,6 +134,77 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
...
+                {
+                    // Players may not interact with that portal.
+                    return new ActivationResult(new GameEventWeenieError(player.Session, WeenieError.PlayersMayNotUsePortal));
+                }
+
+                if ((PortalRestrictions & PortalBitmask.NoPk) != 0 && (playerPkLevel == PKLevel.PK || player.PlayerKillerStatus == PlayerKillerStatus.PK))
```
gmriggs, line 151: I would also recommend changing PortalRestrictions to be the proper PortalBitmask enum type, instead of a nullable int. The default should be Undef

These should be using PortalRestrictions.HasFlag(PortalBitmask.NoNpk) etc.

Source/ACE.Server/WorldObjects/Portal.cs
```diff
@@ -134,6 +134,77 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
...
+                }
+
+                if ((PortalRestrictions & PortalBitmask.NoNPK) != 0 && playerPkLevel == PKLevel.NPK)
+                {
+                    // Non-player killers may not interact with that portal!
+                    return new ActivationResult(new GameEventWeenieError(player.Session, WeenieError.NonPKsMayNotUsePortal));
+                }
+
+                if ((PortalRestrictions & PortalBitmask.OnlyOlthoiPCs) != 0
+                    && player.WeenieClassId != 43480 && player.WeenieClassId != 43481
```
gmriggs, line 170: These should either be using (int)WeenieClassName.ProperType, or there should be a comment about what exactly '43480' etc. is at the bare minimum

Source/ACE.Server/WorldObjects/Portal.cs
```diff
@@ -134,6 +134,77 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
...
+                    return new ActivationResult(new GameEventWeenieError(player.Session, WeenieError.YouMustBeTwoWeeksOldToUsePortal));
+                }
+
+                if (player.AccountRequirements < AccountRequirements)
+                {
+                    // You must purchase Asheron's Call -- Throne of Destiny to use this portal.
+                    return new ActivationResult(new GameEventWeenieError(player.Session, WeenieError.MustPurchaseThroneOfDestinyToUsePortal));
+                }
+
+                if (AdvocateQuest ?? false && !player.IsAdvocate)
```
gmriggs, line 203: Watch out with null coalescing and order of operations. This statement is incorrectly written, and does not do what is intended

Should be (AdvocateQuest ?? false) && !playerIsAdvocate

Source/ACE.Server/WorldObjects/Portal.cs
LtRipley36706:
```cs
public PortalBitmask PortalRestrictions
        {
            get => (PortalBitmask)(GetProperty(PropertyInt.PortalBitmask) ?? (int)PortalBitmask.Unrestricted);
            set { if (value == PortalBitmask.Undef) RemoveProperty(PropertyInt.PortalBitmask); else SetProperty(PropertyInt.PortalBitmask, (int)value); }
        }
```

Source/ACE.Server/WorldObjects/Portal.cs
```diff
@@ -142,53 +142,53 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
...
                 {
                     // Non-player killers may not interact with that portal!
                     return new ActivationResult(new GameEventWeenieError(player.Session, WeenieError.NonPKsMayNotUsePortal));
                 }
 
-                if ((PortalRestrictions & PortalBitmask.OnlyOlthoiPCs) != 0
-                    && player.WeenieClassId != 43480 && player.WeenieClassId != 43481
-                    && player.WeenieClassId != 43493 && player.WeenieClassId != 43494)
+                if (PortalRestrictions.HasFlag(PortalBitmask.OnlyOlthoiPCs)
+                    && !player.WeenieClassName.Equals("olthoiplayer", System.StringComparison.OrdinalIgnoreCase) && !player.WeenieClassName.Equals("olthoiacidplayer", System.StringComparison.OrdinalIgnoreCase)
```
gmriggs, line 170: er not exactly what i mean...

string search is much slower than comparing wcids

are these not available in WeenieClassName enum? if not, just comment them

PR #2952, opened 2020-04-19
Update Fellowship messaging + better FellowPortalSending handling

LtRipley36706:


Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -78,22 +78,30 @@ public void AddFellowshipMember(Player inviter, Player newMember)
 
             if (IsLocked && !LockedMembers.ContainsKey(newMember.Guid.Full))
             {
-                inviter.Session.Network.EnqueueSend(new GameMessageSystemChat("Fellowship is locked", ChatMessageType.Fellowship));
+                inviter.Session.Network.EnqueueSend(new GameEventWeenieErrorWithString(inviter.Session, WeenieErrorWithString.LockedFellowshipCannotRecruit_, newMember.Name));
```
gmriggs, line 81: should be using SendWeenieErrorWithString

Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -78,22 +78,30 @@ public void AddFellowshipMember(Player inviter, Player newMember)
...
             {
+                // todo: can't seem to find pcap of this scenario... seems odd..
                 inviter.Session.Network.EnqueueSend(new GameMessageSystemChat($"{newMember.Name} is already in a fellowship", ChatMessageType.Fellowship));
+                //inviter.Session.Network.EnqueueSend(new GameEventWeenieError(inviter.Session, WeenieError.FellowshipMember)); 
             }
             else
             {
+                if (newMember.IsBusy)
+                {
+                    inviter.Session.Network.EnqueueSend(new GameMessageSystemChat($"{newMember.Name} is busy.", ChatMessageType.Broadcast));
```
gmriggs, line 101: This isn't really needed, and will be more of an annoyance than anything helpful / useful.

LtRipley36706: It was a thing however

gmriggs: Doesn't matter, ACE does not do Busy state management the same as retail

Busy state management is for preventing 2 actions from running at the same time which would cause state inconsistencies

This does not, and is a detrimental change

LtRipley36706: Pcap is a pcap.

gmriggs: The pcap shows nothing about why the retail server returned a busy error. Retail servers were implemented in a completely different manner, and had many more technical restrictions.

The pcap showing a busy error in some cases says nothing about what actions it was actually tied to. There is no technical or gameplay reason for such a restriction.

PR #2986, opened 2020-05-10
Adding ability to exclude specific IPs from max session count.

ahricvirindi:
This will allow an IP exception list for the max number of sessions per IP.  The use case for this would be for accounts that may have multiple bot accounts (buffbot, craftbot, portalbot), but still may want to actively play their non-bot characters.  Or in the instance where an admin/dev needs to do testing of content across multiple accounts.

### DIFF
--- a/Source/ACE.Common/NetworkSettings.cs
+++ b/Source/ACE.Common/NetworkSettings.cs
@@ -42,6 +42,6 @@ namespace ACE.Common
         /// </remarks>
         [System.ComponentModel.DefaultValue(new string[] { })]
         [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
-        public string[] AllowUlimitedSessionsFromIPAddresses { get; set; }
+        public string[] AllowUnlimitedSessionsFromIPAddresses { get; set; }
     }
 }
--- a/Source/ACE.Common/NetworkSettings.cs
+++ b/Source/ACE.Common/NetworkSettings.cs
@@ -1,4 +1,5 @@
 using Newtonsoft.Json;
+using System.Collections.Generic;
 
 namespace ACE.Common
 {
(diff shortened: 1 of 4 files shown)


Source/ACE.Common/NetworkSettings.cs
```diff
@@ -33,5 +34,14 @@ public class NetworkSettings
...
         public int MaximumAllowedSessionsPerIPAddress { get; set; }
+
+        /// <summary>
+        /// Will allow the given IP addresses to have unlimited sessions - recommend only use this for Admins
+        /// </summary>
+        /// <remarks>
+        /// </remarks>
+        [System.ComponentModel.DefaultValue(new string[] { })]
+        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
+        public string[] AllowUlimitedSessionsFromIPAddresses { get; set; }
```
LtRipley36706, line 45: rename to `AllowUnlimitedSessionsFromIPAddresses` ?
```suggestion
        [System.ComponentModel.DefaultValue(new string[] { })]
        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Populate)]
        public string[] AllowUnlimitedSessionsFromIPAddresses { get; set; }
```

Source/ACE.Server/Config.js.docker
```diff
@@ -17,7 +17,11 @@
 
       // MaximumAllowedSessionsPerIPAddress, default: -1
       // -1 means unlimited connections per IP Address.
-      "MaximumAllowedSessionsPerIPAddress": -1
+      "MaximumAllowedSessionsPerIPAddress": -1,
+
+     // AllowUlimitedSessionsFromIPs, default []
+     // Will allow the given IP addresses to have unlimited sessions - recommend only use this for Admins/Devs
+     "AllowUlimitedSessionsFromIPAddresses": []
```
LtRipley36706, line 24: perhaps give an example or two of how this would look with values in it?

PR #2988, opened 2020-05-14
Add cloaks to SelectWieldedClothing()

LtRipley36706:
Thanks @DridsGhost and @OptimShi


Mag-nus:
I think this would benefit from adding a GetEquippedCloak() function in Creature_Equipment

PR #2990, opened 2020-05-16
only send fellowship vital updates to clients w/ fellowship window currently open

gmriggs:


gmriggs:
Well that information was sent in previous versions of ACE, even when the fellowship window wasn't open. A plugin like that would have been possible in previous versions of ACE, or with a future option to toggle this feature

This update was just the result of something schneebly mentioned when he was studying the ACE network traffic, so that it meatches up with retail a bit closer

PR #2994, opened 2020-05-23
Myquests command fixes

trevis:
- use quest_mindelta_rate server prop when displaying myquests output
- add 'Quest list is empty' response to myquests

### DIFF
--- a/Source/ACE.Server/Command/Handlers/PlayerCommands.cs
+++ b/Source/ACE.Server/Command/Handlers/PlayerCommands.cs
@@ -38,6 +38,11 @@ namespace ACE.Server.Command.Handlers
                 return;
             }
 
+            if (session.Player.QuestManager.Quests.Count == 0) {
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Command/Handlers/PlayerCommands.cs
```diff
@@ -48,8 +54,9 @@ public static void HandleQuests(Session session, params string[] parameters)
                     Console.WriteLine($"Couldn't find quest {playerQuest.QuestName}");
                     continue;
                 }
+                var minDelta = (uint)(quest.MinDelta * PropertyManager.GetDouble("quest_mindelta_rate", 1).Item);
```
LtRipley36706, line 57:
```suggestion
                var minDelta = (uint)(quest.MinDelta * PropertyManager.GetDouble("quest_mindelta_rate").Item);
```i think we generally avoid fallback default as that had led to some instances of property desync inbetween saves...

PR #2998, opened 2020-05-25
fix minor off-by-1 numbers

gmriggs:


harliq:
I would change 0 to a 1 and leave the 100

gmriggs:
> I would change 0 to a 1 and leave the 100, since if lowering to 99, the higher end roll chance is decreasing.

Changing it to 1,100 would require all of the ifs to be updated to <= instead of <, and would produce the exact same result as the currently updated code.

This fixes the RNG roll to match with what the intended rates of the if statements. The original was bugged for every if branch, in that something that was intended to have a 20/100 chance was having a 20/101 chance.

gmriggs:
original:

```
	case 2:
		chance = ThreadSafeRandom.Next(0, 100);
		if (chance < 80)
			damageMod = .01;
		else if (chance < 95)
			damageMod = .02;
		else
			damageMod = .03;
		break;
```
		
80/101 = 0.792%		
15/101 = 0.148%
6/101 = 0.059%

currently updated version:

80%
15%
5%

your suggestion:

79%
15%
6%

your suggestion, and also changing every < to <= :

80%
15%
5%

harliq:
You are right, I was looking at it from a >= stand point and not a > stand point.

PR #3014, opened 2020-06-09
Loot Sim update - Cloaks and cantrip counts

harliq:
This updated adds support for cloaks and cantrips


Source/ACE.Server/Factories/LootGenerationFactory_Test.cs
```diff
@@ -193,7 +215,21 @@ public static LootStats LootStats(WorldObject wo, LootStats ls, bool logstats)
                         }
                         break;
                     case ItemType.Clothing:
-                        ls.ClothingCount++;
+                        if (testItem.Name.Contains("Cloak"))
```
gmriggs, line 218: A minor concern would be picking up any other items with "Cloak" in the name, that aren't actually cloak

There is an IsCloak() function which could determine this more reliably

PR #3015, opened 2020-06-09
Change default for equipmentsetid_enabled

LtRipley36706:


gmriggs:
Just a word of caution, this will still be set to 'false' for most servers

This will only be set to true for brand new servers that are installed after this patch

PR #3027, opened 2020-06-16
fixing /removeinst for generator children after reload landblock

gmriggs:


harliq:
I have tested this, and it works.   Of note, you have to reload the landblock for the changes to take effect.  If you remove the child instance,  it will continue to spawn till you reload.  This is a given, but just to avoid confusion for others.

DridsGhost:
I have tested while making two dungeons today. It works just fine.

PR #3031, opened 2020-06-19
Added Variance for Cloak Proc Rate depending on the level of cloak.

harliq:
Reduced Overall chance from 2.5 to 2.0, then added cloak level to proc chance equation.
Cloak Levels proc rates end up as the following
```
Level 1 - 2.1
Level 2 - 2.2
Level 3 - 2.3
Level 4 - 2.4
Level 5 - 2.5

### DIFF
--- a/Source/ACE.Server/Entity/Cloak.cs
+++ b/Source/ACE.Server/Entity/Cloak.cs
@@ -9,7 +9,7 @@ namespace ACE.Server.Entity
 {
     public class Cloak
     {
-        private static readonly float ChanceMod = 2.5f;
+        private static readonly float ChanceMod = 2.0f;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Source/ACE.Server/Entity/Cloak.cs [hunk 1]
```diff
@@ -19,7 +19,7 @@ public static bool TryProcSpell(Creature defender, WorldObject attacker, WorldOb
         {
             if (cloak == null) return false;
 
-            if (!RollProc(damage_percent))
+            if (!RollProc(damage_percent, cloak.ItemMaxLevel.Value))
```
gmriggs, line 22: this will throw a null pointer exception if a cloak doesn't have a level

i will adjust the code in this pr around a bit later today

Issue #3034, opened 2020-06-20
Sessions with an intended AccessLevel of 2 are granted AccessLevel 3


FlaggAC:
### Bug Report or Feature Request (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
Grant an account AccessLevel 2 via a sql query `update ace_auth.account set accesslevel = 2 where ...`
Log in as that user
Delete a lifestone or a player's wielded weapon with the /delete command

### The log given by the failure.
No log, but here are the exact lines of code detailing the issue

https://github.com/ACEmulator/ACE/blob/7661da0a93e69dcbde88d714d0ee17201a5f369d/Source/ACE.Server/Command/CommandManager.cs#L230

https://github.com/ACEmulator/ACE/blob/7661da0a93e69dcbde88d714d0ee17201a5f369d/Source/ACE.Server/Command/Handlers/AdminCommands.cs#L81-L82

### Desired functionality.
Any command marked as  AccessLevel.Envoy should require an AccessLevel of 3, not 2.

FlaggAC:
For reference: another line of code that may be useful for fixing https://github.com/ACEmulator/ACE/blob/master/Source/ACE.Server/WorldObjects/Player.cs#L151

PR #3035, opened 2020-06-21
Fix for sentinel/envoy access level [#3034]

FlaggAC:
Fixes https://github.com/ACEmulator/ACE/issues/3034

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player.cs
+++ b/Source/ACE.Server/WorldObjects/Player.cs
@@ -147,8 +147,10 @@ namespace ACE.Server.WorldObjects
                     IsAdmin = true;
                 if (Session.AccessLevel == AccessLevel.Developer)
                     IsArch = true;
-                if (Session.AccessLevel == AccessLevel.Envoy || Session.AccessLevel == AccessLevel.Sentinel)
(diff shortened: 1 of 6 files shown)


Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -147,8 +147,10 @@ private void SetEphemeralValues()
                     IsAdmin = true;
                 if (Session.AccessLevel == AccessLevel.Developer)
                     IsArch = true;
-                if (Session.AccessLevel == AccessLevel.Envoy || Session.AccessLevel == AccessLevel.Sentinel)
+                if (Session.AccessLevel == AccessLevel.Sentinel)
                     IsSentinel = true;
+                if (Session.AccessLevel == AccessLevel.Envoy)
+                    IsEnvoy = true;
```
LtRipley36706, line 153:
```suggestion
                if (Session.AccessLevel == AccessLevel.Envoy)
                {
                    IsEnvoy = true;
                    IsSentinel = true;
                }
```
IsSentinel is important to the client, so an Envoy still needs this to be true. IsEnvoy isn't a *real* thing so sending it to the client would have no effect.

PR #3041, opened 2020-06-23
Spell counts fix

harliq:
Spell Counts were easy to implement, however there is a rabbit hole this opened due the current ItemDifficulty (Arcane Req) was being calculated via number of spells.  PR completely changes how arcane requirement is calculated.


Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
+                case ItemType.Gem:
                 default:
+                    minItemSpellCraftRange = 1.00f;
+                    maxItemSpellCraftRange = 1.00f;
                     break;
             }
 
-            return spellcraft;
+            // Getting the spell difficulty - Maybe a better way to do this.
+            var maxSpellLevel = wo.GetMaxSpellLevel();
```
gmriggs, line 1516: one thing to be aware of with the current implementation of GetMaxSpellLevel(), is that the result of this function is cached for the object after the first call, for max performance. So it must be ensured that no further spells are added to this object after the first call to GetMaxSpellLevel()

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
+            int maxSpellDiff = 1;
+
+            if (maxSpellLevel == 2)
```
gmriggs, line 1519: consider switching to a 'switch' statement, instead of a bunch of if/else... this particular scenario is perfectly suited to a switch statement, and would perform slightly better, as it would only require 1 evaluation instead of up to 8

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
+            else if (maxSpellLevel == 6)
+                maxSpellDiff = 250;
+            else if (maxSpellLevel == 7)
+                maxSpellDiff = 300;
+            else if (maxSpellLevel == 8)
+                maxSpellDiff = 400;  
+
+            var tItemSpellCraft = maxSpellDiff * ThreadSafeRandom.Next(minItemSpellCraftRange, maxItemSpellCraftRange);
+
+            if (tItemSpellCraft < 0)
```
gmriggs, line 1536: looking at the code, this particular scenario not be possible?

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
-                    difficulty = spellcraft + (ThreadSafeRandom.Next(0, 10) * ThreadSafeRandom.Next(1, 3));
-                    break;
-                case 8:
-                    difficulty = spellcraft + (ThreadSafeRandom.Next(0, 10) * ThreadSafeRandom.Next(1, 3));
-                    break;
-                default:
-                    break;
-            }
+            int wieldReq = 1;
+            int sc = 0;
```
gmriggs, line 1548: some of these vars are confusing named, and don't provide much insight into what they are, without studying the code further

sc, spc, bq

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1053,95 +1054,37 @@ private static int GetSpellDistribution(TreasureDeath profile, out int numMinors
...
+            //  From Mags Data all tiers have about the same chance for a given number of spells on items.  This is the ratio for magical items.
+            //  1 Spell(s) - 46.410 %
+            //  2 Spell(s) - 27.040 %
+            //  3 Spell(s) - 17.850 %
+            //  4 Spell(s) - 6.875 %
+            //  5 Spell(s) - 1.525 %
+            //  6 Spell(s) - 0.235 %
+            //  7 Spell(s) - 0.065 %
+
+            if (nonCantripChance <= 46410)
```
gmriggs, line 1074: is this code no longer based on tier?

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
+            spc = wo.Biota.PropertiesSpellBook.Count();
+
+            if (wo.EpicCantrips.Count > 0)
+                epicAddon = ThreadSafeRandom.Next(1, 5) * wo.EpicCantrips.Count;
+            if (wo.LegendaryCantrips.Count > 0)
+                legAddon = ThreadSafeRandom.Next(5, 10) * wo.LegendaryCantrips.Count;
+
+            if (wo.ItemAllegianceRankLimit.HasValue)
+                sc = wo.ItemAllegianceRankLimit.Value;
+            if (wo.WieldDifficulty.HasValue)
```
gmriggs, line 1562: this if statement should have { } brackets around it

the way it is currently written is very scary, and very prone to error. i honestly don't even know how the compiler would evaluate this one, without stepping through it in the debugger

Source/ACE.Server/Factories/LootGenerationFactory.cs
```diff
@@ -1545,78 +1488,108 @@ private static int GetWorkmanship(int tier)
...
+                tItemSpellCraft = 0;
+
+            int finalItemSpellCraft = (int)Math.Floor(tItemSpellCraft);
+
+            return finalItemSpellCraft;
+
         }
 
-        private static int GetDifficulty(int tier, int spellcraft)
+        private static int GetDifficulty(WorldObject wo, int tier, int itemspellcraft)
```
gmriggs, line 1545: is tier no longer a requirement for this function?

i see that tier is still passed in as a parameter, but it seems to no longer be used

Source/ACE.Server/Factories/LootGenerationFactory.cs
mystic3470: By memory, agree that end of retail 7+ spell items were rare.  But I recall ~4 or 5 spell items being very common.  What was the data used to decide upon the distribution where a 5-spell item should be a 1.525% find?

harliq: From retail data that was collected.  Thats where the data came from

mystic3470: Got it.  Did that include any banding by Tier?  For example, was it the same distribution in PCAP for Drudgehideout item with Lvl 1 spells and Dark Isle item with Lvl 8's?

PR #3042, opened 2020-06-24
Update Solution to use VS2019/ver 16

OptimShi:
Updated ACE.sln to use VS2019.

### DIFF
--- a/Source/ACE.sln
+++ b/Source/ACE.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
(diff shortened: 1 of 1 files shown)


Mag-nus:
Out of curiosity, what does this gain, or affect?

The best I could find is that this controls the icon: https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file?view=vs-2019

```
# Visual Studio Version 16
The major version of Visual Studio that (most recently) saved this solution file. This information controls the version number in the solution icon.
```

OptimShi:
If you have more than one version of Visual Studio installed, it will cause Visual Studio Version Selector to open the appropriate version. As it stands, if you have VS2017 and 2019 installed, double-clicking the .SLN causes 2017 open--which does not work with this.

PR #3044, opened 2020-06-26
Updates to Fellowship locking

LtRipley36706:


Source/ACE.Server/Network/GameEvent/Events/GameEventFellowshipFullUpdate.cs
```diff
@@ -48,9 +37,14 @@ public GameEventFellowshipFullUpdate(Session session)
 
             Writer.Write(Convert.ToUInt32(fellowship.IsLocked));
 
-            // TODO PackableHashTable of fellows departed - fellowsDeparted  -<ObjectID,int>
-            Writer.Write((uint)0x00200000);
-            Writer.Write((uint)0x00200000);
+            PackableHashTable.WriteHeader(Writer, fellowship.DepartedMembers.Count);
+            foreach (var kvp in fellowship.DepartedMembers)
```
gmriggs, line 41: probably requires to be in sorted order by % 32

Source/ACE.Server/Network/GameEvent/Events/GameEventFellowshipFullUpdate.cs
```diff
@@ -48,9 +37,14 @@ public GameEventFellowshipFullUpdate(Session session)
...
+            {
+                Writer.Write(kvp.Key);
+                Writer.Write(kvp.Value);
+            }
+
+            Writer.Write(fellowship.FellowshipLocks);
```
gmriggs, line 47: This does not match https://acemulator.github.io/protocol/Messages/0x02BE-S2C.html, nor does it match any of the code in pdb for acclient.exe

See the relevant functions: UIQueueManager::ProcessNetBlobData -> CM_Fellowship::DispatchUI_FullUpdate -> Fellowship::UnPack
                                                                           -> ClientFellowshipSystem::Handle_Fellowship__FullUpdate
, specifically Fellowship:UnPack

The fellowship data structure ends at DepartedMembers, according to the client processing for this packet

Is this a feature that was added after the client pdb?

Source/ACE.Server/Network/GameEvent/Events/GameEventFellowshipFullUpdate.cs
```diff
@@ -48,9 +37,14 @@ public GameEventFellowshipFullUpdate(Session session)
 
             Writer.Write(Convert.ToUInt32(fellowship.IsLocked));
 
-            // TODO PackableHashTable of fellows departed - fellowsDeparted  -<ObjectID,int>
-            Writer.Write((uint)0x00200000);
-            Writer.Write((uint)0x00200000);
+            PackableHashTable.WriteHeader(Writer, fellowship.DepartedMembers.Count);
```
gmriggs, line 40: This is likely incorrect header. Should be using the example above from the fellowship hash table.

# of departed fellows, and what is likely a static bucket size (16?)

Source/ACE.Server/Network/GameEvent/Events/GameEventFellowshipFullUpdate.cs
LtRipley36706: its not found in UnPack, but is sent across wire. Found in Pack

Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -694,4 +730,40 @@ public void ProcessDropList(Dictionary<uint, WeakReference<Player>> fellowshipMe
             UpdateAllMembers();
         }
     }
+
+    public class DepartedMembersComparer : IComparer<uint>
+    {
+        public static ushort TableSize = 32;
```
gmriggs, line 736: are you sure the table size differed here from the main fellowships table? it seems odd that one would use 16, and another would use 32

if they indeed differed, this should be implemented as a constructor that takes the table size. This logic shouldn't be duplicated for ever new table size

LtRipley36706: i have no idea

Source/ACE.Server/Network/GameEvent/Events/GameEventFellowshipFullUpdate.cs
gmriggs: "Is this a feature that was added after the client pdb?"

Any answer to this question?

The pdb client doesn't use this data after DepartedMembers at all. The only explanation would be if this locking data structure was added at the end of the game, after the pdb client.

These PRs also need an explanation of what the changes actually are, the reasoning for the changes, and repro steps to test them. What effect does this have on the client, and how can it be tested?

LtRipley36706: > Not sure where in the client, but aclogview decodes it
it was sent across the wire. it probably has no purpose to the client

gmriggs: looking in the final client, sub_5BA630 is Fellowship::UnPack

it looks to have no additional logic, and matches the pdb verison... DepartedFellows is the last thing it uses

if this data is completely unused by anything, and has 0 effect on the client or anything in the game, what is the purpose of this code if it does nothing?

LtRipley36706: to fill out aclogview parse.. the data *was* there as `Writer.Write((uint)0x00200000);`

PR #3055, opened 2020-06-30
Fix issue with spent xp verification

LtRipley36706:


Source/ACE.Server/Command/Handlers/DeveloperFixCommands.cs
```diff
@@ -803,68 +803,77 @@ public static void HandleVerifyExperience(Session session, params string[] param
...
+                        aug.PropertiesInt64.TryGetValue(PropertyInt64.AugmentationCost, out var costPer);
 
-                    augXP += costPer * numAugs;
-                }
+                        augXP += costPer * numAugs;
+                    }
 
-                var calculatedSpent = attributeXP + vitalXP + skillXP + augXP + diffXP;
+                    // Asheron's Lesser Benediction augmentation operates differently than all other augs.
+                    var hasLesserBenedictionAug = ctx.CharacterPropertiesQuestRegistry.Count(i => i.CharacterId == player.Guid.Full && i.QuestName.Equals("LesserBenedictionAug")) == 1;
```
gmriggs, line 855: should use Any() at the minimum

looping through every record in the entire unindexed quest registry on the server, every time for each player, seems very unoptimized and slow. does this add any significant speed reduction to running this operation?

this should be building a lookup table beforehand

gmriggs:
```
            HashSet<uint> lesserBenediction = null;

            using (var ctx = new ShardDbContext())
            {
                // Asheron's Lesser Benediction augmentation operates differently than all other augs
                lesserBenediction = ctx.CharacterPropertiesQuestRegistry.Where(i => i.QuestName.Equals("LesserBenedictionAug")).Select(i => i.CharacterId).ToHashSet();
            }

                    if (lesserBenediction.Contains(player.Guid.Full))
                        augXP += 2000000000;
```

LtRipley36706: sounds good to me

PR #3058, opened 2020-07-02
Fix issue with destroying stack in containers

LtRipley36706:


OptimShi:
This fixes the use case I noticed.

PR #3061, opened 2020-07-04
Add support for QuestBits emotes

LtRipley36706:


Source/ACE.Server/Managers/QuestManager.cs
```diff
@@ -613,5 +613,41 @@ public void OnDeath(WorldObject killer)
             if (Creature.KillQuest3 != null)
                 player.QuestManager.HandleKillTask(Creature.KillQuest3, Creature);
         }
+
+        public bool HasQuestBitsOn(string questFormat, int bits)
```
gmriggs, line 617: maybe a comment for this function that it returns true if any of the quest bits are on, and not necessarily all of the quest bits

for example, if i pass quest bits A & B to this function, and the player only has quest bit A, it will still return true

LtRipley36706: is there a better code for has all the bits passed in?

gmriggs: ``(quest.NumTimesCompleted & bits) == bits`` would be having all of the flags

seems odd that this is checking against 'NumTimesCompleted', instead of a flags field

LtRipley36706: i doubt they made any change to the fields inferred in QuestManager. i'm imagining very simple 0x1, 0x2, 0x4, 0x8, 0x10 type bits mostly

PR #3078, opened 2020-07-13
Update Program_DbUpdates.cs

Zurgash:
Adding a World Customization log to generate an error file when SQL errors are caught.

First crack at this, been awhile since I have used C# if there is a cleaner way to do this. Let me know!

### DIFF
--- a/Source/ACE.Server/Factories/LootTables.cs
+++ b/Source/ACE.Server/Factories/LootTables.cs
@@ -2193,6 +2193,7 @@ namespace ACE.Server.Factories
         {
             43048,  // Knorr Academy Breastplate
             43049,  // Knorr Academy Gauntlets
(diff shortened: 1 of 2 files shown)


Mag-nus:
Our existing Log4Net configuration can be used for this as well.

Lines like the following, which write directly to the console, bypass any lognet collection and settings:
Console.WriteLine($"Searching for World Customization SQL scripts .... ");

What you could do is replace the existing Console.WriteLine and follow the pattern we've established which is similar to:
Log("[WORLD_CUSTOMIZATION] Some Message....");

Then, you can filter based on that tag in the log4net and pipe those messages to a separate log (if that is the intent).

Zurgash:
I am just getting back to this. I will definitely look into trying this method. Does ACE have documentation on how Log4Net works?

PR #3085, opened 2020-07-18
Fix issue with SendMotionAsCommands and stance

LtRipley36706:


Source/ACE.Server/WorldObjects/Player_Location.cs
```diff
@@ -578,12 +578,14 @@ public void SendMotionAsCommands(MotionCommand motionCommand)
             if (FastTick)
             {
                 var actionChain = new ActionChain();
-                EnqueueMotionAction(actionChain, new List<MotionCommand>() { motionCommand }, 1.0f, MotionStance.NonCombat);
+                //EnqueueMotionAction(actionChain, new List<MotionCommand>() { motionCommand }, 1.0f, MotionStance.NonCombat);
```
gmriggs, line 581: This will break the previous usage of this function for the instances that call it from combat mode

LtRipley36706: except none can do that currently, and healer called the correct animation sequence when in combat and out of combat now

gmriggs:
```
			var animLength = DatManager.PortalDat.ReadFromDat<MotionTable>(mTableID).GetAnimationLength(MotionStance.NonCombat, MotionCommand.HouseRecall, MotionCommand.Ready);
            Console.WriteLine(animLength);
			15.060241

            var animLength = DatManager.PortalDat.ReadFromDat<MotionTable>(mTableID).GetAnimationLength(MotionStance.AtlatlCombat, MotionCommand.HouseRecall, MotionCommand.Ready);
            Console.WriteLine(animLength);
			0
```

The intention of the code here is to always call HouseRecall from a specific MotionStance, as calling it from anything but NonCombat is an invalid animation, and returns a 0 time.

The code above is bad pattern to use, as it leads to these types of errors from CurrentMotionState thrashing in the system, or from ActionChains. When refactoring, the functionality should remain the same for existing places which call this method.

PR #3087, opened 2020-07-20
[DEBUG] Log landblock group thread collisions that could throw exceptions

Mag-nus:
Doesn't prevent the exception, but, hopefully it can give me the stack trace that is suspect.


Mag-nus:
Added temp code that might help avoid exceptions.

PR #3094, opened 2020-07-27
Add support for Enlightenment

LtRipley36706:
Backend support for using the Font of Enlightenment and Rebirth found in the Temple of the Font located at Viridian Rise


LtRipley36706:
Content paired with this PR can be found at ACEmulator/ACE-World-16PY-Patches#550

PR #3108, opened 2020-08-05
Add fix-up script for PetDevices

LtRipley36706:


Database/Updates/Shard/2020-08-05-00-Update-PetDevices.sql
```diff
@@ -0,0 +1,179 @@
+INSERT INTO biota_properties_int (object_Id, `type`, value)
```
gmriggs, line 1: can be done with 1 insert statement... doesn't need 1 separate lookup for each individual wcid

PR #3109, opened 2020-08-08
add support for faction monsters not attacking same-factioned players or their combat pets, unless provoked

gmriggs:
Todo: ace's usage of Tolerance.Retaliate will need to be investigated for this. currently, if a faction player is standing next to a monster belonging to the same faction, and then a non-factioned player runs into the mix, the monster will start attacking the non-factioned player, and then can possibly switch to attacking the factioned player afterwards (fixed)

also an open question, what is supposed to happen if a faction player is standing next to a monster belonging to the same faction, and then the player spawns a summon? i assume the summon should have the player's Faction1Bits copied over to it, so that it doesn't attack the factioned mob on sight? (fixed)


gmriggs:
fixed the 'todo' scenario above, although it only applies for faction mobs, and prevent the scenario described before from happening with faction mobs. retaliate behavior currently remains the same.

investigating the summoning scenario next, and some other possible scenarios

gmriggs:
added support for CombatPets not attacking mobs belonging to the same faction as the pet owner

one thing to note about this currently, if the pet owner attacks the same-factioned mob, the faction mob will retaliate, and the CombatPet will continue to idle, not defending the player

anomalies still left on the todo list: if i attack a same-factioned mob standing in a group of other same-factioned mobs, should they all aggro me? currently using the DamageHistory method instead of a HashSet<guid> retaliateTargets, i wouldn't have a DamageHistory entry for the other mobs, so they would de-aggro quickly.

another similar scenario, if i attack a same-factioned mob, but i am evaded, it will begin attacking me, and then quickly de-aggro, since an evade is not inserted into their DamageHistory

considering switching to a HashSet<guid> retaliateTargets instead of DamageHistory.HasDamager(guid) to remedy these 2 situations

PR #3121, opened 2020-08-14
Protect against bad positions in data

LtRipley36706:


Source/ACE.Server/Physics/PhysicsGlobals.cs
```diff
@@ -6,7 +6,8 @@ namespace ACE.Server.Physics
 {
     public class PhysicsGlobals
     {
-        public static readonly float EPSILON = 0.00019999999f;
+        //public static readonly float EPSILON = 0.00019999999f;
+        public static readonly float EPSILON = 0.0001f;
```
gmriggs, line 10: You can't change this constant. This is shared between acclient

PR #3125, opened 2020-08-15
Change AttemptToFixRotation to only affect read in from DB

LtRipley36706:


gmriggs:
I misunderstood https://github.com/ACEmulator/ACE/pull/3121, specifically where IsRotationValid() was being called

In the current implementation, IsRotationValid() is being called anytime .Location for any object on the server is being referenced. This is extremely bad design, and 1000% the wrong place / frequency to be calling IsRotationValid(). On a large server, this will result in 10s of thousands of calls to IsRotationValid() per second.

IsRotationValid() should be called when a position / rotation is first read in from the database

Mag-nus:
Ripley, a good place to put the position check might be in WorldObject.cs, line 133, at the end of the WorldObject(Biota) ctor.

Source/ACE.Server/Entity/PositionExtensions.cs
```diff
@@ -325,6 +325,9 @@ public static bool IsRotationValid(this Quaternion q)
 
         public static bool AttemptToFixRotation(this Position pos, WorldObject wo, PositionType positionType)
         {
+            if (wo is SpellProjectile)
```
gmriggs, line 328: this bit shouldn't be needed now?

Source/ACE.Server/Entity/PositionExtensions.cs
```diff
@@ -325,6 +325,9 @@ public static bool IsRotationValid(this Quaternion q)
...
+                return false;
+
```
LtRipley36706, line 330:
```suggestion
```

PR #3126, opened 2020-08-16
Lugian hammer

ziang4891:
- Set subtype ThreadSafeRandom.Next() upper bounds to array.Length - 1
- Reorder the MutateMeleeWeapon() subtype case statements because of the addition and removal mentioned below
- Add the Lugian Hammer weenies to the HeavyWeaponsMatrix array
- Remove unneeded comment, in the LightWeaponsMatrix array
- Remove duplicate Board with Nail weenies mislabeled as Stone Mace, in the FinesseWeaponsMatrix array

### DIFF
--- a/Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
@@ -33,22 +33,22 @@ namespace ACE.Server.Factories
             {
                 case 0:
                     // Heavy Weapons
-                    subtype = ThreadSafeRandom.Next(0, LootTables.HeavyWeaponsMatrix.Length);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.HeavyWeaponsMatrix.Length - 1);
                     weaponWeenie = LootTables.HeavyWeaponsMatrix[subtype][eleType];
                     break;
                 case 1:
                     // Light Weapons
-                    subtype = ThreadSafeRandom.Next(0, LootTables.LightWeaponsMatrix.Length);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.LightWeaponsMatrix.Length - 1);
                     weaponWeenie = LootTables.LightWeaponsMatrix[subtype][eleType];
                     break;
                 case 2:
                     // Finesse Weapons;
-                    subtype = ThreadSafeRandom.Next(0, LootTables.FinesseWeaponsMatrix.Length);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.FinesseWeaponsMatrix.Length - 1);
                     weaponWeenie = LootTables.FinesseWeaponsMatrix[subtype][eleType];
                     break;
                 default:
                     // Two handed
-                    subtype = ThreadSafeRandom.Next(0, LootTables.TwoHandedWeaponsMatrix.Length);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.TwoHandedWeaponsMatrix.Length - 1);
                     weaponWeenie = LootTables.TwoHandedWeaponsMatrix[subtype][eleType];
                     break;
             }
--- a/Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
+++ b/Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
@@ -33,22 +33,22 @@ namespace ACE.Server.Factories
             {
                 case 0:
                     // Heavy Weapons
-                    subtype = ThreadSafeRandom.Next(0, 23);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.HeavyWeaponsMatrix.Length);
                     weaponWeenie = LootTables.HeavyWeaponsMatrix[subtype][eleType];
                     break;
                 case 1:
                     // Light Weapons
-                    subtype = ThreadSafeRandom.Next(0, 19);
+                    subtype = ThreadSafeRandom.Next(0, LootTables.LightWeaponsMatrix.Length);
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -33,37 +33,47 @@ public static WorldObject CreateMeleeWeapon(TreasureDeath profile, bool isMagica
...
                     weaponWeenie = LootTables.TwoHandedWeaponsMatrix[subtype][eleType];
                     break;
             }
 
             var wo = WorldObjectFactory.CreateNewWorldObject((uint)weaponWeenie);
 
             if (wo != null && mutate)
-                MutateMeleeWeapon(wo, profile, isMagical, weaponType, subtype);
+                if (!MutateMeleeWeapon(wo, profile, isMagical))
+                    wo = null;
```
gmriggs, line 60: what is the reasoning for setting the entire object to null if the mutation fails? This seems like a non-standard concept not used anywhere else in the loot generator

Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -33,37 +33,47 @@ public static WorldObject CreateMeleeWeapon(TreasureDeath profile, bool isMagica
...
 
             return wo;
         }
 
-        private static void MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, bool isMagical, int weaponType, int subtype)
+        private static bool MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, bool isMagical)
         {
-            Skill wieldSkillType = Skill.None;
+            if ((wo.WeaponSkill != Skill.HeavyWeapons || wo.WeaponSkill != Skill.LightWeapons || wo.WeaponSkill != Skill.FinesseWeapons || wo.WeaponSkill != Skill.TwoHandedCombat)
```
gmriggs, line 67: This is written incorrectly, and will cause every mutation of every melee weapon to fail. The intended logic here is &&, not ||

This should also be written as a switch/case statement, as it would execute faster than tons of separate if checks

Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -79,86 +89,62 @@ private static void MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, boo
...
-            switch (weaponType)
+            switch (wieldSkillType)
             {
-                case 0:
-                    // Heavy Weapons
-                    wieldSkillType = Skill.HeavyWeapons;
-
-                    switch (subtype)
+                case Skill.HeavyWeapons:
+                    switch (wo.W_WeaponType)
```
gmriggs, line 95: also hesistant about this change, as it is highly dependent on the PropertyInt.WeaponType fields being set consistently across every item in the database, which would need to be verified. it needs to be ensured that this logic matches up completely with the previous logic

Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -318,40 +269,34 @@ private static void MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, boo
...
-                            break;
-                        case 7:
-                            weaponDefense = GetMaxDamageMod(profile.Tier, 18);
-                            weaponOffense = GetMaxDamageMod(profile.Tier, 22);
-                            break;
-                        case 8:
-                        case 9:
-                        case 10:
-                        default:
+                        case AttackType.Thrust: // Assagai | Pike | Corsesca | Magari Yari
```
gmriggs, line 278: having the switch based on AttackType here also seems very iffy, and non-standard

Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -318,40 +269,34 @@ private static void MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, boo
...
                             weaponDefense = GetMaxDamageMod(profile.Tier, 15);
                             weaponOffense = GetMaxDamageMod(profile.Tier, 25);
                             damage = GetMeleeMaxDamage(wieldSkillType, wieldDiff, LootWeaponType.Spears);
                             break;
+                        case AttackType.Slash:
+                            // Great Star Mace | Quadrelle | Khanda-handled Mace | Tetsubo
+                            weaponDefense = GetMaxDamageMod(profile.Tier, 22);
+                            weaponOffense = GetMaxDamageMod(profile.Tier, 18);
+
+                            if ((wo.Name.Any() != false) && (wo.Name.Contains("Nodachi") || wo.Name.Contains("Shashqa") || wo.Name.Contains("Spadone")))
```
gmriggs, line 288: string search is slow, and logic based on item names is very iffy, and will very likely run into edge cases and other bugs here. there has to be better data to drive this logic than item names

Source/ACE.Server/Factories/LootGenerationFactory_Melee.cs
```diff
@@ -318,40 +269,34 @@ private static void MutateMeleeWeapon(WorldObject wo, TreasureDeath profile, boo
...
+                            {
+                                weaponDefense = GetMaxDamageMod(profile.Tier, 20);
+                                weaponOffense = GetMaxDamageMod(profile.Tier, 20);
+                            }
+
+                            if (wo.SetupTableId >= 33560802 && wo.SetupTableId <= 33560807) // Greataxe
```
gmriggs, line 294: logic driven by SetupIDs is also very arbitrary, and non-standard

PR #3130, opened 2020-08-17
add treasure item type chance tables

gmriggs:
This fixes an issue with Mixed Equipment Chests (DeathTreasureType 2001) such as 48743 - Legendary Chest, where they shouldn't be spawning regular gems.

This can be seen in the Mag-Loot logs from retail, ie. Container_Legendary_Chest_T8.txt. From 5,826 items in there, 255 were gems, and 100% of those gems were Aetheria. No regular gems appeared to have spawned in these chests in retail.

Aetheria is currently an 'additive' separate from gems, which can still be selected via these profiles. The MundaneItemChance for these chests is 100, but the Min/Max mundane items are 0... this has the effect in ace currently of allowing Aetheria to be spawned as an additive to mundane, even when no other mundane items can spawn.

This PR builds on the existing LootBias enum, with a structure that more closely matches retail. Even though the LootBias enum had entries defined for MixedEquipment and MagicEquipment chests, it was currently unused by any of the lootgen logic.

The TreasureItemTypeChances class has been added, which defines the chances for each TreasureItemType, with different loot profiles possible there. Previously, ACE was using 1 'global' profile for the most part, with the option for LootBias to be passed in as Armor/Weapon/Jewelry, which restricted to only those TreasureItemTypes spawning.

This concept has been extended with data structure that more closely matches retail in TreasureItemTypeChances. The existing logic and profiles should still work exactly the same, and the Default/global profiles still exist and populated with the same data. The structure allows for new profiles to be easily defined, such as what was LootBias was moving towards before.

A new profile has been added for MixedEquipment and MagicEquipment (they were considered equivalent in the LootBias enum). The item type chance table for this profile has been set as a mixture between MagLoot logs for these chests, and the current global default in ace. These #s were adjusted and playtested for something that 'felt' good to me, but any of these #s can be further adjusted

Repro steps:

/teleloc 0xC6A90024 [104.717110 89.921745 42.005001] -0.839618 0.000000 0.000000 -0.543177
appraise a legendary chest
/crack
open chest
repeat

Expected:

no gems ever spawn, they should not exist in the profile

Actual:

regular gems were spawning previously

### DIFF
--- a/Source/ACE.Server/Factories/Entity/TreasureItemTypeChance.cs
+++ b/Source/ACE.Server/Factories/Entity/TreasureItemTypeChance.cs
@@ -1,0 +1,16 @@
+using ACE.Server.Factories.Enum;
+
+namespace ACE.Server.Factories.Entity
+{
+    public class TreasureItemTypeChance
+    {
+        public TreasureItemType TreasureItemType;
+        public float Chance;
+
+        public TreasureItemTypeChance(TreasureItemType treasureItemType, float chance)
+        {
+            TreasureItemType = treasureItemType;
+            Chance = chance;
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


gmriggs:
Requesting reviews from @harliq and @ziang4891

PR #3131, opened 2020-08-17
add spell selection tables

gmriggs:
This PR updates the int[][] spell tables in LootTables.cs and moves to data structures and methods which more closely match retail

The existing data from LootTables remains exactly the same. It is just being ported to a more semantic structure, that is easier to read and understand, easier to maintain, and less prone to manual data entry errors from raw int arrays.

- SpellLevelProgression has been added, which eliminates the need to manually specify each individual Level 1-8 for the spells. For any SpellId, SpellLevelProgression.GetSpellLevels(spellId) can be called, which returns a List&lt;SpellId&gt; for all of the associated spell levels. Note that this differs from SpellCategory: ie., Strength Self and Srength Other are both in the StrengthRaising category, but are part of different SpellLevelProgression groups.

- The entries for HeavyWeapons, LightWeapons, FinesseWeapons, and MissileWeapons have been updated in the SpellId enum. Comments for the original Sword/Axe/Dagger/Bow field names have been left in the comments next to each one. This is to prevent many possible errors and confusion about 'which is the correct spell for MissileWeapons', for example, since multiple enums such as the legacy CrossbowMastery and ThrownWeaponsMastery may also map to MissileWeapons, where only the legacy BowMastery is the one that should be used.

- shuffledValues array building has been optimized with a faster method. Shuffle() function has been written to use the standard ThreadSafeRandom, instead of new Random(). Calling new Random() is not standard in the project, and can be especially dangerous if the code is used in a .net standard / .net framework project, as quickly calling new Random() in quick succession will result in the same values being returned in .net framework projects

- some open questions / possible errors were discovered in the existing LootTables spell tables. I have asked ziang and others about these, waiting for confirmation on what the values should be

- for building the interface to the original [][] multidimensional array API, since this was completely statically defined before, and parts have been moved to building dynamically through SpellProgression, measuring the performance of this section and any possible overhead it might introduce was a concern. Luckily, building each individual spell table on average took about 0.3ms. The entire process for all the tables should only consume about 4ms total, one time. The first time an individual spell table is referenced, there would be a 0.3ms delay

- ChanceTables.cs has been renamed to QualityChance.cs

- The original spell selection groups from retail pre-T7 have been included for reference. The key is PropertyInt.TsysMutationData &gt;&gt; 24. The groups are:

```
1 - gems
2 - jewelry
3 - crowns
4 - orbs, casters
5 - wands, staffs, sceptres, batons
6 - melee and missile weapons
7 - bracers, breastplates, coats, cuirasses, girths, hauberks, pauldrons, chest armor, sleeves
8 - shields
9 - gauntlets, sollerets, sandals, boots
10 - helms, basinets, helmets, coifs, cowls, heaumes, kabutons
11 - boots, sandals, shoes, slippers, sollerets
12 - breeches, jerkins, shirts, pants, tunics, doublets, trousers, pantaloons
13 - caps, qafiyas, turbans, fezs, berets
14 - cloth gloves (1 entry?)
15 - greaves, leggings, tassets, leather pants
16 - dinnerware
```

However, this is just for reference currently, and the current ace lootgen spell selection 'groups' remain completely the same. The current ace spell selection groups:

- Armor
- Gems
- Jewelry
- MeleeWeapons
- MissileWeapons
- Scrolls
- Wands

ace further divides most of these groups into 'spells' and 'cantrips'


gmriggs:
Requesting review from @harliq and @ziang4891

PR #3133, opened 2020-08-17
adding loot table comments and melee fixes from 4eyebiped

gmriggs:


gmriggs:
Requesting review from @harliq and @ziang4891

Issue #3113, opened 2020-08-12
Trading issue


Starcrusher55:
### Bug Report or Feature Request (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
Initiate a trade that will give items from an overencumbered character to one that has no burden.  Error message indicates the overencumbered character cannot complete the trade even though the trade would not give him any burden bac, only take away

LtRipley36706:
fixed in master

PR #3151, opened 2020-08-22
Fix for trading from characters at 300% burden

mcreedjr:
I believe this addresses the issue #3113. Sorry if this was discussed in a dev channel as I don't frequent them as much as I'd like, and is somehow not the right fix. Just let me know and I'll close it out.

### DIFF
--- a/Source/ACE.Server/WorldObjects/Container.cs
+++ b/Source/ACE.Server/WorldObjects/Container.cs
@@ -432,7 +432,7 @@ namespace ACE.Server.WorldObjects
             TooEncumbered = false;
             NotEnoughFreeSlots = false;
 
-            if (this is Player player && !player.HasEnoughBurdenToAddToInventory(worldObjects))
+            if (this is Player player && worldObjects.Count >=1 && !player.HasEnoughBurdenToAddToInventory(worldObjects))
             {
                 TooEncumbered = true;
                 return false;
--- a/Source/ACE.Server/WorldObjects/Container.cs
+++ b/Source/ACE.Server/WorldObjects/Container.cs
@@ -432,7 +432,7 @@ namespace ACE.Server.WorldObjects
             TooEncumbered = false;
             NotEnoughFreeSlots = false;
 
-            if (this is Player player && worldObjects.Count >=1 && !player.HasEnoughBurdenToAddToInventory(worldObjects))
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Container.cs
```diff
@@ -432,7 +432,7 @@ public bool CanAddToInventory(List<WorldObject> worldObjects, out bool TooEncumb
             TooEncumbered = false;
             NotEnoughFreeSlots = false;
 
-            if (this is Player player && !player.HasEnoughBurdenToAddToInventory(worldObjects))
+            if (this is Player player && worldObjects.Count >=1 && !player.HasEnoughBurdenToAddToInventory(worldObjects))
```
LtRipley36706, line 435:
```suggestion
            if (this is Player player && worldObjects.Count >= 1 && !player.HasEnoughBurdenToAddToInventory(worldObjects))
```

mcreedjr:
I was able to reproduce it using these steps:

1. Installed latest code base, world database and confirmed all DB updates applied 
2. Created two new characters
3. Created some items in the inventory of one character using /ci developer command until I could no longer create new items
4. Buffed the character with Strength spell
5. Created more items in the characters inventory
6. Killed the char using /die to remove the Strength buff
7. This put the character at 300% burden
8. Initiated trade between two characters and placed items on the side of the character currently with 300% burden and left nothing in the trade side of the unburdened character
9. Attempted to complete trade and it failed because the burdened character was still at 300% burden, but was not taking on any additional weight

LtRipley36706:
> I was able to reproduce it using these steps:
> 
> 1. Installed latest code base, world database and confirmed all DB updates applied
> 2. Created two new characters
> 3. Created some items in the inventory of one character using /ci developer command until I could no longer create new items
> 4. Buffed the character with Strength spell
> 5. Created more items in the characters inventory
> 6. Killed the char using /die to remove the Strength buff
> 7. This put the character at 300% burden
> 8. Initiated trade between two characters and placed items on the side of the character currently with 300% burden and left nothing in the trade side of the unburdened character
> 9. Attempted to complete trade and it failed because the burdened character was still at 300% burden, but was not taking on any additional weight

cool thanks for the steps.. i had created and maxxed out a char but i didn't do the buff and add more.. will try that next

PR #3161, opened 2020-08-23
add void damage modifier for player targets from retail

gmriggs:
http://acpedia.org/wiki/Announcements_-_11th_Anniversary_Preview#Void_Magic_and_You.21

"Creatures under Asheron’s protection take half damage from any nether type spell."

I probably skimmed over this when originally researching Void Magic, and it didn't click with me that "creatures under Asheron's protection" meant "players". I might have misinterpreted as Lifestone Protection or something..


gmriggs:
Superceded by https://github.com/ACEmulator/ACE/pull/3168

PR #3162, opened 2020-08-23
add slayerMod to void dot tick calculation

gmriggs:
weeping wand supposedly boosted void dot ticks in pvp

dirty fighting bleed dots were supposedly unaffected by any slayer properties on the weapon


gmriggs:
closing - after talking with original reporters more, it was determined slayerMod didn't affect dots in retail

PR #3169, opened 2020-08-26
Add support to EmoteManager for ResistSpell emote category

LtRipley36706:


Source/ACE.Server/WorldObjects/WorldObject_Magic.cs
```diff
@@ -339,6 +339,8 @@ public bool TryResistSpell(WorldObject target, Spell spell, WorldObject caster =
                         player.Session.Network.EnqueueSend(new GameMessageSystemChat($"{targetCreature.Name} resists your spell", ChatMessageType.Magic));
 
                     player.Session.Network.EnqueueSend(new GameMessageSound(player.Guid, Sound.ResistSpell, 1.0f));
+
+                    targetCreature.EmoteManager.OnResistSpell(player);
```
gmriggs, line 343: should be moved down to the end of if (resisted)

if a CombatPet casts a spell on a monster, and the spell is resisted, the monster should still be calling EmoteManager.OnResistSpell(combatPet)

gmriggs: er since combat pets don't cast spells, probably not the best example... but if we have monsters attacking monsters from factions eventually, the concept is still sound... this shouldn't be dependent on the attacker being a player

PR #3171, opened 2020-08-26
fix exception when giving mana to an item w/ ItemMaxMana, but not ItemCurMana

gmriggs:
- /ci 30187
- /ci manastonegreat
- double click mana stone
- click 30187 - Hunter's Crystal

Hunter's Crystal has ItemMaxMana=10000, but ItemCurMana=null

the server gets an exception, although it doesn't crash the server. the player is then locked in perma hourglass state.

the correct behavior from retail needs to be confirmed here. could mana stones give mana to items where ItemMaxMana has value, but ItemCurMana=null?

at the minimum, it shouldn't get an exception / lock player into perma hourglass state..

one thing to note about this, if this was possible in retail, then the current ace logic would then consider the rare to be in the 'potential pool' of items to give mana to afterwards, if the player double clicks a mana stone, and then clicks their character

### DIFF
--- a/Source/ACE.Server/WorldObjects/ManaStone.cs
+++ b/Source/ACE.Server/WorldObjects/ManaStone.cs
@@ -86,7 +86,7 @@ namespace ACE.Server.WorldObjects
                         useResult = WeenieError.ActionCancelled;
                     else
                     {
-                        if (!target.DeleteObject(player))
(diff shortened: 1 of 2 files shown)


gmriggs:
also fixes a bug with using an empty mana stone on an equipped item doesn't destroy the item properly

PR #3187, opened 2020-08-30
update weapon swapping in combat to be synced closer to retail

gmriggs:
This update delays the actual PutObjInContainer server response/action to after the player switches to HandCombat (where appropriate). This also has the effect of delaying the visual weapon swap, until the player reaches HandCombat mode.

ACE was already doing the stance swapping correctly, it just wasn't delaying the PutObjInContainer response

This could potentially fix the weapon swapping bugs spigot mentioned, when the player has higher latency to the server


ziang4891 (dismissed):
Been using on my server for awhile now without any noticeable, deleterious effects

LtRipley36706 (approved):

PR #3197, opened 2020-09-02
adding item type chances data for reference

gmriggs:
These would map to the treasure_death table, for the *_Selection_Chances columns

To use this data, we would either need to add a translation layer, from retail TreasureItemType -> ace TreasureItemType, or we would need to update ace to use something more akin to the retail TreasureItemType

This data is currently just for reference, and is currently not being used by the loot generator


LtRipley36706 (dismissed):

ziang4891 (dismissed):

gmriggs:
updated with slightly improved formatting

LtRipley36706 (approved):

PR #3199, opened 2020-09-02
adding chance tables for gem count

gmriggs:
This PR adds the treasure_gem_count table, keyed by GemCode (TsysMutationData >> 8) and Tier

This converts the global flat ranges that are the same for every tier, into different profiles, with different chances for each # of gems for each tier

Tier 1-6 data is defined, with tiers 7 and 8 needing to be added. T7-8 are currently using the T6 profile.

If an item is missing a GemCode, the original ACE method is used as a fallback

The Database\Updates\World\treasure-gem-count.sql script is temporary, just to get this PR out for testing. This will be converted into a proper update to Database\Base\WorldBase.sql, with the data being moved to ACE-World-16PY-Patches

PR #3201, opened 2020-09-02
adding gem material chance tables

gmriggs:
This builds on https://github.com/ACEmulator/ACE/pull/3199

Previously, ACE was rolling for a random # between 10-50 for the gem material types, each with an equal chance

This PR adds RollGemType(), which uses a process that is matched up closer to retail. There are 2 separate chance tables used for this: first, each tier has a different set of chances of rolling a certain gem quality, which is a # between 1 and 6. Each gem quality then has another chance table, with different material types possible for each one.


gmriggs:
Requesting a review from @ziang4891 and @harliq

LtRipley36706 (approved):

ziang4891 (approved):

ziang4891 (approved):

LtRipley36706 (approved):

PR #3213, opened 2020-09-06
adding reference wcid tables

gmriggs:
This follows a similar structure as some of the previous lootgen prs, where the data is provided as a reference point, and the existing lootgen logic remains the same


gmriggs:

ziang4891 (approved):

LtRipley36706 (approved):

PR #3224, opened 2020-09-11
add debug line to AddWorldObjectInternal for Players

gmriggs:


gmriggs:
nm, previous line should already cover this scenario

PR #3242, opened 2020-09-17
faction mobs vs. mobs not belonging to same faction

gmriggs:


DridsGhost:
Preliminary testing looks pretty decent. Spells do not seem to have an affect in any scenario when mobs fight.

Currently with this PR, a Faction mob will attack any mob with or without a faction.

If the mob does not have a faction that is being attacked by projectile, or life/creature it will not aggro.

DridsGhost:
Projectile spells work, and aggro opposing mobs. Life and Creature vulns are still being ignore.

LtRipley36706 (dismissed):

LtRipley36706 (approved):

PR #3253, opened 2020-09-25
adding weapon speed mutations

gmriggs:
This extends the loot system with support for mutating the weapon speed (WeaponTime) for melee and missile weapons

This particular mutation makes use of QualityChance. An initial roll is performed, which determines if the WeaponTime will be mutated at all. The chance for this initial roll succeeding is based on tier, and optionally the TreasureDeath profile can contain a LootQualityMod, which can increase the chances of some mutations happening, such as this one.

If the initial roll succeeds, QualityChance then returns a qualityLevel between 1-12, with the range and chance curves also based on tier / qualityMod. This qualityLevel is the major factor in determining the weapon speed modification, along with some extra rng, which also ends up smoothly interpolating been the qualityLevels.

The range of possible weaponSpeedMods is: 67.5% - 100%


gmriggs:

ziang4891 (approved):

LtRipley36706 (approved):

LtRipley36706 (approved):

PR #3271, opened 2020-10-04
Update NoDrop_Landblocks to include Mhoire Armory

ziang4891:


Source/ACE.Server/WorldObjects/Corpse.cs
```diff
@@ -397,6 +397,7 @@ public void TryGenerateRare(DamageHistoryInfo killer)
         /// </summary>
         public static HashSet<ushort> NoDrop_Landblocks = new HashSet<ushort>()
         {
+            0x00EA,     // Mhoire Armory
```
gmriggs, line 400: should be after colo

all of these landblocks are specified in order

ziang4891: Fixed

gmriggs (approved):

PR #3290, opened 2020-10-11
adding eor mutation scripts for ArmorLevel

gmriggs:
see https://github.com/ACEmulator/ACE/pull/3290/commits/d8d18cfe3f4d3dfae79860e1ff9bcab341f44f52 for the updates specific to this pr

- verified extremities cap out at 344, non-extremities 314, and shields 274, as per retail

an interesting note, even though the py16 mutation scripts and the eor scripts both randomize between (-5, 5), this range was inclusive/exclusive. in the mutation system, this range is cast to floats, and the result is eventually cast back to an int. so in terms of actual inclusive range, retail actually used (-5, 4) here. that is why everything caps out at +4, instead of +5


gmriggs:

LtRipley36706 (approved):

LtRipley36706 (approved):

LtRipley36706 (approved):

harliq (approved):

PR #3291, opened 2020-10-11
quick fix for spellbook reordering

gmriggs:
For some reason, the WeeniePropertiesSpellBook collection from EF is being sorted by SpellId, instead of by Id

Repro steps:

- Create a monster that has a spellbook with spells not in ascending SpellId order
- Appraise monster in-game
- /debugspellbook

Expected:

Spellbook is in original order in DB

Actual:

Spellbook is in SpellID order

Spellbook order matters for monsters


LtRipley36706 (approved):

Source/ACE.Database/WorldDatabase.cs
```diff
@@ -107,7 +107,7 @@ public virtual Weenie GetWeenie(WorldDbContext context, uint weenieClassId)
                 weenie.WeeniePropertiesSkill = context.WeeniePropertiesSkill.Where(r => r.ObjectId == weenie.ClassId).ToList();
             }
 
-            weenie.WeeniePropertiesSpellBook = context.WeeniePropertiesSpellBook.Where(r => r.ObjectId == weenie.ClassId).ToList();
+            weenie.WeeniePropertiesSpellBook = context.WeeniePropertiesSpellBook.Where(r => r.ObjectId == weenie.ClassId).OrderBy(i => i.Id).ToList();
```
Mag-nus, line 110: I'd really rather fix this in the creature code where it does the spell work where order by id is required.

Putting this code here is just confusing because it goes into an ICollection, then a Dictionary.

However, if this is the allowed approach temporarily, it should at least be commented as to why were ordering by id. It goes against the pattern here and will easily be missed in the future if this code is reworked and/or adapted.

gmriggs: this really is a bug for all spellbooks. spellbooks should be read in order, from the database. not sorted by ascending spell order

LtRipley36706 (approved):

PR #3315, opened 2020-10-19
Landblock SpawnEncounters threading improvement

Mag-nus:
Landblock.SpawnEncounters is run from a separate Task.

It does a lot of position/cell work on that task and only enqueues the AddWorldObject onto the thread safe ActionQueue.

I suspect these tasks were hitting AdjustCells at the same time LandblockManager is unloading landblocks (and thus cells).

This caused thread collisions on the AdjustCells dictionary.

Wrapping all this work up into the thread safe ActionQueue should solve the issue.

### DIFF
--- a/Source/ACE.Server/Entity/Landblock.cs
+++ b/Source/ACE.Server/Entity/Landblock.cs
@@ -272,11 +272,11 @@ namespace ACE.Server.Entity
 
                 if (wo == null) continue;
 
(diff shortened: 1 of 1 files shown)


Mag-nus:
While the adjustcell fix is in the concurrentDictionary PR, this should still go in.

gmriggs (approved):

PR #3324, opened 2020-10-23
improving /addenc, adding /removeenc

gmriggs:
- the .sql output of /addenc is now more consistent with /createinst, in that it adds the encounter to the db w/ checks, and then outputs the entire state of the encounters in the db for that landblock to the sql file. this includes the delete statement, and any pre-existing encounters that were already in the database

- a /removeenc command has been added, which works on the last appraised object

- cleaned up both /removeenc and /removeinst to properly handle removal when there is only 1 encounter / instance in the landblock remaining

### DIFF
--- a/Source/ACE.Database/WorldDatabaseWithEntityCache.cs
+++ b/Source/ACE.Database/WorldDatabaseWithEntityCache.cs
@@ -379,6 +379,11 @@ namespace ACE.Database
             }
         }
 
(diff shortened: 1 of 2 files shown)


LtRipley36706 (approved):

DridsGhost:
Works like a dream! Much gratitude for the addtion!

PR #3380, opened 2020-12-04
Update Docker files due to NET 5.0 changes

LtRipley36706:
See https://github.com/dotnet/dotnet-docker/issues/2375


LtRipley36706:
This does not actually change to NET 5.0 sdk/runtime, just realigns the names for eventual 5.0 changeover

PR #3391, opened 2020-12-15
Send current attacker message to client.

jspashett:
Send the current attacker message to the client so that the client can
use the last attack keyboard shortcut to select the last attacker.

Note that these will be sent often when there are multiple attackers.
The message should only be sent when the attacker changes

### DIFF
--- a/Source/ACE.Server/Entity/Spell.cs
+++ b/Source/ACE.Server/Entity/Spell.cs
@@ -238,11 +238,16 @@ namespace ACE.Server.Entity
         {
             get
             {
-                return Category == SpellCategory.DamageLowering     // encompasses both blood and spirit loather, inconsistent with spirit drinker in dat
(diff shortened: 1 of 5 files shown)


jspashett:
I seems it might be the case that this message will be sent when a projectile is fired from a creature, rather than if/when it hits a player. That would be wrong.

jspashett:
> 
> 
> I seems it might be the case that this message will be sent when a projectile is fired from a creature, rather than if/when it hits a player. That would be wrong.

This is incorrect. I cannot currently find a place to issue this message without sprikling if statements around the code base. .e.g TryResistSpell should sometimes issue the message, as well as many other places. There doesn't a appear to be a (single)place / callback to capture an incoming attack on a creature/player

gmriggs (approved):

PR #3416, opened 2021-01-10
Update console title to show server name

FlaggAC:
This is very useful for anyone running multiple servers on the same machine (particularly windows servers).

### DIFF
--- a/Source/ACE.Server/Program.cs
+++ b/Source/ACE.Server/Program.cs
@@ -122,8 +122,6 @@ namespace ACE.Server
             if (IsRunningInContainer)
                 log.Info("ACEmulator is running in a container...");
 
-            Console.Title = @$"ACEmulator - v{ServerBuildInfo.FullVersion}";
-
             var configFile = Path.Combine(exeLocation, "Config.js");
             var configConfigContainer = Path.Combine(containerConfigDirectory, "Config.js");
 
@@ -148,6 +146,8 @@ namespace ACE.Server
 
             log.Info("Initializing ConfigManager...");
             ConfigManager.Initialize();
+
+            Console.Title = @$"{ConfigManager.Config.Server.WorldName} - v{ServerBuildInfo.FullVersion}";
 
             if (ConfigManager.Config.Offline.PurgeDeletedCharacters)
             {
--- a/Source/ACE.Server/Program.cs
+++ b/Source/ACE.Server/Program.cs
@@ -147,7 +147,10 @@ namespace ACE.Server
             log.Info("Initializing ConfigManager...");
             ConfigManager.Initialize();
 
-            Console.Title = @$"{ConfigManager.Config.Server.WorldName} - v{ServerBuildInfo.FullVersion}";
+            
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Program.cs
```diff
@@ -149,6 +147,8 @@ public static void Main(string[] args)
             log.Info("Initializing ConfigManager...");
             ConfigManager.Initialize();
 
+            Console.Title = @$"{ConfigManager.Config.Server.WorldName} - v{ServerBuildInfo.FullVersion}";
```
LtRipley36706, line 150:
```suggestion
            Console.Title = @$"ACEmulator - v{ServerBuildInfo.FullVersion}:  {ConfigManager.Config.Server.WorldName}";
```

gmriggs (approved):
at first i was concerned with the removal of ACEmulator, and was going to suggest ACEmulator - WorldName, or WorldName - ACEmulator

But I see now the default WorldName in Config.js is ACEmulator. This should be fine to me

Source/ACE.Server/Program.cs
gmriggs: this will show ACEmulator - v#: ACEmulator for the default scenario, which seems kind of messy

We should decide on some kind of definitive formatting though, since Window title is something important

LtRipley36706:
> at first i was concerned with the removal of ACEmulator, and was going to suggest ACEmulator - WorldName, or WorldName - ACEmulator
> 
> But I see now the default WorldName in Config.js is ACEmulator. This should be fine to me

I'd prefer the name still be there, ideally probably better to split it up in to two changes. first one as-is, second one append world name if its different from default?

gmriggs:
> I'd prefer the name still be there, ideally probably better to split it up in to two changes. first one as-is, second one append world name if its different from default?

Yeah something like that

LtRipley36706 (changes requested):
I can't make changes to the PR but this is what I would propose:
starting at line 125:

```cs
            Console.Title = @$"ACEmulator - v{ServerBuildInfo.FullVersion}";
            
            var configFile = Path.Combine(exeLocation, "Config.js");
            var configConfigContainer = Path.Combine(containerConfigDirectory, "Config.js");

            if (IsRunningInContainer && File.Exists(configConfigContainer))
                File.Copy(configConfigContainer, configFile, true);

            if (!File.Exists(configFile))
            {
                if (!IsRunningInContainer)
                    DoOutOfBoxSetup(configFile);
                else
                {
                    if (!File.Exists(configConfigContainer))
                    {
                        DoOutOfBoxSetup(configFile);
                        File.Copy(configFile, configConfigContainer);
                    }
                    else
                        File.Copy(configConfigContainer, configFile);
                }
            }

            log.Info("Initializing ConfigManager...");
            ConfigManager.Initialize();

            if (ConfigManager.Config.Server.WorldName != "ACEmulator")
                Console.Title = @$"{ConfigManager.Config.Server.WorldName} | {Console.Title}";
```

FlaggAC:
been busy irl but ill make the changes here when I can

LtRipley36706 (approved):

PR #3419, opened 2021-01-10
add support for Hatred1Bits

gmriggs:


gmriggs:
content ended up being implemented without requiring this feature

PR #3420, opened 2021-01-10
Prevent future instances of dynamic GUIDs in World Database

LtRipley36706:


LtRipley36706:
This code cannot go in until after world database 209 is released, to ensure proper support going forward.

Mag-nus (approved):

PR #3423, opened 2021-01-14
rush next player save after successful tailoring operation

gmriggs:


LtRipley36706 (dismissed):

Mag-nus:
This doesn't solve the problem. The race condition will still exist and items will be lost.

If you absolutely insist this go into master, please use the following condition, which we use elshewhere for this bandaid.

 if (PropertyManager.GetBool("player_receive_immediate_save").Item)

Mag-nus (approved):

PR #3435, opened 2021-01-21
Fix query issue with recent MySQL versions

LtRipley36706:
Resolves GUID recapture issue with MySQL 8.0.22+

### DIFF
--- a/Source/ACE.Database/ShardDatabase.cs
+++ b/Source/ACE.Database/ShardDatabase.cs
@@ -88,7 +88,8 @@ namespace ACE.Database
             // https://stackoverflow.com/questions/50402015/how-to-execute-sqlquery-with-entity-framework-core-2-1
 
             // This query is ugly, but very fast.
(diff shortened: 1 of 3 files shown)


Source/ACE.Database/ShardDatabase.cs
```diff
@@ -115,7 +116,7 @@ public List<(uint start, uint end)> GetSequenceGaps(uint min, uint limitAvailabl
 
                 while (reader.Read())
                 {
-                    var gap_starts_at               = reader.GetFieldValue<double>(0);
+                    var gap_starts_at               = reader.GetFieldValue<long>(0);
                     var gap_ends_at_not_inclusive   = reader.GetFieldValue<decimal>(1);
```
Mag-nus, line 120: Should the second field be long as well? I think so.

LtRipley36706: according to the ready the second field is still decimal, and the third was unchanged as well

Mag-nus (approved):

PR #3440, opened 2021-01-28
Force "lost" session characters to log out on shutdown

LtRipley36706:
Prevent server from failing to finish clean shutdown due to bugged sessionless characters not fully logging out


Mag-nus:
Something is not right here.

A player should never have a null session. It's set in the ctor of the Player and never released.

I suspect this is trying to fix custom code? I would prefer not to see fixes for 3rd paty code put into the core of ACE.

More specifically, this code gives the indication to a developer that a Player.Session can be null, when in fact, it cannot.

PR #3445, opened 2021-02-03
Add TryFixExceptionOnSave to workaround SaveCharacter desync fails

LtRipley36706:


Mag-nus:
I don't believe the issue is in the db here.

I think that the ACE object ACE.Database.Models.Shard.Character.CharacterPropertiesQuestRegistry collection is getting messed up and ending up with two entries that have the same QuestName property.

As a temp fix, I would:
- Iterate through the collection and log in detail with ERROR each record that has the QuestName in question.
- Remove one of the quest name entries and log the action.
- Try to save the character again.

It's either that, or, the Character EF tracking is some how getting out of sync and EF can't recover. I'd try the above first though before we dig deeper.

PR #3446, opened 2021-02-04
Add support for servers to periodically communicate with external services

amoeba:
Hey all, I was talking about something I've built with @LtRipley36706 the other day and he thought other devs would be interested in seeing a PR so here it is:

It would be really interesting if ACE servers could communicate with external services. There's been a ton of discussion about _how_ to do this, going back to https://github.com/ACEmulator/ACE/issues/718. Providing an HTTP API is a really powerful idea but presents a ton of problems such as rate limiting, running read-only replicas of the database, and so on.

A lighter weight and simpler approach is to have ACE servers _send_ any information to external services periodically. The biggest downside here is that the feature only works if the external service stays available.

What I've done here is created a new `*Manager` class called `HeartbeatManager` that based on code that already existed in ACE that sends a JSON payload to a configurable HTTP endpoint at a configurable interval. Think [webhook](https://en.wikipedia.org/wiki/Webhook).

The immediate thing ACE servers could do with this is broadcast whether the server is up and how many players are online to services such as https://treestats.net. People are very interested in how many players are on each server but the TreeStats way of tracking [server populations](https://treestats.net/player_counts/) requires players on each server to run my Decal plugin. And even then, the temporal resolution of the population data depends on how many players are running the plugin. With my system here, ACE servers could send their `OnlineCount` to TreeStats at intervals and we'd all have much better data to look at.

Some notes:

1. This PR isn't done, I'm having some issues figuring out default values in the config
2. The payload is hardcoded to send `WorldName` and `OnlineCount` but I think we could find some easy way to let users configure the Manager to send more info such as any config properties (xp rate, pk/npk, etc) or even information obtained from database queries such as total characters.
3. This type of approach could even be used to send webhooks to Discord servers, Google Sheets, etc
3. I'm open to another name. Maybe even `WebhookManager`

What do you all think? I'm open to both thoughts on the concept as well as specific comments on the PR.

### DIFF
--- a/Source/ACE.Server/Managers/HeartbeatManager.cs
+++ b/Source/ACE.Server/Managers/HeartbeatManager.cs
@@ -88,7 +88,7 @@ namespace ACE.Server.Managers
                 }
                 catch (Exception e)
                 {
-                    log.Debug("Exception: " + e.Message);
+                    log.Debug("Exception while sending Heartbeat: " + e.Message);
                 }
             }
         }
--- a/Source/ACE.Server/Managers/HeartbeatManager.cs
+++ b/Source/ACE.Server/Managers/HeartbeatManager.cs
@@ -17,19 +17,31 @@ namespace ACE.Server.Managers
         /// <summary>
         /// The rate at which HeartbeatManager.Tick() executes
         /// </summary>
-        private static readonly RateLimiter updateHeartbeatManagerRateLimiter = new RateLimiter(1, TimeSpan.FromSeconds(10));
+        private static RateLimiter updateHeartbeatManagerRateLimiter;
 
-        private static Uri heartbeatUri;
+        /// <summary>
+        /// Endpoint to send heartbeats to
+        /// </summary>
+        private static Uri endpoint;
 
         public static void Initialize()
         {
+            //if (!ConfigManager.Config.Server.Heartbeat.Enabled)
+            //{
+            //    log.Debug("Not starting HeartbeatManager because it's disabled in config");
 
-            // TODO: Factor out into config
-            heartbeatUri = new Uri("https://servers.treestats.net/heartbeat");
+            //    return;
+            //}
+
+            // endpoint = new Uri(ConfigManager.Config.Server.Heartbeat.Endpoint);
+            endpoint = new Uri("https://treestats-servers.herokuapp.com/");
+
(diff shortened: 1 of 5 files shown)


gmriggs:
👍 to the idea for sure!

Maybe someone can ping @FlaggAC about this as well... he brought up a discussion a few weeks ago about integrating the rcon protocol

amoeba:
> 👍 to the idea for sure!
> 
> Maybe someone can ping @FlaggAC about this as well... he brought up a discussion a few weeks ago about integrating the rcon protocol

Interesting, @gmriggs. Thanks for the pointer. I hadn't heard about that work but that's really cool.

Source/ACE.Server/Managers/HeartbeatManager.cs
```diff
@@ -0,0 +1,96 @@
...
+                try
+                {
+                    HeartbeatPayload body = new HeartbeatPayload
+                    {
+                        OnlineCount = PlayerManager.GetOnlineCount(),
+                        WorldName = ConfigManager.Config.Server.WorldName
+                    };
+
+                    HttpContent content = new StringContent(JsonConvert.SerializeObject(body));
+                    response = client.PostAsync(endpoint, content).Result;
```
enknamel, line 82: Accessing Result will block. Is it safe to do this on this thread?

amoeba: Yeah, I thought this was fine but can change if this is problematic. Thanks for catching this @enknamel

FlaggAC:
Hey yes I spoke with the owner of battlemetrics.com and it might be a good thing for any AC Servers that wish to opt in to this. It would show public chat history, player login history, etc. - You can get an idea by going to https://www.battlemetrics.com/subscription/rcon  and seeing what it offers. 

Feel free to ping me in Discord if you want to work together on this - the guy sent me info on how to get it completely set up.

amoeba:
That's a pretty neat service, @FlaggAC , thanks for the pointer. Gives me some ideas for things I could add onto TreeStats.

I definitely like the idea of implementing something like rcon that gives servers two-way messaging and remote control. That said, I also like the relatively small size of the change I'm proposing here and am probably willing to put more time into it since it's scratching my own itch.

PR #3448, opened 2021-02-05
Fix crash on Vendor inventory load

LtRipley36706:


Source/ACE.Server/WorldObjects/Vendor.cs
```diff
@@ -230,7 +230,7 @@ private void LoadInventory()
                 }
             }
 
-            if (!PropertyManager.GetBool("vendor_shop_uses_generator").Item)
+            if (!PropertyManager.GetBool("vendor_shop_uses_generator").Item && Biota.PropertiesGenerator != null)
```
gmriggs, line 233: null check should be first

PR #3451, opened 2021-02-07
Attempt Z position adjustments on failed spawns

LtRipley36706:
Prior to #3447 there were corpses that would place items they contained at their same LOC. This didn't work for all the items, so there are many of these items "stuck" in limbo. This attempts to resolve the limbo for many, if not most, of them

### DIFF
--- a/Source/ACE.Server/Entity/Landblock.cs
+++ b/Source/ACE.Server/Entity/Landblock.cs
@@ -810,7 +810,7 @@ namespace ACE.Server.Entity
 
         private bool AddWorldObjectInternal(WorldObject wo)
         {
-            for (var i = 0; i < 6; i++)
+            for (var i = 0; i <= 5; i++)
             {
                 wo.CurrentLandblock = this;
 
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Entity/Landblock.cs
```diff
@@ -824,18 +825,40 @@ private bool AddWorldObjectInternal(WorldObject wo)
...
+                    {
+                        log.Debug($"AddWorldObjectInternal: couldn't spawn 0x{wo.Guid}:{wo.Name} [{wo.WeenieClassId} - {wo.WeenieType}]{(retried ? $", including {retries} retries," : "")} at {wo.Location.ToLOCString()}");
+
+                        var retryLOC = new Position(wo.Location);
+                        retryLOC.PositionZ += 0.05f * (wo.ObjScale ?? 1.0f);
+                        wo.Location = retryLOC;
+
+                        addWorldObjectInternalRetries[wo.Guid.Full] = retries + 1;
+
+                        goto retry;
```
gmriggs, line 850: no goto

gmriggs (changes requested):

PR #3460, opened 2021-02-11
Give ability to change what level fellowship XP is shared evenly regardless of level

Delubear:


gmriggs:
This looks like it should be fine to me

Only suggestion is to alphabetize the property in the list, so that the property list is easily manageable / understandable, and so people reading through the properties can easily find them

gmriggs (approved):

Delubear:
> This looks like it should be fine to me
> 
> Only suggestion is to alphabetize the property in the list, so that the property list is easily manageable / understandable, and so people reading through the properties can easily find them

Fixed the position in the list.

Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -454,9 +454,10 @@ private void CalculateXPSharing()
 
             var fellows = GetFellowshipMembers();
 
-            var allOver50 = !fellows.Values.Any(f => (f.Level ?? 1) < 50);
+            var allEvenShareLevel = PropertyManager.GetLong("fellowship_even_share_level", 50).Item;
```
LtRipley36706, line 457:
```suggestion
            var allEvenShareLevel = PropertyManager.GetLong("fellowship_even_share_level").Item;
```

LtRipley36706 (approved):

gmriggs (approved):

PR #3472, opened 2021-02-22
Endpoint Mismatch should be debug, not warn

Mag-nus:
This can happen when clients are connected to a server.
Those clients remain untouched.
Server crashes and restarts.
NEW clients connect.
The original clients are still open and trying to talk to the server, resulting in this message.


gmriggs (approved):

ziang4891:
Should those old clients be forced to end, as I'm guessing that the server doesn't recognize them as valid clients, anymore?  Will they end on their own, when not getting any response from the restarted server?

PR #3502, opened 2021-03-17
Fix issue with vtank corpse loot stall

LtRipley36706:


Source/ACE.Server/WorldObjects/Container.cs
```diff
@@ -712,7 +712,11 @@ public string InUseMessage
 
         public virtual void Open(Player player)
         {
-            if (IsOpen) return;
+            if (IsOpen)
+            {
+                player?.SendTransientError(InUseMessage);
```
gmriggs, line 717: shouldn't be using player?

should just be player

gmriggs (changes requested):

gmriggs (approved):

PR #3528, opened 2021-04-21
Apply 2014 Aegis PvP nerf

FlaggAC:


Source/ACE.Server/WorldObjects/SpellProjectile.cs
```diff
@@ -437,6 +437,14 @@ public override void OnCollideObject(WorldObject target)
 
             bool isPVP = sourcePlayer != null && targetPlayer != null;
 
+            //http://acpedia.org/wiki/Announcements_-_2014/01_-_Forces_of_Nature - Aegis is 72% effective in PvP
```
gmriggs, line 440: would prefer if this were downstream further in the GetAbsorbMod functions, but should be good enough

to url link seems to have some of the _'s replaced with spaces

gmriggs (approved):

PR #3543, opened 2021-05-17
adding portal re-use timer

gmriggs:
This could potentially fix some errors with players claiming they are getting teleported out of some dungeons immediately when they teleport in

This adds the 'You have been teleported too recently!' message similar to retail

The implementation for this isn't the cleanest, however it evolved from some earlier versions where the code was slightly cleaner, but had gameplay issues during playtesting. I opted here to do what was called for in code to make everything feel solid in-game.


Mag-nus:
Should PropertyFloat.LastPortalTeleportTimestamp be marked [Ephemeral] so that it's not saved to the database?

gmriggs:
> Should PropertyFloat.LastPortalTeleportTimestamp be marked [Ephemeral] so that it's not saved to the database?

Yes, that might be a good idea

That could also solve one of the main issues I ran into when applying this logic to all teleports, instead of just portals. Technically logging in counts as a "teleport", or at least the client sends the same OnLoginComplete during the initial login that it also sends  when it's ready to exit portal space from a regular portal.

I initially ran into an issue where a player logging in and then running immediately into a portal would trigger the 'you have teleported too recently' message, so it was further restricted to being applied just to portals

Mag-nus (approved):

PR #3566, opened 2021-05-30
refactor vendor code, add support for PropertyInt.Unique on vendors

gmriggs:
This started out as a simple task of adding PropertyInt.Unique support for players purchasing items from vendors.

I found that adding this feature was more difficult than it should have been, due to the way the existing vendor code was written. I also started to find issues with the existing vendor code.

(note that when referring to buy/sell, it is always from the perspective of the player)

- The handling of players purchasing Unique items (items sold to the vendor by other players, not to be confused with PropertyInt.Unique) was very convoluted. The main issue is that Unique items were removed from the vendor list *way* too early in the process: during initial validation. If the validation failed at any point afterwards (and there are many, many ways it can fail afterwards) the vendor then has to add the unique item back to its Uniques list.

There were gaps in the code where errors could occur, and the Unique item was never added back to the vendor's Uniques list. This was all pointless mangagement. The item shouldn't be removed from the vendor's Uniques list until all validation has passed, and no more errors can occur, when the player actually purchases the item.

- The feature to ensure a player has enough inventory space / side container slots / burden capacity to receive items was implemented decently enough, with only a couple of bugs, however the main issue with this code is that it was copy/pasted in various parts all throughout the system, with only a few minor changes in some of the parts (such as the messages the player receives on failure). There should be no code duplication for the common parts, and only the parts that differ in each section should be re-specified.

A new helper class has been added for this, ItemsToReceive. This greatly reduces the amount of duplicate code in all of the places that use it: Vendor, Player_Commerce, Player_Inventory, Advocate, and Gem.

- The AllItems collection for vendors was a useful idea in theory. However, it was implemented inefficiently, and was only used in 2 places throughout the entire system. (approaching the vendor, which is used when the player double-clicks the vendor, and when a transaction has completed to get the latest inventory, which was subsequently used by RotUniques, even though it only needed to iterate over the Uniques, and finally by FindObject, which had a flag to check for the last used vendor).

The core collections are stored separately as DefaultItems and UniqueItems, and AllItems was a getter that dynamically built a new combined list of DefaultItems and UniqueItems each time it was called. Removing AllItems and replacing it with a similarly easy-to-use but more efficient concept was fairly straightforward.

For vendor approach and anything else that needs to iterate over all items in the vendor's inventory, forEachItem() now iterates over both collections:

Previous code:
```
foreach (var item in vendor.AllItems)
{
	DoSomething(item);
}
```
New code:
```
vendor.forEachItem((item) =>
{
	DoSomething(item);
});
```
Similarly, for FindObject:
```
vendor.AllItems.TryGetValue(objectGuid, out var item);
```
Becomes:
```
vendor.TryGetItemForSale(objectGuid, out var item);
```
So a similar interface, but the underlying implementation is much more efficient, without recreating a duplicate / combined collection on each call.

- The overly confusing buy data flow:

Player.HandleActionBuyItem -> Vendor.BuyItems_ValidateTransaction -> Player.FinalizeBuyTransaction -> Vendor.BuyItems_FinalTransaction

has been cleaned up. Vendor.BuyItems_FinalTransaction was entirely unneeded, and all it was doing was maintaining Uniques on (some) failure causes. Player.FinalizeBuyTransaction also had a confusing aspect where it performed 1 last final validation, which should have been in the prior validation method. There is no reason to split this validation logic up. All of the validation logic should be accessible in 1 place.

Vendor.BuyItems_ValidateTransaction still chains to Player.FinalizeBuyTransaction, which is still somewhat confusing, however this is commented well throughout, to make it clear and readily apparent whats going on, so the reader of the code is aware of the control flow no matter where they are viewing the code from. It would be nice if the validate() function could return a bool, and then the original caller could continue onwards. However, this would require the validation function returning many different out params, so the existing structure was kept in place there.

- The order of the functions is now in a more logical order that is in line with the natural control flow, and the order they are called in the code. Similar functions are grouped together, instead of randomly scattered throughout. This unfortunately makes the git diff a bit confusing to compare.

- Vendor.LastPlayer has been replaced with Vendor.LastPlayerInfo, to ensure the vendor is only holding onto a WorldObjectInfo weak reference to the player. I don't think the vendor would have held onto any player references for any extended lengths of time, but it's always good to ensure this can never happen.

- Vendor.LoadInventory was cleaned up. Again, duplicated code here. 2 giant loops that were doing almost the exact same thing (maybe even literally the exact same thing?) This code has also been optimized where it can perform the dupe wcid/palette/shade combo check before ever instantiating the WorldObject.

- Missing calls to WorldObject.Destroy() on error scenarios have been added for GuidManager

- Stackable logic has been cleaned up and simplified. There where also some scenarios where stackables were being handled inconsistently. ie., if a worldObject is Stackable, however its MaxStackSize is null for whatever reason, most of the code was considering this to be the same as MaxStackSize being 1, and other code was handling this in odd ways, sometimes equating it to being an infinite MaxStackSize. This has been normalized in those places to the rest of the system.

- ItemProfile now reads amount in as int instead of uint, as per the client and protocol specs. ItemProfile now has a normalized method for ensuring validity, and verification of amount > 0

- VerifySellItems() has been normalized to include all of the verifications. Previously, the verification of items the player is attempting to sell to the vendor was split up into 2 places -- VerifySellItems(), and inline logic in HandleActionSellItem() afterwards. There was really no need for this 2-step process before, and there were even some redundancies between them. All of the verification logic is in 1 method now, and there are some performance improvements as well.

There is an ongoing process for reviewing this code and testing everything, but everything has looked good enough for awhile now to push out the PR for others to review / test as well


ziang4891 (approved):

LtRipley36706 (dismissed):

LtRipley36706 (approved):

ziang4891 (approved):

LtRipley36706:
code on CE is stable with no issues reported

PR #3582, opened 2021-06-10
improve precaching performance from 5 mins -> 5 seconds

gmriggs:


ziang4891:
Difference is night and day.  I wholeheartedly approve of this change

PR #3587, opened 2021-06-12
Improve World DB Startup Caching

Mag-nus:
This is significantly faster than what is currently in master, and just slightly slower than PR #3582

It is now a single threaded operation, just like #3582 but instead of stitching the objects together manually, it lets EF stitch the objects together.

EF Tracking is required for the stitching to work.

This is the recommended pattern for these types of actions when dealing with EF.

I think similar improvements can be made to startup player caching

### DIFF
--- a/Source/ACE.Database/WorldDatabase.cs
+++ b/Source/ACE.Database/WorldDatabase.cs
@@ -117,6 +117,62 @@ namespace ACE.Database
         /// <summary>
         /// This will populate all sub collections except the following: LandblockInstances, PointsOfInterest
         /// </summary>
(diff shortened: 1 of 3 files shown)


gmriggs:
measured on 2600k:

master: ~180s
https://github.com/ACEmulator/ACE/pull/3587: 18s
https://github.com/ACEmulator/ACE/pull/3582: 5s

10x improvement from master, but still 3-4x slower than https://github.com/ACEmulator/ACE/pull/3582

still 10x improvement from master, so marking as approved

gmriggs (approved):

PR #3590, opened 2021-06-13
update pass interval

gmriggs:


Source/ACE.Server/WorldObjects/Player.cs
```diff
@@ -1154,5 +1154,10 @@ public void HandleActionEnterPkLite()
             });
             actionChain.EnqueueChain();
         }
+
+        /// <summary>
+        /// Rate limiter for /passwd command
+        /// </summary>
+        public DateTime LastPassTime { get; set; }
```
Mag-nus, line 1161: I think this should be a Session var, not a Player var.

Mag-nus (changes requested):

Mag-nus (approved):

PR #3612, opened 2021-07-11
gmriggs objcell patch

Mag-nus:


gmriggs:
as mentioned, not 100% confident emptycell isn't further modified downstream, and since readonly doesn't truly enforce no changes to the object, can't be certain without further evaluation of the calling functions

gmriggs (approved):

PR #3618, opened 2021-07-16
fix unintended multidamage

gmriggs:


ziang4891 (approved):
Tested, and it does mitigate the issue caused by incorrect data with DamageType combined with MultiStrike

LtRipley36706 (approved):

PR #3619, opened 2021-07-17
Add and switch to EMU time for Game World

LtRipley36706:
This change switches PortalYearTicks from UtcNowToLoreTime to a new time called UtcNowToEMUTime.

This switch results in time always rolling forward, as well as syncing all emulated servers up to the same source. The in-game calendar, day-night cycles and weather patterns would be identical across all worlds going forward. Previously there was varied results depending on when the server started up as the "seed" changed at that time. 

As a side benefit to this change, the Journal section of the client (Quill, Journal pages with timers) will now track correctly should anyone wish to use that feature, which previously was unreliable since server time would roll backwards upon restart.


JohnnyCodin:
I realize this is an old change, but did this affect Calendar_Events as well?

PR #3624, opened 2021-07-21
Fix issue with TimeZones on platforms other than Windows

LtRipley36706:
See [here](https://devblogs.microsoft.com/dotnet/cross-platform-time-zones-with-net-core/) for issue explanation 

Adding and using [this](https://github.com/mattjohnsonpint/TimeZoneConverter) nuget package to resolve until .NET6

> .NET 6 will have built-in support for IANA and Windows time zones in a cross-platform manner, removing the need for the TimeZoneConverter library. If you are planning to use .NET 6 (or higher), you don't need to use the TimeZoneConverter library!

### DIFF
--- a/Source/ACE.Common/DerethDateTime.cs
+++ b/Source/ACE.Common/DerethDateTime.cs
@@ -1,5 +1,7 @@
 using System;
 
+using TimeZoneConverter;
+
 namespace ACE.Common
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


gmriggs (approved):

Source/ACE.Common/DerethDateTime.cs
```diff
@@ -1059,6 +1061,6 @@ public static DerethDateTime ConvertRealWorldToLoreDateTime(DateTime dateTime)
         /// <summary>
         /// Converts the <see cref="DateTime.UtcNow"/> object to a new <see cref="DerethDateTime"/> object set to EMU Standard Sync Time.
         /// </summary>
-        public static DerethDateTime UtcNowToEMUTime => new DerethDateTime((DateTime.UtcNow - TimeZoneInfo.ConvertTimeToUtc(retailDayLast_RealWorld, TimeZoneInfo.FindSystemTimeZoneById("Eastern Standard Time"))).TotalSeconds);
+        public static DerethDateTime UtcNowToEMUTime => new DerethDateTime((DateTime.UtcNow - TimeZoneInfo.ConvertTimeToUtc(retailDayLast_RealWorld, TZConvert.GetTimeZoneInfo("Eastern Standard Time"))).TotalSeconds);
```
gmriggs, line 1064: a specific timezone shouldn't be hardcoded into the emu like this.

if anything, this should be UTC.

gmriggs: the more i look at this, the more it just seems wrong

who originally approved this pr?

answer: no one. and this is why we have code reviews, and this is why this practice of checking in prs without approval was supposed to be banned long ago...

PR #3626, opened 2021-07-22
Add "You Chicken Out." response to crafting cancellation

LtRipley36706:


Mag-nus:
I think you can move up the lines of "if (!response) return;" above the "var player = Player;".

This will save calls to PalyerManager.GetOnlinePlayer when !response

LtRipley36706:
> I think you can move up the lines of "if (!response) return;" above the "var player = Player;".
> 
> This will save calls to PalyerManager.GetOnlinePlayer when !response

shuffled around the ones that could be

Mag-nus (approved):

PR #3630, opened 2021-07-27
Add motion to Fellowship recruiting

LtRipley36706:
BowDeep on the recruiter occurs on successful recruitment

### DIFF
--- a/Source/ACE.Server/Entity/Fellowship.cs
+++ b/Source/ACE.Server/Entity/Fellowship.cs
@@ -174,6 +174,8 @@ namespace ACE.Server.Entity
             }
 
             UpdateAllMembers();
+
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -174,6 +174,9 @@ public void AddConfirmedMember(Player inviter, Player player, bool response)
             }
 
             UpdateAllMembers();
+
+            if (inviter.CurrentMotionState.Stance == MotionStance.NonCombat) // only do this motion if inviter is at peace, other times motion is skipped.
```
gmriggs, line 178: the standard way to do this would be inviter.CombatMode == CombatMode.NonCombat, but this should still be fine

gmriggs (approved):

PR #3634, opened 2021-08-02
Re-verify items in Secure Trade before executing trade

LtRipley36706:


LtRipley36706:
running already on CE.

harliq (approved):

PR #3650, opened 2021-08-15
Generator System Revision 4 - Profile Delays

LtRipley36706:


Source/ACE.Server/Entity/GeneratorProfile.cs
```diff
@@ -71,7 +72,7 @@ public int CurrentCreate
             get
             {
                 if (!GeneratedTreasureItem)
-                    return Spawned.Count + SpawnQueue.Count;
+                    return Spawned.Count + SpawnQueue.Count - RemoveQueue.Count(r => Spawned.ContainsKey(r.objectGuid));
```
gmriggs, line 75: this is doing an o(n) lookup, which is no good. should be transformed into o(1)

The logic is also written in a way while it might technically be producing the correct outcome, semantically it is incorrect. It shouldn't be "counting a bool", it should be returning 1 if something exists, and 0 if something doesn't exist.

Source/ACE.Server/Entity/GeneratorProfile.cs
```diff
@@ -174,6 +172,8 @@ public DateTime GetSpawnTime()
         /// </summary>
         public void Enqueue(int numObjects = 1)
         {
+            if (numObjects < 1) return;
```
gmriggs, line 175: if CurrentCreateCache is no longer a thing, then this line should be removed

LtRipley36706: it was no longer necessary. CurrentCreate is now back to simple counts
```cs
        /// <summary>
        /// The total # of active spawned objects + awaiting spawning
        /// </summary>
        public int CurrentCreate
        {
            get
            {
                if (!GeneratedTreasureItem)
                {
                    return Spawned.Count + SpawnQueue.Count;
                }
                else
                {
                    if (Spawned.Count > 0 || SpawnQueue.Count > 0)
                        return 1;
                    else
                        return 0;
                }
            }
        }
```

PR #3686, opened 2021-10-02
Adjust Hook worldobject to catch null crashes

LtRipley36706:


Source/ACE.Server/WorldObjects/Hook.cs
```diff
@@ -27,7 +27,7 @@ public class Hook : Container
 
         public bool HasItem => Inventory != null && Inventory.Count > 0;
 
-        public WorldObject Item => Inventory != null ? Inventory.Values.FirstOrDefault() : null;
+        public WorldObject Item => Inventory?.Values.FirstOrDefault();
```
gmriggs, line 30: Should not have anything related to null around it.

It is impossible for Inventory to be null

Source/ACE.Server/WorldObjects/Hook.cs
```diff
@@ -62,7 +62,7 @@ public override ActivationResult CheckUseRequirements(WorldObject activator)
             var rootHouse = House.RootHouse;
             var houseOwner = rootHouse.HouseOwner;
 
-            var houseHooksVisible = rootHouse.HouseHooksVisible ?? true;
+            var houseHooksVisible = rootHouse?.HouseHooksVisible ?? true;
```
gmriggs, line 65: This check is pointless. The line immediately above where it says rootHouse.HouseOwner will have already crashed.

Source/ACE.Server/WorldObjects/Hook.cs
```diff
@@ -236,7 +236,7 @@ public void UpdateHookVisibility()
         {
             if (!HasItem)
             {
-                if (!(House.HouseHooksVisible ?? false))
+                if (!(House?.RootHouse?.HouseHooksVisible ?? false))
```
gmriggs, line 239: I was okay with putting this around 1 place where the actual crash occurred, and maybe some logging to find out *why* that happened, and which specific house was null, however i'm not as okay with sprinkling this logic around randomly throughout the program. House and RootHouse shouldn't be null. All of this code implies that they can be null, and no explanation as to why.

PR #3728, opened 2021-11-01
Add Thread Safety to AllegianceManager PassXP

Mag-nus:
This follows the same thread-safety pattern used in other parts of AllegianceManager, like HandlePlayerDelete.

The intent here is that AllegianceManager should mange its own thread safety. Callers shouldn't have to worry about if an AllegianceManager action should be thread safe or not.


Mag-nus:
This needs testing.

gmriggs (approved):

PR #3739, opened 2021-11-11
Update to .NET6

LtRipley36706:
This updates ACEmulator server and test projects to .NET6, as well as updates latest versions of many shared packages for solution.

This also includes changes to remove obsoleted WebRequest and HttpWebRequest code.


Mag-nus:
Right now .NET 6 is showing much larger GC pauses that can occasionally "lock up" the server. In testing I have seen pauses of 5s - 7s.

PR #3745, opened 2021-11-23
Wire up support for RotationSpeed

LtRipley36706:
Co-Authored-By: gmriggs <8909245+gmriggs@users.noreply.github.com>

### DIFF
--- a/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
+++ b/Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
@@ -3068,7 +3068,7 @@ namespace ACE.Server.WorldObjects
             set { if (!value.HasValue) RemoveProperty(PropertyFloat.RotationSpeed); else SetProperty(PropertyFloat.RotationSpeed, value.Value); }
         }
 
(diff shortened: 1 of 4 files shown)


Source/ACE.Server/WorldObjects/WorldObject_Properties.cs
```diff
@@ -3059,6 +3059,17 @@ public bool DontTurnOrMoveWhenGiving
...
+        /// <summary>
+        /// Determines the rotation speed for projectiles in global X
+        /// </summary>
+        public double? RotationSpeed
+        {
+            get => GetProperty(PropertyFloat.RotationSpeed);
+            set { if (!value.HasValue) RemoveProperty(PropertyFloat.RotationSpeed); else SetProperty(PropertyFloat.RotationSpeed, value.Value); }
+        }
+
+        public bool HasMissileFlightPlacement => CSetup.PlacementFrames.ContainsKey((int)ACE.Entity.Enum.Placement.MissileFlight);
```
gmriggs, line 3071: seems like this would be worth caching somewhere into a simple bool. doing a dictionary lookup for every projectile launched is not exactly ideal

Issue #1215, opened 2019-01-01
Olthoi Harvesters & Gardeners Do Not Drop Loot in Arcade North on ML


someeodtech:
### Bug Report or Feature Request (mark with an `x`)
```
- [x ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.

Killing any Olthoi Harvester (11480, 11727) or Olthoi Gardener (11479. 11726) in Olthoi Arcade North in ML

### The log given by the failure.

N/A

gmriggs:
Fixed by @dgatewood in https://github.com/ACEmulator/ACE/pull/1205

gmriggs:
The fix from @dgatewood has been ported to https://github.com/ACEmulator/ACE/pull/1338, and is now in the master branch

PR #3755, opened 2021-12-10
Update allow verdant force gem to all melee weps

Schneebs:
added per gmriggs in case some melee weps were missed in the recipe

gmriggs — Today at 1:42 AM
if something has a simple rule like "100% of the melee weapons in the game", RecipeManager_New is also worth looking at


Schneebs:
ACE change needs to be implemented prior to #1215 or it will break all verdant force applications.

PR #3763, opened 2021-12-23
Catch null Weenie on Appraisal

LtRipley36706:


Source/ACE.Server/Network/Structure/AppraiseInfo.cs
```diff
@@ -545,7 +545,7 @@ private void BuildCreature(Creature creature)
             if (PropertiesInt.ContainsKey(PropertyInt.EncumbranceVal) && !NPCLooksLikeObject)
                 PropertiesInt.Remove(PropertyInt.EncumbranceVal);
 
-            if (NPCLooksLikeObject && !creature.Weenie.GetProperty(PropertyInt.EncumbranceVal).HasValue)
+            if (NPCLooksLikeObject && !(creature.Weenie?.GetProperty(PropertyInt.EncumbranceVal).HasValue ?? false))
```
gmriggs, line 548: a WorldObject's weenie should not be null

it should be discovered why this is happening

gmriggs (approved):

PR #3775, opened 2022-01-06
Remove incorrect spells learned from Stipends

LtRipley36706:


Boop-AC:
4538 (Incantation of Finesse Weapon Mastery Self) shouldn't be on this list since it is craftable

PR #3839, opened 2022-04-27
very basic preliminary framework for slag drops

gmriggs:
wip code from months ago

slag drops from olthoi players killing monsters is the only piece that has really been fleshed out

slag drops from olthoi players killing players is still todo, and has placeholder code


ziang4891 (approved):
Play tested on my own server, and this PR worked well.  Drops a static amount from player kills, but that can be adjusted later.  Olthoi play is still an optional feature, anyways.

LtRipley36706 (approved):

PR #3902, opened 2022-09-20
Fix: Attributes able to be augmented above 100 innate

alissa0:
This is an attempt to amicably resolve an issue in which players are able to augment innate attributes above 100 innate. In an attempt to respect the ambiguous nature of this functionality, it was written such that server operators must opt-in to this functionality.

**To reproduce**:
Create a character with 99 innate coordination, purchase the Oswald's Enhancement augmentation once = 104 innate coordination.

**Resolution**
This adds a new server side setting ("attribute_augmentation_safety_cap") which, when enabled, will prevent players with >= 96 in an innate attribute from purchasing augmentations for said attributes. Furthermore, the 'verify-attributes' command will retroactively fix characters who have taken advantage of this by redistributing 5 points of an attribute in which this was exploited to the lowest available attribute (so that the points are not lost), while resetting the player's attribute reset timer so that they may legitimately redistribute the points at they please.


Source/ACE.Server/Managers/PropertyManager.cs
```diff
@@ -510,6 +510,7 @@ public static void LoadDefaultProperties()
                 ("allow_pkl_bump", new Property<bool>(true, "enables retail behavior where /pkl checks for entry collisions, bumping the player position over if standing on another PKLite. This effectively enables /pkl door skipping from retail")),
                 ("allow_summoning_killtask_multicredit", new Property<bool>(true, "enables retail behavior where a summoner can get multiple killtask credits from a monster")),
                 ("assess_creature_mod", new Property<bool>(false, "(non-retail function) If enabled, re-enables former skill formula, when assess creature skill is not trained or spec'ed")),
+                ("attribute_augmentation_safety_cap", new Property<bool>(false, "if TRUE players are not able to use attribute augmentations if the innate value of the target attribute is >= 96. All normal restrictions to these augmentations still apply.")),
```
LtRipley36706, line 513:
```suggestion
                ("attribute_augmentation_safety_cap", new Property<bool>(true, "if TRUE players are not able to use attribute augmentations if the innate value of the target attribute is >= 96. All normal restrictions to these augmentations still apply.")),
```

LtRipley36706 (changes requested):

LtRipley36706 (approved):

PR #3917, opened 2022-10-26
No burden before level 10 (configurable)

Advan-tage:
added no burden for players up to a configurable level (default set to 10) as to keep metas from benefiting. but still able to be helpful for mules or key pullers.


Source/ACE.Server/Entity/Fellowship.cs
```diff
@@ -565,6 +565,8 @@ public void SplitXp(ulong amount, XpType xpType, ShareType shareType, Player pla
         /// <param name="player">The fellowship member who originated the luminance</param>
         public void SplitLuminance(ulong amount, XpType xpType, ShareType shareType, Player player)
         {
+            return; //Hard disable luminance
```
OptimShi, line 568: There's several of these `Hard disable luminance` hacks in this PR which obviously break desired behavior.

PR #3920, opened 2022-10-28
Add loot to OlthoiPlayers killed by PKs

LtRipley36706:
For OlthoiPlayers who have no vitae, and are killed by PKs, they are meant to drop top tier loot and sometimes a Mutated Olthoi Gland

Additionally, they drop nothing to mobs, and never store their last corpse location. 

Also when PKs are killed by OlthoiPlayers, they too never store their last corpse location, nor drop any of their possessions so no notification is made about Augs or items dropped/retained.


Source/ACE.Server/WorldObjects/Player_Death.cs
```diff
@@ -1034,19 +1034,66 @@ public List<WorldObject> HandleDestroyBonded()
...
+            else if (killerIsPkPlayer)
+            {
+                if (hadVitae)
+                    return new();
 
-            if (!corpse.TryAddToInventory(slag))
-                log.Warn($"Player_Death: couldn't add item to {Name}'s corpse: {slag.Name}");
+                var items = LootGenerationFactory.CreateRandomLootObjects(new()
+                {
+                    TreasureType                            = 2000,
```
gmriggs, line 1060: should this be added to the world db?

LtRipley36706: moved to db with coded fall back

PR #3922, opened 2022-11-20
Adjust crafting motion

LtRipley36706:
Also disallowed crafting when not at peace per weenie errors seen in retail

### DIFF
--- a/Source/ACE.Server/Managers/RecipeManager.cs
+++ b/Source/ACE.Server/Managers/RecipeManager.cs
@@ -47,6 +47,12 @@ namespace ACE.Server.Managers
                 return;
             }
 
+            if (player.CombatMode != CombatMode.NonCombat)
+            {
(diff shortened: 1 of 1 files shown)


jtg0285 (commented):
"Also disallowed crafting when not at peace per weenie errors seen in retail"

This is not retail accurate.  Retail would throw a weenie error, but it would also take you out of combat so that subsequent attempts WOULD succeed.

LtRipley36706:
PCAP doesn't show that. The one I reference the WERROR on shows two failed attempts to craft, and after the second the player initiated the combat mode change.

Gordian3:
This PR should have had a server setting to control it and not just made the change.
Most people did not want this regardless of the retail discussion.   I understand the ACE mission but a server control setting really was needed for this.  @LtRipley36706

PR #3941, opened 2022-12-16
Allow certain crafting to occur in combat based on server property

Gordian3:
A recent change https://github.com/ACEmulator/ACE/pull/3922 made it so players may no longer craft arrows while in combat mode.  Since most players don't seem to like this change, this PR adds a Server Property to control the behavior.  The default is to disallow combat mode crafting.  However, with this setting, server operators may choose to enable in combat mode crafting (generally arrows).

### DIFF
--- a/Source/ACE.Server/Managers/PropertyManager.cs
+++ b/Source/ACE.Server/Managers/PropertyManager.cs
@@ -502,7 +502,7 @@ namespace ACE.Server.Managers
                 ("advanced_combat_pets", new Property<bool>(false, "(non-retail function) If enabled, Combat Pets can cast spells")),
                 ("advocate_fane_auto_bestow", new Property<bool>(false, "If enabled, Advocate Fane will automatically bestow new advocates to advocate_fane_auto_bestow_level")),
                 ("aetheria_heal_color", new Property<bool>(false, "If enabled, changes the aetheria healing over time messages from the default retail red color to green")),
-                ("allow_combat_mode_craft", new Property<bool>(false, "enables ability to do things like craft arrows when not in peace mode")),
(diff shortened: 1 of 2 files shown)


Gordian3:
This isn't ready yet.  This would be my 1st contribution on this repo.   I'm trying to figure out the process here.  I was looking for a label "Do Not Merge" or something like that but I can't set labels.   Anyway, I need to do testing but I guess wouldn't mind comments and pointers for the process to follow here.

gmriggs:
You would need to re-add the stance swapping code back in that the original PR removed. Simply playing the crafting motion from any stance won't work as expected in all scenarios. Certain animations are only available from certain stances, and crafting motions need to be played from the NonCombat stance. The original code before the PR had all of the stance-swapping logic to drop out of combat mode, wait the appropriate amount of time, and then perform the crafting operation.

Gordian3:
Thanks!  Will update, test and push changes this weekend.

Gordian3:
Will do testing this weekend.  Do not merge.

Gordian3:
Tested.  Observed and fixed issue with the ordering.

Tests:
1. Default for the "allow_combat_mode_craft" is false
2. Behavior when false matches changes added previously.
3. Enable of "allow_combat_mode_craft"
4. Manual operation allows for crafting when in combat mode.  It will leave player in peace mode (matching original implementation).  This is as opposed to being prevented from creating the arrows while in combat.
5. Tested multiple variations of the above.
6. Tested with vTank with various settings.  Arrow crafting in combat mode works and vTank restores combat to continue attacks. (Note: Also observed this with the setting 'false')

I believe this is ready for review.
@gmriggs

gmriggs:
Looks good to me

Gordian3:
I will update: https://github.com/ACEmulator/ACE/wiki/Server-Configurable-Options once this is approved or merged.

LtRipley36706 (approved):

gmriggs (approved):

PR #3942, opened 2022-12-16
Switch GetXPBonus to use GetEnchantments_TopLayer

LtRipley36706:


Source/ACE.Server/WorldObjects/Managers/EnchantmentManager.cs
```diff
@@ -721,7 +729,7 @@ public virtual float GetVitalMod_Multiplier(CreatureVital vital)
         /// </summary>
         public virtual float GetXPBonus()
         {
-            var enchantments = GetEnchantments(SpellCategory.TrinketXPRaising);
+            var enchantments = GetEnchantments_TopLayer(SpellCategory.TrinketXPRaising);
```
gmriggs, line 732: if this function doesn't already exist, then it should probably just sort by descending StatModVal, and return the first one

gmriggs: foreach (var enchantment in enchantments.OrderByDescending(i => i.StatModValue).Take(1))

Source/ACE.Entity/Models/PropertiesEnchantmentRegistryExtensions.cs
```diff
@@ -162,6 +162,32 @@ select categories.OrderByDescending(c => c.PowerLevel)
             }
         }
 
+        public static List<PropertiesEnchantmentRegistry> GetEnchantmentsTopLayerBySpellCategory(this ICollection<PropertiesEnchantmentRegistry> value, SpellCategory spellCategory, ReaderWriterLockSlim rwLock, HashSet<int> setSpells)
```
gmriggs, line 165: all of these functions are no longer needed

gmriggs (dismissed):
looks good to me, assuming this works in-game. another option would be to sort by descending powerlevel, although this should be equivalent

gmriggs (approved):

PR #3944, opened 2022-12-23
Server setting to enable substituting legendary key rewards emotes

Gordian3:
This PR provides a server setting "award_wcid_for_legendary" that allows for substitution of legendary keys with the provided WCID.  For instance, a server offering legendary keys for sale on a vendor for 1 MMD per pull may set this to 20630.  This allows players to not have to deal with large numbers of unstackable legendary keys from Kill Tasks and other quests.  Another option would be Promissory Notes (43901).  The server operator can provide any WCID but generally it would be something stackable.

This change only impacts emote based giving.  It does not change any looting of legendary keys from corpses or chests such as at the casinos.

The default option is "0" which is interpreted as disabled so there is no substitution.

The actual rewards to users depend a bit on how the quest emotes are configured.  If they offer 5 rewards, each of 1 Aged Legendary Key, then the player would get 5 different of the substituted item.  You can see this in the Tou Tou KTs.  For Frozen Valley Gurog, it is a singe reward of multiple keys.  In this case the new items (promissory notes in this case) are stacked.  At Hoshino's the Durable keys become stacks of 10 of the item.

AgedLegendaries_grp_5
AgedLegendaries_grp_5_log

DurableKey_subst_reward
DurableKey_subst_reward_log

### DIFF
--- a/Source/ACE.Server/WorldObjects/Player_Inventory.cs
+++ b/Source/ACE.Server/WorldObjects/Player_Inventory.cs
@@ -3383,7 +3383,34 @@ namespace ACE.Server.WorldObjects
             Prev_PutItemInContainer[1] = Prev_PutItemInContainer[0];
             Prev_PutItemInContainer[0] = new PutItemInContainerEvent(itemGuid, containerGuid, placement);
         }
-        
+
+        // Map of the various legendary keys that may be given on emote/quest and the number of uses.
+        // Static map for quick comparisons
+        static Dictionary<uint, int> _LegendaryKeyUses = new Dictionary<uint, int>()
+        {
+            {48746, 1 }, // Aged Legendary Key
+            {48747, 1 }, // 24 hour, single use
+            {48748, 2 }, // 24 hour, 2 use
+            {48749, 3 }, // 24 hour, 3 use
+            {48750, 4 }, // 24 hour, 4 use
+            {48914, 1 }, // 24 hour, 1 use, quest legendary chest key
+            {51558, 1 }, // 1 use
+            {51586, 3 }, // 24 hour, 3 use, quest legendary chest key
+            {51648, 3 }, // 24 hour, 3 use, quest legendary chest key
+            {51954, 10 }, // Durable Legendary Key
+            {51963, 25 }, // 25 use
+            {52010, 5 }, // 24 hour, 5 use, quest legendary chest key
+            {72048, 1 }, // 24 hour, 1 use, quest legendary chest key non-pcap
+            {72338, 3 }, // 24 hour, 3 use, quest legendary chest key non-pcap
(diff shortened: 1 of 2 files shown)


Source/ACE.Server/WorldObjects/Player_Inventory.cs
```diff
@@ -3383,7 +3383,34 @@ private void OnPutItemInContainer(uint itemGuid, uint containerGuid, int placeme
             Prev_PutItemInContainer[1] = Prev_PutItemInContainer[0];
             Prev_PutItemInContainer[0] = new PutItemInContainerEvent(itemGuid, containerGuid, placement);
         }
-        
+
+        // Map of the various legendary keys that may be given on emote/quest and the number of uses.
+        // Static map for quick comparisons
+        static Dictionary<uint, int> _LegendaryKeyUses = new Dictionary<uint, int>()
+        {
+            {48746, 1 }, // Aged Legendary Key
```
Gordian3, line 3391: If there is something I can use other than the literal values, please provide an example if that is preferred.

LtRipley36706 (commented):
This kind of thing should be managed at the content level and not in code and applied globally really

Gordian3:
> This kind of thing should be managed at the content level and not in code and applied globally really

It's a server option.  Controlling it at the content level would require a large number of changes and would not be as dynamic.  Additionally, it would cause more merge conflicts for custom servers.

LtRipley36706:
> > This kind of thing should be managed at the content level and not in code and applied globally really
> 
> It's a server option.  Controlling it at the content level would require a large number of changes and would not be as dynamic.  Additionally, it would cause more merge conflicts for custom servers.
> 

I understand its goal, however its application is simultaneously too broad and too specific. This just ends up adding complexity to the code thats better handled in content

Gordian3:
To make it more generic, I could switch this to a string property that contains the name of a file.  The file could be CSV or perhaps JSON list of WCID substitutions.  Once the property is set, the list is loaded (or reloaded).  This sets a dictionary/map of original and substituted WCIDs.  Then when an emote award is given, the list is searched to see if it has the WCID and if so then it substitutes the new value.

Also, during server startup it would also check if this value is set and if so, will load the file.

To undo the substitutions, simply specify an empty file.

Thoughts @LtRipley36706 ?   Would this be generic enough of a utility?

Source/ACE.Server/WorldObjects/Player_Inventory.cs
gmriggs: WeenieClassName

gmriggs:
Something like this seems like it would be a perfect use of the new mod system:

https://github.com/ACEmulator/ACE/wiki/Modding

PR #3945, opened 2022-12-29
Catch and Fix NaN floats in Positions

LtRipley36706:
Co-Authored-By: Mag-nus <Mag-nus@users.noreply.github.com>

### DIFF
--- a/Source/ACE.Database/Adapter/BiotaConverter.cs
+++ b/Source/ACE.Database/Adapter/BiotaConverter.cs
@@ -530,28 +530,12 @@ namespace ACE.Database.Adapter
                     // Entity Framework is unable to store NaN floats in the database and results in an error of:
                     // ERROR 1054: Unknown column 'NaN' in 'field list'
 
-                    var isNaNCount = 0;
-                    if (float.IsNaN(entity.AnglesX))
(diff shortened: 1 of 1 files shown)


Source/ACE.Database/Adapter/BiotaConverter.cs
```diff
@@ -527,6 +527,33 @@ public static ACE.Database.Models.Shard.Biota ConvertFromEntityBiota(ACE.Entity.
                 {
                     var entity = new BiotaPropertiesPosition { ObjectId = biota.Id, PositionType = (ushort)kvp.Key, ObjCellId = kvp.Value.ObjCellId, OriginX = kvp.Value.PositionX, OriginY = kvp.Value.PositionY, OriginZ = kvp.Value.PositionZ, AnglesW = kvp.Value.RotationW, AnglesX = kvp.Value.RotationX, AnglesY = kvp.Value.RotationY, AnglesZ = kvp.Value.RotationZ };
 
+                    // Entity Framework is unable to store NaN floats in the database and results in an error of:
+                    // ERROR 1054: Unknown column 'NaN' in 'field list'
+
+                    var isNaNCount = 0;
+                    if (float.IsNaN(entity.AnglesX))
+                    {
+                        entity.AnglesX = 0;
```
gmriggs, line 536: if only some of these values are adjusted, but not others, you could end up with a non-normalized quaternion, which would be very bad

a better strategy might be if any fields are NaN, then to set the quaternion to identity. this would match the same strategy used by the acclient physics engine

gmriggs (approved):

PR #3975, opened 2023-04-17
Update BasePropertyDesc.cs

LtRipley36706:
Clean up compiler warnings

### DIFF
--- a/Source/ACE.DatLoader/Entity/BasePropertyDesc.cs
+++ b/Source/ACE.DatLoader/Entity/BasePropertyDesc.cs
@@ -36,7 +36,7 @@ namespace ACE.DatLoader.Entity
         PropertyInheritanceType m_inheritanceType;
         PropertyDatFileType m_datFileType;
         PropertyPropagationType m_propagationType;
-        PropertyCachingType m_cachingType;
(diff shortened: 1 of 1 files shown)


Source/ACE.DatLoader/Entity/BasePropertyDesc.cs
```diff
@@ -36,7 +36,7 @@ public class BasePropertyDesc : IUnpackable
         PropertyInheritanceType m_inheritanceType;
         PropertyDatFileType m_datFileType;
         PropertyPropagationType m_propagationType;
-        PropertyCachingType m_cachingType;
+        //PropertyCachingType m_cachingType;
```
LtRipley36706, line 39: @OptimShi This variable was never assigned a value, was that missed in Unpack somewhere?

OptimShi: Honestly, so many of these properties/var names were just guessing based on the class details in the PDB that cachingType entry could been any one of those "bool" values, or it could be a run-time value (and I'm sure some of those bool var names are mislabled/flat out wrong, too)

PR #3977, opened 2023-05-01
Fix healing snapshot

Boop-AC:
Changes the healing (kit) code so that amount and success are determined at the beginning of the healing animation rather than at the end in order to match retail behavior. For example, if you have 400/500 health, begin healing, and take 50 damage during the animation ACE currently allows you to heal all 150 (max ending 500/500) instead of only the initial 100 (max ending 450/500). 
Retail video as evidence: https://www.youtube.com/watch?v=lK4RQiDiKOg
Best timestamp is at 2:18:29. Starts healing at 485/514 health. Receives 46 damage (439 health), is healed by another's spell for 75 (514 health), then finishes healing for 29 health points (initial missing health). 
Some more timestamps: 
Receiving damage during animation:
1:13:16, 2:30:43, 2:38:09
Receiving healing (spell) during animation:
2:17:00, 2:17:32, 2:18:40, 2:18:51, 2:19:35, 2:26:33, 2:35:26, 2:39:52

### DIFF
--- a/Source/ACE.Server/WorldObjects/Healer.cs
+++ b/Source/ACE.Server/WorldObjects/Healer.cs
@@ -128,6 +128,14 @@ namespace ACE.Server.WorldObjects
 
             healer.IsBusy = true;
 
+            //Skill check and amount healed are locked at the beginning of the attempt
+            var difficulty = 0;
+            var critical = false;
+            uint staminaCost = 0;
+            var vital = target.GetCreatureVital(BoosterEnum);
+            var skillCheck = DoSkillCheck(healer, target, vital, ref difficulty);
+            var healAmount = skillCheck ? GetHealAmount(healer, target, vital, out critical, out staminaCost) : 0;
+
             var motionCommand = healer.Equals(target) ? MotionCommand.SkillHealSelf : MotionCommand.SkillHealOther;
 
             var motion = new Motion(healer, motionCommand);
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/WorldObjects/Healer.cs
```diff
@@ -263,18 +268,11 @@ public uint GetHealAmount(Player healer, Player target, CreatureVital vital, out
...
-            var ratingMod = target.GetHealingRatingMod();
-
-            healAmount *= ratingMod;
-
             // chance for critical healing
             criticalHeal = ThreadSafeRandom.Next(0.0f, 1.0f) < 0.1f;
             if (criticalHeal) healAmount *= 2;
 
             // cap to missing vital
-            var missingVital = vital.MaxValue - vital.Current;
```
gmriggs, line 277: should be ``var missingVital = Math.Min(vital.Missing, prevMissingVital);`` i think

if the player gets healed from another source while they are performing the healing animation, it seems like the pre-cap to max vital could have an overage without this line

Boop-AC: That is intentional. The "Receiving healing (spell) during animation:" timestamps in the commit comment show that the displayed heal amount was not reduced by any healing received during the animation. It also appears that the stamina consumption is in line with the displayed amount and not the actual amount. 
Personally I think it's a dumb retail bug and would have no issues correcting the displayed amount to the actual amount healed, but I suppose it needs to stay that way to be retail accurate.

gmriggs (approved):

Issue #3919, opened 2022-10-28
New Server Setup Error: System.InvalidCastException 


delizin:
### Bug Report or Feature Request (mark with an `x`)
```
- [ x ] bug report -> please search issues before submitting
- [ ] feature request
```

### Repro steps.
Follow the steps here to setup the database using MariaDB 10.10: https://github.com/ACEmulator/ACE/wiki/ACE-Hosting
Then the following steps to setup for Windows: https://github.com/ACEmulator/ACE/wiki/ACE-Hosting-Windows

### The log given by the failure.
2022-10-27 22:57:59,682 INFO : Initializing ConfigManager...
2022-10-27 22:57:59,683 INFO : Pruning invalid friends from all friend lists...
2022-10-27 22:58:00,870 INFO : Pruned 0 invalid friends found on friend lists.
2022-10-27 22:58:00,870 INFO : Automatic Server version check started...
2022-10-27 22:58:01,127 INFO : Current Server Binary: 1.48.4280.20221019161439-master-fa6fe39
2022-10-27 22:58:01,430 INFO : Latest Server Version is v1.48.4280 -- No Update Required!
2022-10-27 22:58:01,433 INFO : Automatic World Database Update started...
2022-10-27 22:58:01,477 INFO : Current World Database version: Base - v0.8.5 | Patch - v0.9.262
2022-10-27 22:58:01,652 INFO : Latest patch version is v0.9.262 -- No Update Required!
2022-10-27 22:58:01,652 INFO : Automatic World Database Update complete.
Unhandled exception. System.InvalidCastException: Object cannot be cast from DBNull to other types.
   at System.DBNull.System.IConvertible.ToInt32(IFormatProvider provider)
   at MySql.Data.MySqlClient.Driver.LoadCharacterSets(MySqlConnection connection)
   at MySql.Data.MySqlClient.Driver.Configure(MySqlConnection connection)
   at MySql.Data.MySqlClient.MySqlConnection.Open()
   at ACE.Server.Program.GetContentFolder() in C:\projects\ace\Source\ACE.Server\Program_DbUpdates.cs:line 153
   at ACE.Server.Program.AutoApplyWorldCustomizations() in C:\projects\ace\Source\ACE.Server\Program_DbUpdates.cs:line 185
   at ACE.Server.Program.Main(String[] args) in C:\projects\ace\Source\ACE.Server\Program.cs:line 205
[...]

PR #4003, opened 2023-05-13
Mysqlconnector3919: Fix auto update issues that arise in newer versions of MariaDb

alissa0:
When setting up a new dev environment recently I encountered the same issue described in #3919 

Some cursory research lead me to a discussion about this on [Stack Overflow](https://stackoverflow.com/questions/74060289/mysqlconnection-open-system-invalidcastexception-object-cannot-be-cast-from-d)

I decided to implement MySqlConnector as an alternative to MySql.Data, but it wasn't a drop-in replacement as it lacks support for the MySqlScript methods (including the StatementExecuted event) so some light refactoring had to be done.


gmriggs:
Is this a widely-recognized solution, or just switching to another library that doesn't happen to have the issue?

IIRC, I had an issue within the past year when upgrading an existing ACE DB to a newer MariaDB version. I think it was maybe 10.6, but it was over a year ago, and I can't remember the exact version. Anyway, there was an issue with utf8mb4 where it couldn't handle the older table format anymore. I mostly got utf8mb4 errors, but I think I might have seen a similar DBNull error in the mix as well.

I'm not sure if it's the same issue, but I ask because I found a bunch of varied solutions at the time. Some of them involved adjusting or switching MySqlConnector versions, but the cleanest solution I found was to simply upgrade my existing DB to the newer utf8mb4 format.

alissa0:
I spent a little bit of time trying to find additional information about this but I haven't really been able to find any extensive discussion around this issue aside from the stack overflow thread I linked initially :(
I was setting this up under a fresh archlinux installation in WSL2. As far as I can tell, it's using utf8mb4 across the board and I'm running into the issue around collations having NULL ids with the current driver.

So in short I don't think this is a widely recognized solution but the package does seem to be a widely utilized open source alternative to the one provided by Oracle. However, I can understand the hesitation surrounding changing something like this so if there's any additional info I can provide or research avenues I can explore let me know and I'll be happy to help!

gmriggs:
Okay, this sounds like a different issue than the upgrade one I had last year then.

Mag-nus:
I haven't spent enough time to give a proper review, but it just feels wrong going from what is the defacto standard for .NET to a 3rd party lib to fix an issue.

What I also found interesting is what we have in there now:
```
<Target Name="ChangeAliasesOfStrongNameAssemblies" BeforeTargets="FindReferenceAssembliesForReferences;ResolveReferences">
    <ItemGroup>
      <ReferencePath Condition="'%(FileName)' == 'MySqlConnector'">
        <Aliases>MySqlConnectorAlias</Aliases>
      </ReferencePath>
    </ItemGroup>
  </Target>
```

That also is worth a review.

LtRipley36706 (approved):
Confirmed this change fixes issue with MariaDB 10.10+ as well as appears to work with latest MySQL.

Both database engines perform correct out of box setup and installation.

PR #4015, opened 2023-06-10
Update DatFileType extended info

LtRipley36706:
Updated several properties in DatFileType to match what's in the decompiled client

### DIFF
--- a/Source/ACE.DatLoader/DatFileType.cs
+++ b/Source/ACE.DatLoader/DatFileType.cs
@@ -53,8 +53,7 @@ namespace ACE.DatLoader
         LandBlockInfo           = 2, // DB_TYPE_LBI
 
         [DatDatabaseType(DatDatabaseType.Cell)]
-        [DatFileTypeIdRange(0x01010000, 0x013EFFFF)]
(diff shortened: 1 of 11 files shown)


Source/ACE.DatLoader/DatFileType.cs
```diff
@@ -54,7 +54,7 @@ public enum DatFileType : uint
 
         [DatDatabaseType(DatDatabaseType.Cell)]
         [DatFileTypeIdRange(0x01010000, 0x013EFFFF)]
-        EnvCell                    = 3, // DB_TYPE_ENVCELL
+        Cell                    = 3, // DB_TYPE_CELL
```
gmriggs, line 57: this is going to break third party programs. this shouldn't be changed at this point

technically this range is envcells, so if it's labeled as cell in the pdb, the pdb would be inaccurate there.

gmriggs (changes requested):
see notes

Source/ACE.DatLoader/DatFileType.cs
LtRipley36706: reverted

gmriggs (approved):

PR #4084, opened 2024-01-07
Use known UTC offset instead of TimeZoneConverter (3rd party)

Mag-nus:


LtRipley36706:
Keep in mind that last time I tried to switch this, *nix users were still having issues with System.TimeZoneInfo

LtRipley36706:
See #3884

LtRipley36706 (approved):

PR #4085, opened 2024-01-07
Remove some unreferenced System packages

Mag-nus:
Not sure why these were ever added in the first place. They are not needed as nuget to build.


Mag-nus:
These were added during this commit: https://github.com/ACEmulator/ACE/pull/2853

but they shouldn't be needed to support ARM 64. Targeting arm during publish and installing the correct .NET arm package on the host should be enough.

LtRipley36706 (approved):

PR #4102, opened 2024-02-09
Adjust House UI for MaintenanceFree

LtRipley36706:


Source/ACE.Server/Network/Structure/HouseData.cs
```diff
@@ -56,6 +56,13 @@ public void SetRentItems(List<WorldObject> rentItems)
         /// </summary>
         public void SetPaidItems(SlumLord slumlord)
         {
+            if (slumlord.House?.HouseStatus == HouseStatus.InActive)
```
gmriggs, line 59: not sure if i agree with this part

LtRipley36706: Just makes the UI consistent. Changing the MaintenanceFree variable puts the green "rent paid" text on the UI, might as well indicate "paid in full" rent on both that UI and Slumlord UI by showing that as complete, otherwise it is potentially confusing that rent is paid but numbers say nothing is paid, and slumlord keeps rejecting attempts to fill the bucket.

gmriggs (approved):

PR #4112, opened 2024-02-18
Improve Sphere stack usage

Mag-nus:
Several of the functions in Sphere take parameters that were not used. This PR creates overloads that retain the existing signatures, marks them obsolete, and removes the unused parameters from the core function.

GetHashCode is optimized

FindTimeOfCollision is marked Static


Mag-nus:
Note, this maintains backwards compatibility with 3rd party users

LtRipley36706 (approved):

PR #4113, opened 2024-02-18
Improve sphere heap usage

Mag-nus:
Pass Center and Radius to static functions as parameters instead of instantiating a new Sphere on the heap and passing that through the code.

Reference: https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/performance/

This is built on top of the improve stack PR.


Mag-nus:
This also keeps the existing public API in tact

LtRipley36706 (dismissed):

PR #4127, opened 2024-03-03
Wire up GameEventMagicPurgeBadEnchantments

LtRipley36706:
According to PCAPs GameEventMagicPurgeBadEnchantments was sent to clients for players with Enduring Enchantment augmentation.

Client has slightly different check in CEnchantmentRegistry::PurgeBadEnchantmentList versus CEnchantmentRegistry::PurgeEnchantmentList
```cpp
    if ( v4->data._smod.type & 0x2000000 || v4->data._duration == -1.0 )
```
vs
```cpp
    if ( v4->data._duration == -1.0 )
```

### DIFF
--- a/Source/ACE.Server/Network/Structure/Enchantment.cs
+++ b/Source/ACE.Server/Network/Structure/Enchantment.cs
@@ -95,6 +95,9 @@ namespace ACE.Server.Network.Structure
             {
                 StatModType = spell.StatModType;
                 StatModKey = spell.StatModKey;
+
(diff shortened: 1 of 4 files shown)


LtRipley36706:
It turns out there are quite a lot of spells that would/should be considered Beneficial and marked as such in their StatModKey but are not currently. This would result in most "buffs" getting purged upon death. The Spells in the DAT has its own bitfield definition for Beneficial which would *seemingly* be the same in StatModKey. Using the DAT to supplement the missing bitfield from the server could work until the data were to be changed later.

PR #4137, opened 2024-03-20
Add option for unlimited GUID sequence gaps

LtRipley36706:
This allows servers to load *all* possible GUID sequence gaps instead of hard coded limit of 10,000,000.


Mag-nus:
Maybe we just make unlimited the default.

The negative impact is when a server becomes so guid fragmented, many objects will be required to represent all the sequence gaps.

We could add fragmentation detection at startup to warn server ops when to defragment, ie: if > gaps.count > 1000000 please use the offline consolidation tool

PR #4154, opened 2024-04-24
Test Lock vs ReaderWriterLockSlim

Mag-nus:
This is a rough test switching biota property get/set to be protected by a lock vs a readerwriterlockslim.

ReaderWriterLockSlim is more efficient when you have many readers vs seldom writers.

WIthout contention, Lock is ~5x more performant than ReaderWriterLockSlim.EnterReadLock

ACE current landblock group and threading model should result in minimal cases where multiple readers are accessing the same biota simultaneously, thus, lock may be the preferred choice.


Mag-nus:
For clarification, after testing on .NET 8, performance is not 5x, but maybe 0.5x

LtRipley36706:
Closed and Reopened because the merge status was busted.

PR #4182, opened 2024-05-15
Efnet8

Mag-nus:


Mag-nus:
This is simply updating the packages, scaffolding, and then copying the previous OnConfiguring.

No other changes have been made.

The scaffold decoration has changed and thus, it looks like more changes than it really is.

Mag-nus:
Incomplete: This still needs the DbContext custom code that is found in OnConfiguring().

Do not merge until OnCofiguring is fixed.

Mag-nus:
DbContext configurations are correct

LtRipley36706 (approved):

PR #4206, opened 2024-07-23
Improve performance with CharacterOptionExtensions

FlaggAC:
Improves performance with CharacterOptionExtensions. There is some overhead with calling GetAttributeOfType, so we are pre-caching these results.

Moves CharacterOptionExtensions from ACE.Entity to ACE.Server, as ACE.Entity's .NET Standard doesn't support FrozenDictionary or Enum.GetValues<T>.

I could rewrite this in a way that is compatible with .NET Standard, but it seemed more appropriate for the class to be in ACE.Server anyway so I moved it for now.

### DIFF
--- a/Source/ACE.Entity/Enum/CharacterOption.cs
+++ b/Source/ACE.Entity/Enum/CharacterOption.cs
@@ -180,14 +180,24 @@ namespace ACE.Entity.Enum
 
     public static class CharacterOptionExtensions
     {
+        private static Dictionary<CharacterOption, CharacterOptions1Attribute> CharacterOptions1Attributes { get; }
+        private static Dictionary<CharacterOption, CharacterOptions2Attribute> CharacterOptions2Attributes { get; }
+
+        static CharacterOptionExtensions()
+        {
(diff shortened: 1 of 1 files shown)


Source/ACE.Server/Entity/CharacterOptionExtensions.cs
Mag-nus, line 1: CharacterOptionExtensions should either remain in CharacterOptions.cs, or, if it is to be a separate file, it should be in the same folder as CharacterOptions.cs

Also, standard ACE using format is System on top, ACE below, separated by a space, ie:

using System;
using System.Collections.Frozen;
using System.Collections.Generic;
using System.Collections.ObjectModel;
using System.Linq;

using ACE.Entity;
using ACE.Entity.Enum;

Mag-nus (changes requested):

gmriggs:
where did CharacterOptionExtensions sit in the profile of the entire server?

before/after benchmarks for these changes?

Mag-nus (approved):

PR #4212, opened 2024-08-11
Performance improvements to Physics.Common.WeenieObject

FlaggAC:
Entire set of fields can be marked as readonly, as it never changes outside of the constructor.
Reduces unnecessary reflection operations by pre-fetching them during constructor
Changes WeenieObject to a sealed class to guarantee immutability
Immutability allows a shared DummyObject to be used by PhysicsObjs which do not have a weenie.

The reflection calls are made frequently by `PhysicsObj.FindObjCollisions(Transition)`, which is a very frequently-called method, so this is expected to improve performance. The 'is' operator is fast by reflection standards, but there is still some overhead with it, which should not be ignored considering the high frequency of the method calls.

ID and UpdateTime are commented out because they are never used.

### DIFF
--- a/Source/ACE.Server/Physics/Animation/MotionInterp.cs
+++ b/Source/ACE.Server/Physics/Animation/MotionInterp.cs
@@ -176,7 +176,7 @@ namespace ACE.Server.Physics.Animation
         {
             if (PhysicsObj == null) return;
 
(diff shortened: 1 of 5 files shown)


Mag-nus (approved):

Mag-nus:
Merging. Will have to re-apply one fix immediately after.

PR #4224, opened 2024-09-15
Update output of DateTime strings

LtRipley36706:
Define and use explicit formatting for DateTime objects used throughout source


Mag-nus:
Since this is for logging, can we use a version of ISO 8601
https://en.wikipedia.org/wiki/ISO_8601

LtRipley36706:
some of it ends up user facing, but most of it is for debug/logging.

I'm not certain if you're asking if the program itself can switch to ISO 8601 or if you're wanting the output to conform to a specific format, but personally I'd prefer when looking at logs/debugging that have purposeful datetimes being shown to see it in the format defined explicitly in the code. This is what is essentially the default "G" format seen here: https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings

`2009-06-15T13:45:30 -> 6/15/2009 1:45:30 PM (en-US)`

But because you cannot rely on system locale, culture and charsets to not change (as seen in the upgrade to Ubuntu 24.04 from 22.04) Microsoft seemed to recommend explicit format definitions.

Mag-nus:
I'm fine with explicit overrides. My main issue is with M/d/yyyy. Month/Day/Year and Day/Month/Year are mixed in various regions of the world but yyyy-MM-dd is fixed globally which is what ISO 8601 defines.

ISO 8601 defines a few formats that you might find visually appealing that still work globally and abide by the ISO standard.

LtRipley36706:
Good with `yyyy-MM-dd h:mm:ss tt`?

PR #4318, opened 2025-06-06
Remove Duplicative words and fix typos in comments

KT053374:


KT053374:
> looks like something that shouldn't have been sent as pr?

Apologies

PR #4328, opened 2025-06-26
Database tooling additions for Linux/macOS users

KT053374:
Added .sh scripts for Linux/macOS users for DB export and rebase.


LtRipley36706 (approved):

KT053374:
> the docker container name specified in docker-compose.yml is ace-db and not acedb

You are correct in this observation.  I noticed that several of the original .bat files, which were referenced to build the .sh files, also contain the same incorrect container name.  I will work on correcting these to match the docker-compose.yml and the environment variables in docker.env

PR #4486, opened 2026-08-29
fix give stack amount failure message

gmriggs:
Thanks to Xenocide for reporting this issue!


LtRipley36706 (approved):

LtRipley36706:
do you think the code should sanitize the amount it shows the player incase client is sending screwy value? 

eg If amount is greater than max size, use the stacksize max?

gmriggs:
it already does in the 1 function that calls it, HandleActionGiveObjectRequest:

https://github.com/gmriggs/ACE/blob/25ea2267e2c20acc511112dbeb34de8f104060b0/Source/ACE.Server/WorldObjects/Player_Inventory.cs#L3199
https://github.com/gmriggs/ACE/blob/25ea2267e2c20acc511112dbeb34de8f104060b0/Source/ACE.Server/WorldObjects/Player_Inventory.cs#L3224

